PR #6, opened 2020-03-25
Validating auth signatures

SmailBarkouch:
Edited server:
- Added a verification to the auth provided from "NewAccount"  matches the other credentials in the struct when decrypted. (src/api/new_account.rs)

Edited core:
- Made changes to crypto in terms of "RsaCryptoService" and "CryptoService".
- Added a code to create an auth, which is of type "EncryptedValue". (src/account_api.rs)


Parth:
There are some garbage files still here, maybe you need to pull from `master` to fix

core/src/account_api.rs
```diff
@@ -1,33 +1,147 @@
...
+    }
+}
+
 pub trait AccountApi {
     fn new_account(account: &Account) -> Result<(), Error>;
 }
 
 pub struct AccountApiImpl;
 
+pub trait AuthService {
```
Parth, line 41: Break this out into it's own file

core/src/account_api.rs
```diff
@@ -1,33 +1,147 @@
...
+    }
+}
+
 impl From<reqwest::Error> for Error {
     fn from(e: reqwest::Error) -> Self {
         NetworkError(e)
     }
 }
 
+impl From<AuthError> for Error {
```
Parth, line 128: Should go away

core/src/account_api.rs
```diff
@@ -1,33 +1,147 @@
...
+            AuthError::IncorrectAuth(IncorrectUsername) => IncorrectUsername,
+            AuthError::IncorrectAuth(ExpiredAuth) => ExpiredAuth,
+            _ => CryptoError
+        }
+    }
+}
+
 impl AccountApi for AccountApiImpl {
     fn new_account(account: &Account) -> Result<(), Error> {
+        let auth = AuthServiceImpl::generate_auth(&account.keys, &account.username)?;
```
Parth, line 140: Consider doing this in `AccountService` and using `lockbook_api.new_account`

core/src/account_api.rs
```diff
@@ -77,4 +191,30 @@ mod integration_tests {
             }
         }
     }
+
+    #[test]
```
Parth, line 195: Ways to improve test:
+ Make a `Clock` which gives `AuthService` time, and mock it out here
+ Hold the `key` constant and then assert specific values
+ Make sure your tests cover all the branches of your code

PR #12, opened 2020-04-12
Implements rocket::FromForm for requests in core

tvanderstad:


tvanderstad:
not tryna introduce a rocket dependency in core

PR #16, opened 2020-04-13
Swapping out rusqlite for sled

raayan:
Swapping out legacy-_finance_ `rusqlite` for hot and fresh `sled` 
Adds `tempfile` to make a `TempDirBackedDb` for use in tests!
Also re-organized the project in this a bit, not attached to it at all.
Screen Shot 2020-04-13 at 11 19 54 AM


Parth:
`TempDirBackedDb` does this expose anything in memory?

if it's just a KV store, I can see us just mocking it out with hashtables.

PR #26, opened 2020-04-19
Legit File Sync with Crypto and iOS Client

raayan:


docs/README.md
```diff
@@ -115,3 +115,5 @@ If you don't tell `rustup` to use nightly you'll get the following error
 ```shell script
 error[E0554]: `#![feature]` may not be used on the stable release channel
 ```
+
```
Parth, line 118: ?

core/src/client/mod.rs
```diff
@@ -12,37 +13,48 @@ pub use self::change_file_content::{
...
+pub use self::get_file::{get_file, GetFileError, GetFileRequest};
 pub use self::get_updates::{get_updates, FileMetadata, GetUpdatesError, GetUpdatesRequest};
 pub use self::move_file::{move_file, MoveFileError, MoveFileRequest, MoveFileResponse};
 pub use self::new_account::{new_account, NewAccountError, NewAccountRequest, NewAccountResponse};
 pub use self::rename_file::{rename_file, RenameFileError, RenameFileRequest, RenameFileResponse};
+use crate::service::file_encryption_service::EncryptedFile;
+use crate::{API_LOC, BUCKET_LOC};
 
 #[derive(Debug)]
 pub enum ClientError {
```
Parth, line 25: This seems like an anti-pattern, bunching these together. The larger this grows the more things people have to handle, these calls should return their own enums.

core/src/client/mod.rs
```diff
@@ -12,37 +13,48 @@ pub use self::change_file_content::{
...
+    fn change_file(params: &ChangeFileContentRequest) -> Result<u64, ClientError>;
 }
 
 pub struct ClientImpl;
-
 impl Client for ClientImpl {
-    fn new_account(api_location: String, params: &NewAccountRequest) -> Result<(), ClientError> {
-        new_account(api_location, params).map_err(|err| ClientError::AccountError(err))
+    fn new_account(params: &NewAccountRequest) -> Result<(), ClientError> {
+        new_account(API_LOC.to_string(), params).map_err(|err| ClientError::CreateAccount(err))
```
Parth, line 44: Just don't map these

core/src/lib.rs
```diff
@@ -24,20 +27,42 @@ pub mod repo;
...
+    DefaultFileRepo,
+    DefaultAcountRepo,
+    DefaultFileEncryptionService,
+>;
 
 static FAILURE_DB: &str = "FAILURE<DB_ERROR>";
 static FAILURE_ACCOUNT: &str = "FAILURE<ACCOUNT_MISSING>";
 
+static FAILURE_META_CREATE: &str = "FAILURE<META_CREATE>";
+// static FAILURE_META_UPDATE: &str = "FAILURE<META_UPDATE>";
```
Parth, line 60: use it or lose it

core/src/model/file.rs
```diff
@@ -0,0 +1,8 @@
+use serde::{Deserialize, Serialize};
+use std::clone::Clone;
+
+#[derive(Clone, PartialEq, Debug, Deserialize, Serialize)]
+pub struct File {
```
Parth, line 5: `No usages found in All Places`

core/src/repo/file_metadata_repo.rs
```diff
@@ -49,10 +67,14 @@ impl FileMetadataRepo for FileMetadataRepoImpl {
         Ok(file_metadata)
     }
 
-    fn last_updated(db: &Db) -> Result<i64, Error> {
-        Ok(Self::get_all(db)?
-            .iter()
-            .fold(0, |max, meta| max.max(meta.updated_at)))
+    fn last_updated(db: &Db) -> Result<u64, Error> {
+        Ok(Self::get_all(db)?.iter().fold(0, |max, meta| {
+            // if meta.status != Status::Local {
```
Parth, line 72: What's up with this commented out code?

core/src/service/account_service.rs
```diff
@@ -40,21 +40,32 @@ impl<Crypto: PubKeyCryptoService, AccountDb: AccountRepo, ApiClient: Client> Acc
...
             public_key,
         };
 
-        ApiClient::new_account(API_LOC.to_string(), &new_account_request)?;
+        ApiClient::new_account(&new_account_request)?;
+
+        Ok(account)
+    }
+
+    fn load_account(db: &Db, username: String, key_string: String) -> Result<Account, Error> {
```
Parth, line 64: Let's make this concept called `import` everywhere. "Import existing Lockbook"

core/src/service/crypto_service.rs
```diff
@@ -14,16 +15,17 @@ use self::rand::RngCore;
...
 use self::rsa::{PaddingScheme, PublicKey, RSAPrivateKey, RSAPublicKey};
 
-#[derive(PartialEq, Debug, Clone)]
+#[derive(PartialEq, Debug, Clone, Deserialize, Serialize)]
 pub struct EncryptedValue {
     pub garbage: String,
 }
 
-#[derive(PartialEq, Debug)]
+#[derive(PartialEq, Debug, Deserialize, Serialize)]
```
Parth, line 23: `DecryptedValue` shouldn't get `Serialize`d anywhere

core/src/service/file_metadata_service.rs
```diff
@@ -1,140 +1,392 @@
...
+        FileRepoError(repo::file_repo::Error),
         MetadataRepoError(repo::file_metadata_repo::Error),
+        SerderError(serde_json::error::Error),
+        FileCreateError(file_encryption_service::FileCreationError),
     }
 }
 
 pub trait FileMetadataService {
-    fn update(db: &Db) -> Result<Vec<FileMetadata>, Error>;
+    fn sync(db: &Db) -> Result<Vec<FileMetadata>, Error>;
```
Parth, line 30: Don't group `Error`'s like this. If these return different `Error`s expose that to the caller

core/src/service/file_metadata_service.rs
```diff
@@ -1,140 +1,392 @@
...
+        debug(format!("Remote Updates {:?}", updates_remote));
 
-        updates
-            .into_iter()
-            .for_each(|meta| match FileMetadataDb::insert(&db, &meta) {
-                Ok(_) => {}
-                Err(err) => {
-                    error(format!("Insert Error {:?}", err));
+        // Create all the "new" files
+        let mut all_files = FileMetadataDb::get_all(&db)?;
```
Parth, line 75: I'm sure this could be done without `mut`. Keeping track of all the changes to `all_files` could be hard to reason about. Surely there's some way to do something like:
```
all_files
.filter
.for_each
...
```

core/src/service/file_metadata_service.rs
```diff
@@ -1,140 +1,392 @@
...
+        all_files.retain(|file| file.status == Status::New);
```
Parth, line 76: `all_files` is no longer what this should be called

core/src/repo/file_metadata_repo.rs
```diff
@@ -4,40 +4,58 @@ use serde_json;
 use sled;
 
 use crate::error_enum;
-use crate::model::file_metadata::FileMetadata;
+use crate::model::file_metadata::{FileMetadata, Status};
 use sled::Db;
+use uuid::Uuid;
 
 error_enum! {
     enum Error {
```
Parth, line 12: group errors bad

core/src/service/file_metadata_service.rs
```diff
@@ -1,140 +1,392 @@
...
+                        status: Status::Synced,
+                    },
+                )?)
             });
+        let errors_remote = updates_remote_res
+            .into_iter()
+            .filter_map(Result::err)
+            .collect::<Vec<Error>>();
+        error(format!("Remote Errors: {:?}", errors_remote));
+        Ok(FileMetadataDb::get_all(&db)?)
```
Parth, line 159: I think this overall strategy is good, and conducive to being called periodically. I think we could do a bit of localized cleanup.

Areas of improvement:
Once the client's are back to returning their specific error messages, `update`'s error scenarios is the type of thing we'd want to handle.

They're pretty well "documented/coded" by @tvanderstad, we should probably match on each api call and handle the failures we can. For example:

`SendFailed` let's retry? 
`EditConflict` there are diffs to resolve. 

But I think the overall structure is good, this is all the happy paths, and I'm okay with adding the failure condition in subsequent PR's. We might be well served by doing some work in `docs` and thinking out all the paths.

core/src/service/crypto_service.rs
raayan: So how does this go over FFI then?

core/src/service/file_metadata_service.rs
raayan: Yeah this is a large PR ☝️ is exactly what I said was to be worked on in subsequent PRs

core/src/client/mod.rs
raayan: Unsure what you mean

core/src/service/crypto_service.rs
raayan: I don't see any benefit to no deserializing/serializing it when `.secret` is essentially the same thing. Being able to say we don't `Serialize` `DecryptedValue` doesn't actually _mean_ anything

core/src/repo/file_metadata_repo.rs
raayan: unhelpful comment

core/src/service/file_metadata_service.rs
raayan: subsequent PR

PR #62, opened 2020-05-10
Updating Reqwest and some Catalyst Cruft

raayan:
Updating the reqwest version since the old version we were using doesn't let us build for catalyst. Also why use an old version?
Beginnings of getting catalyst to work.


core/Cargo.toml
```diff
@@ -8,19 +8,22 @@ license = "BSD-3-Clause"
...
+crate-type = ["lib", "staticlib"]
 
 [dependencies]
-reqwest = "0.9.24"
+reqwest = { version = "0.10.4", features = ["blocking", "json"] }
 base64 = "0.11.0"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0.44"
 uuid = { version = "0.8.1", features = ["v4"] }
-sled = { version = "0.31.0" }
```
Parth, line 19: Will we always need this change?

core/Cargo.toml
```diff
@@ -8,19 +8,22 @@ license = "BSD-3-Clause"
...
+sled = { git = "https://github.com/spacejam/sled", branch = "master" }
 tempfile = { version = "3.1.0" }
 rsa = { version = "^0.2", features = ["serde1"] }
 rand = "0.7"
 sha2 = "0.8.0"
 aead = "0.2.0"
 aes-gcm = "0.5.0"
 termion = "1.5.5"
+
+[profile.release]
```
Parth, line 28: What do

core/Xargo.toml
```diff
@@ -0,0 +1,4 @@
+std = { features = ["jemalloc"] }
```
Parth, line 1: What do

core/build_for_xcode.sh
```diff
@@ -0,0 +1,77 @@
...
+BUILD_CATALYST=1
+DO_COPY=1
+
+for arg in "$@"
+do
+    case $arg in
+        -f|--fat)
+        BUILD_FAT=1
+        BUILD_CATALYST=0
+        echo "Only building the fatty"
```
Parth, line 13: Only building fat binary.

core/Cargo.toml
raayan: No once it's resolved and he bumps the version it goes back, I'll add a TODO

core/Xargo.toml
raayan: It sets the memory allocator to `jemalloc`

PR #63, opened 2020-05-10
Apple ecosystem CI

raayan:


Parth:
```
	 cd core
        cargo install cargo-lipo
        rustup target add aarch64-apple-ios x86_64-apple-ios
        ./build_for_xcode.sh -f
        cd ../clients/ios
        xcodebuild -project ios_client.xcodeproj -allowProvisioningUpdates 
```
this is the only bit that's going to survive out of this PR.

see `core/Makefile` for whoever picks this up.

Issue #47, opened 2020-04-27
Enforce username rules


Parth:
Valid usernames consist of only a-z (lowercase), 0-9. 

1. Enforce this rule at the api level.
2. Write integration tests that check these rules.

PR #64, opened 2020-05-11
Enforce Username Rules

SmailBarkouch:
Adds case sensitivity for usernames and adds an alphanumeric check (Close: #47).


integration_tests/tests/new_account.rs
```diff
@@ -62,3 +62,71 @@ fn test_new_account_duplicate() {
...
+        Err(TestError::NewAccountError(new_account::Error::API(
+            NewAccountError::UsernameTaken
+        )))
+    );
+}
+
+fn new_account_alphanumeric_username() -> Result<(), TestError> {
+    let account = generate_account();
+
+    client::new_account::send(
```
tvanderstad, line 103: This test doesn't need to make two accounts, I was just being clever with the last one. The idea was that if I made two accounts that were the same the second one would give me a `NewAccountError::UsernameTaken`, so I made one with lowercase and one with uppercase and checked for that error

integration_tests/tests/new_account.rs
```diff
@@ -62,3 +62,71 @@ fn test_new_account_duplicate() {
...
+            username: account.username.clone(),
+            auth: AuthServiceImpl::<ClockImpl, RsaImpl>::generate_auth(&account).unwrap(),
+            public_key: serde_json::to_string(&account.keys.to_public_key()).unwrap(),
+        },
+    )?;
+
+    client::new_account::send(
+        api_loc(),
+        &NewAccountRequest {
+            username: "Smail&$@(".to_string(),
```
tvanderstad, line 115: We could probably use a couple more test cases for extra certainty and documentation of the behavior. Try an emoji, try a chinese character, whatever else you think of, the tests are pretty easy to write. It might be a good idea to make the username string a parameter of this function so that `test_new_account_alphanumeric_username()` below can look like
```
#[test]
fn test_new_account_alphanumeric_username() {
    assert_matches!(
        new_account_alphanumeric_username("Smail&$@"),
        Err(TestError::NewAccountError(new_account::Error::API(
            NewAccountError::InvalidUsername
        )))
    );
    assert_matches!(
        new_account_alphanumeric_username("漢字"),
        Err(TestError::NewAccountError(new_account::Error::API(
            NewAccountError::InvalidUsername
        )))
    );
    ...
}
```

core/src/model/api.rs
```diff
@@ -160,6 +163,7 @@ pub enum NewAccountError {
     InvalidAuth,
     ExpiredAuth,
     UsernameTaken,
+    InvalidUsername,
```
tvanderstad, line 166: I think this is actually the only place we need to add this error. If someone passes an invalid username to CreateFile, that username won't be found because we wouldn't have allowed it to be created, and we'd return UserNotFound

SmailBarkouch: ah okay

integration_tests/tests/change_file_content.rs
```diff
@@ -57,9 +56,52 @@ fn test_change_file_content() {
     assert_matches!(change_file_content(), Ok(_));
 }
 
-fn change_file_content_file_not_found() -> Result<(), TestError> {
+fn change_file_content_case_insensitive_username() -> Result<(), TestError> {
```
Parth, line 59: You seem to have added the same handful of tests to a handful of flows, I think it would be more meaningful and less noisy to have a set of well defined tests around new accounts, and then assume (rightfully) that usernames can't change.

server/src/index_db/create_file.rs
```diff
@@ -46,11 +46,10 @@ pub async fn create_file(
     file_path: &String,
 ) -> Result<i64, Error> {
     let version = generate_version(client).await?;
-
     client.execute("
 INSERT INTO files (file_id, file_name, file_path, username, file_content_version, file_metadata_version, deleted)
 VALUES ($1, $2, $3, $4, $5, $6, $7);
-", &[file_id, file_name, file_path, username, &version, &version, &false]).await?;
+", &[file_id, file_name, file_path, &username.to_lowercase(), &version, &version, &false]).await?;
```
Parth, line 52: I think this sort of business logic should be the caller's responsibility, we don't want these `repo` methods to get arbitrarily complicated

server/src/services/change_file_content.rs
```diff
@@ -9,6 +9,9 @@ pub async fn handle(
     server_state: &mut ServerState,
     request: ChangeFileContentRequest,
 ) -> Result<ChangeFileContentResponse, ChangeFileContentError> {
+    if !request.username.chars().all(|x| x.is_digit(36)) {
```
Parth, line 12: You just need to make sure no one can insert an invalid username, then you don't need to check it ever again. If the username is not in the db, then it's invalid. If it's in the db it must have been valid.

integration_tests/tests/change_file_content.rs
tvanderstad: We discussed this, it turns out Postgres will error if you even just check if a strange string exists in a table, so such a case would result in an internal error. I preferred it this way; what do you think?

server/src/services/mod.rs
```diff
@@ -6,3 +6,7 @@ pub mod get_updates;
 pub mod move_file;
 pub mod new_account;
 pub mod rename_file;
+
+pub fn username_is_valid(username: String) -> bool {
```
Parth, line 10: You should borrow here, then callers don't need to `clone`. If you really need to mutate internally you can `clone` yourself. But instead you could probably write this in a way that doesn't require you to create a new string.

PR #66, opened 2020-05-14
Rewrite of client-server interface

tvanderstad:
- swaps Rocket for Hyper, the fastest http server, which Rocket is built on anyway. This reduces our dependencies and lets us use rust stable instead of nightly.
- creates a common set of requests, responses, and errors in lockbook-core that are shared into lockbook-server and have simple, consistent serialization and deserialization


server/src/index_db/generate_version.rs
```diff
@@ -13,12 +13,13 @@ impl From<PostgresError> for Error {
     }
 }
 
-pub fn generate_version(client: &mut PostgresClient) -> Result<i64, Error> {
+pub async fn generate_version(client: &mut PostgresClient) -> Result<i64, Error> {
     let version = client
         .query_one(
             "SELECT CAST(EXTRACT(EPOCH FROM NOW()) * 1000000 AS BIGINT);",
```
tvanderstad, line 19: should be 1000 for milliseconds (or update documentation)

server/src/index_db/generate_version.rs
```diff
@@ -13,12 +13,13 @@ impl From<PostgresError> for Error {
     }
 }
 
-pub fn generate_version(client: &mut PostgresClient) -> Result<i64, Error> {
```
tvanderstad, line 16: This logic should be part of a database transaction. Potential race condition: read time for edit, read time for updates, get updates, write edit at past time (never gets sent in update poll)

core/src/model/api.rs
```diff
@@ -0,0 +1,185 @@
...
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
+pub struct GetUpdatesRequest {
+    pub username: String,
+    pub auth: String,
+    pub since_version: u64,
+}
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
+pub struct GetUpdatesResponse {
```
tvanderstad, line 101: add timestamp of update

PR #68, opened 2020-05-17
Standard Logger

raayan:
removing @parth's logger for the standard logger


tvanderstad:
what are the benefits?

raayan:
@tvanderstad less code and it's the ergonomic way to do it 
Our dependencies use the standard log interface so having multiple way to go about "logging" makes things tedious

PR #80, opened 2020-05-29
Android Client Infrastructure

Parth:
+ Cleaned up docs a bit
+ Added build instructions for `android` and `ios`
+ Created a `Makefile` for `core` targets
+ `Dockerfile` for building `android` and performing a smoke test
+ Github actions wired up to check android.
+ Setup `reqwests` in a way that doesn't require `openssl` anymore :100:


Parth:
closes #83

Issue #96, opened 2020-05-31
Working Concurrency Model


tvanderstad:
The state of concurrency is bad. If a client pulls a file, tells the server to update the file's contents, the server updates the version, then another client pulls, tells the server to update, and the server updates the version, when the server write the file content it will only have one of the changes. Additionally, in this scenario and in general because of S3 eventually consistent reads after updates, users may not have updated file contents when they think they do.

After syncing with @Parth offline, the new model will be to create a new S3 file for each version of the file. The key for the file will be `file_id:version`. Old versions will be deleted after new versions are written and the server will log an error if the deletion fails. Clients will read files at the version expected, and if the version in Postgres is ahead of the version in S3, the S3 read will fail and the client will wait until it succeeds (at which point S3 and Postgres are consistent). +1 for immutability in distributed systems.

PR #97, opened 2020-05-31
Endpoint transactions

tvanderstad:
Each endpoint runs in one transaction; pattern that lets us manage consistency guarantees more easily. This is pre-work for #96 and #98.

### DIFF
--- a/server/src/services/change_file_content.rs
+++ b/server/src/services/change_file_content.rs
@@ -9,8 +9,16 @@ pub async fn handle(
     server_state: &mut ServerState,
     request: ChangeFileContentRequest,
 ) -> Result<ChangeFileContentResponse, ChangeFileContentError> {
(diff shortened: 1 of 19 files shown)


server/src/services/change_file_content.rs
```diff
@@ -9,8 +9,16 @@ pub async fn handle(
     server_state: &mut ServerState,
     request: ChangeFileContentRequest,
 ) -> Result<ChangeFileContentResponse, ChangeFileContentError> {
+    let transaction = match server_state.index_db_client.transaction().await {
+        Ok(t) => t,
+        Err(e) => {
+            println!("Internal server error! Cannot begin transaction: {:?}", e);
```
Parth, line 15: You have `env_logger` on your path, these should probably all be `error!`s

PR #100, opened 2020-06-02
New Concurrency Model

tvanderstad:
S3 files named with content versions. No file updates, all creates and deletes. Client never gets out-of-date file thinking it's complete, and therefore never updates a file overwriting un-pulled changes.


server/src/index_db/update_file_metadata_and_content_version.rs
```diff
@@ -13,7 +13,7 @@ pub async fn update_file_metadata_and_content_version(
     transaction: &Transaction<'_>,
     file_id: &String,
     old_metadata_version: u64,
-) -> Result<u64, Error> {
+) -> Result<(u64, u64), Error> {
```
Parth, line 16: Maybe slap some `typedefs` on these bad bois in future PR's

PR #102, opened 2020-06-03
Verify public key and private key

### LINKED ISSUE #18: Verify public key on account creation server side
When we create an account, core sends to server a username and a public key.

Currently we check that this username is not taken. 

We will soon also check that the username conforms to valid rules (#47).

We should also check that this public key is a valid public key. Closely related to (#41)

### LINKED ISSUE #41: account import improvements
~Take an `account_string`, not a `user_name` and `key_string`.~

Currently importing an account simply inserts it into `CoreDB`. We should consider doing the following:
+ In crypto expose key validation, upon successful `RSAPrivateKey` serialization, validate that this is a key.
+ Verify this is the correct public key for this username #39 
+ Consider interactions with #40

SmailBarkouch:
Verify public key server side (close #18) and verify private key coredb side (close #41)


core/src/repo/account_repo.rs
```diff
@@ -8,6 +9,7 @@ error_enum! {
         SledError(sled::Error),
         SerdeError(serde_json::Error),
         AccountMissing(()),
+        InvalidPrivateKey(RSAError),
```
Parth, line 12: Why is this here?

clients/cli/src/utils.rs
```diff
@@ -41,6 +41,9 @@ pub fn get_account(db: &Db) -> Account {
                 "No account found, run init, import or help. Error: {:?}",
                 err
             ),
+            Error::InvalidPrivateKey(err) => {
```
Parth, line 44: It's only here because it's in sled, but I don't think it needs to be in sled in the first place

PR #140, opened 2020-07-07
Sync integration tests and bugfixes

Parth:
Added integration tests that that exercised:
+ Creating an account
+ Exporting an account string
+ Importing an account string
+ Syncing early account state
+ Creating files and folders and syncing
+ Editing files and folders and syncing

The cool thing about these tests is that I assert various repo level things must be exactly the same post syncs (all files in db1 == all files in db2). Let's me catch all sorts of small errors that would get missed if you were just making sure `id`s or file contents are equal.

Also added some `dev_utils` and added the command I used to free up several hundred gigabytes on our compute node and locally.

### DIFF
--- a/core/src/client.rs
+++ b/core/src/client.rs
@@ -132,6 +132,7 @@ impl Client for ClientImpl {
             .map_err(Error::SendFailed)?;
         let status = response.status().as_u16();
         let response_body = response.text().map_err(Error::ReceiveFailed)?;
+        println!("MINIO RESPONSE: {} {:?}", status, &response_body);
         let encrypted_file: Document = Document {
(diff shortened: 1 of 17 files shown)


core/src/client.rs
```diff
@@ -132,6 +132,7 @@ impl Client for ClientImpl {
             .map_err(Error::SendFailed)?;
         let status = response.status().as_u16();
         let response_body = response.text().map_err(Error::ReceiveFailed)?;
+        println!("MINIO RESPONSE: {} {:?}", status, &response_body);
```
tvanderstad, line 135: womp

PR #141, opened 2020-07-07
iOS Files and Folders

raayan:
A lot of good changes to `lib.rs` and many UI updates/re-organizing that should be good for moving forward.


core/src/lib.rs
```diff
@@ -73,34 +78,78 @@ pub type DefaultFileService = FileServiceImpl<
...
+                match result {
+                    Ok(value) => Value {
+                        success: CString::new(json!(value).to_string()).unwrap().into_raw(),
+                    },
+                    Err(err) => Value {
+                        error: CString::new(format!("{:?}", err)).unwrap().into_raw(),
+                    },
+                }
+            },
+            error: LockbookError::Database,
```
Parth, line 115: Is this supposed to be a match or something?

raayan: No not yet that was just me making an enum

PR #142, opened 2020-07-08
Server integration tests

tvanderstad:
Updates server integration tests for files and folders so they match the coverage we had before


integration_tests/src/new_account_tests.rs
```diff
@@ -0,0 +1,159 @@
...
+                rsa_key(&account.keys.to_public_key(), &folder_key)
+            ),
+            Err(Error::<NewAccountError>::Api(
+                NewAccountError::InvalidUsername
+            ))
+        );
+    }
+
+    // #[test]
+    // fn new_account_invalid_public_key() {
```
Parth, line 101: ?

tvanderstad: these don't pass yet bc we don't check signatures

PR #144, opened 2020-07-10
Clients get documents from server instead of directly from S3

tvanderstad:


server/src/file_service.rs
```diff
@@ -305,6 +305,59 @@ pub async fn rename_document(
...
+            );
+            GetDocumentError::InternalError
+        }
+    })?;
+    if status.deleted {
+        return Err(GetDocumentError::DocumentDeleted);
+    }
+    if status.content_version != request.content_version {
+        return Err(GetDocumentError::StaleVersion);
+    }
```
tvanderstad, line 336: deletes in S3 are eventually consistent so we check postgres to see if the document is deleted or out of date - because we never edit objects in S3, these are kind of both just checking if the file is supposed to already be deleted, but distinguishing the cases for potential user-friendliness

PR #168, opened 2020-07-28
Android client

### LINKED ISSUE #84: list files in android client

### LINKED ISSUE #85: view / edit files in android client

### LINKED ISSUE #86: sync in android client

### LINKED ISSUE #107: Export accounts using QR codes android
Depends on #114

+ Within settings -> create an `Export keys to another device`
+ Reach into `core` and grab an `account_string` from `AccountService`
+ Create a QR code with this string. Use [this](https://github.com/kenglxn/QRGen) library to create QR codes.
+ The screen should show the QR code at the top, and then the account string below it, with a button that allows you to copy the `account_string` to your clipboard.

### LINKED ISSUE #114: Expose a settings activity
+ In `ListFiles` top right expose an "overflow menu" 
image
+ In here have one of the menu items be "Settings"

### LINKED ISSUE #115: Protect keys with biometrics
+ In the screen for creating a new lockbook, have a checkbox that says `"Protect this key using fingerprint (recommended)"`. Have it checked by default.
+ Once they hit the continue button, do the call into core first, if it succeeds, complete the fingerprint registration flow.
+ TBD if there needs to be a pin backup.
+ In settings create an activity for the following settings:

1. Disabled - ignore the fingerprint
2. Protect key export with fingerprint (recommended) - only require fingerprint if they're going to export (#107) the key.
3. Use fingerprint every time you open the app (strict mode).

### LINKED ISSUE #117: Enhance account import for the android client
Depends on #88 

+ Currently account import only takes an account string.
+ Add a button with the QR icon above the account string `EditText`
+ Once tapped it should launch a sub activity
+ Use [this library](https://github.com/dm77/barcodescanner) to scan a QR code.

### LINKED ISSUE #118: On android, once a user is onboarded, don't let them see welcome again
Once a user has either created a new lockbook or imported an existing account while they're still using the app for the first time they can hit back and potentially totally lose access to their account (if they generate a new key for example). That username is gone forever.

Do this once they make it to this point.

Consider interactions with #117.

### LINKED ISSUE #130: Android list files
Call into core list files in a folder hierarchy

### LINKED ISSUE #131: Add android floating action button
Add a floating action button that allows you to create a document

SmailBarkouch:
Android client for Lockbook

Added android api interface based off lib.rs
Added floating action button (close #131 )
Displays files and folders (close #130 , close #84 )
Once the user is logged in, they stay logged in (close #118 )
Enhanced account import with qr code (close #117 )
Added sync through refresh (close #86 )
Files are able to be edited (close #85 )
Added biometrics for security (close #115 )
Added settings (close #114 )
Added account string exporting with qr code or account string (close #107 )
Added an activity to display info about a file
Added the ability to make a file
Added renaming functionality
Added a splash screen for those with biometrics
Added a deletion button (though it doesnt work since it isn't implemented yet in core)


clients/android/app/src/main/java/app/lockbook/loggedin/listfiles/ClickInterface.kt
```diff
@@ -0,0 +1,7 @@
+package app.lockbook.loggedin.listfiles
+
+interface ClickInterface {
+    fun onItemClick(position: Int)
+    fun onLongClick(position: Int) {
```
Parth, line 5: is there a linter / fmter we can enable for this project?

Parth:
let's move the "Protect your key with bio" to settings

core/src/lib.rs
```diff
@@ -91,9 +95,18 @@ pub type DefaultFileService = FileServiceImpl<
     DefaultFileEncryptionService,
 >;
 
+fn serialize_to_jstring<U: Serialize>(env: &JNIEnv, result: U) -> jstring {
```
Parth, line 98: This probably doesn't need to be in here

Parth: only used in one file, and it's not this file

core/src/lib.rs
```diff
@@ -91,9 +95,18 @@ pub type DefaultFileService = FileServiceImpl<
...
+    let serialized_result =
+        serde_json::to_string(&result).expect("Couldn't serialize result into result string!");
+    env.new_string(serialized_result)
+        .expect("Couldn't create JString from rust string!")
```
Parth, line 102: is this avoidable?

core/src/lib.rs
```diff
@@ -360,11 +374,48 @@ pub fn get_root(config: &Config) -> Result<FileMetadata, GetRootError> {
...
+    UnexpectedError(String),
+}
+
+pub fn get_children(config: &Config, id: Uuid) -> Result<Vec<FileMetadata>, GetChildrenError> {
+    let db = connect_to_db(&config).map_err(GetChildrenError::UnexpectedError)?;
+
+    match DefaultFileMetadataRepo::get_children(&db, id) {
+        Ok(file_metadata_list) => Ok(file_metadata_list),
+        Err(err) => Err(GetChildrenError::UnexpectedError(format!(
+            "DA FUDGE {:#?}",
```
Parth, line 392: probably can remove, see other examples

core/src/lib.rs
SmailBarkouch: In a way it is and it isn't. If I want to not use expects, I would have to get rid of this function and put this in every place it is used, which is a lot. Every JNI or Serde error is handled through this function. Could be other ways to simplify it and not have to use expects

SmailBarkouch: But these also kinda need to work, or else I can't return anything.

clients/android/app/src/main/java/app/lockbook/InitialLaunchFigureOuter.kt
```diff
@@ -1,25 +1,127 @@
...
         }
     }
-}
\ No newline at end of file
+
+    private fun isBiometricsOptionsAvailable(): Boolean =
+        BiometricManager.from(applicationContext)
+            .canAuthenticate() == BiometricManager.BIOMETRIC_SUCCESS
+
+    private fun checkBiometricOptions(pref: SharedPreferences) {
```
Parth, line 54: a name like `checkBiometricOptions` is very similar to `isBiometricsOptionsAvailable`. You probably want to call this something like `performBiometricFlow`

clients/android/app/src/main/java/app/lockbook/loggedin/listfiles/ClickInterface.kt
```diff
@@ -0,0 +1,6 @@
+package app.lockbook.loggedin.listfiles
+
+interface ClickInterface {
+    fun onItemClick(position: Int)
+    fun onLongClick(position: Int) {}
```
Parth, line 5: Why are these different?

clients/android/app/src/main/java/app/lockbook/loggedin/mainscreen/MainScreenActivity.kt
```diff
@@ -0,0 +1,43 @@
...
+
+import android.content.Intent
+import android.os.Bundle
+import android.view.Menu
+import android.view.MenuItem
+import androidx.appcompat.app.AppCompatActivity
+import app.lockbook.R
+import app.lockbook.loggedin.settings.SettingsActivity
+
+class MainScreenActivity : AppCompatActivity() {
```
Parth, line 11: This should probably be called `ListFilesActivity` (or something else) `Main` is too ambiguous

clients/android/app/src/main/java/app/lockbook/loggedin/mainscreen/MainScreenFragment.kt
```diff
@@ -0,0 +1,138 @@
...
+        val adapter = FilesAdapter(mainScreenViewModel)
+
+        binding.mainScreenViewModel = mainScreenViewModel
+        binding.filesList.adapter = adapter
+        binding.filesList.layoutManager = LinearLayoutManager(context)
+        binding.lifecycleOwner = this
+
+        binding.filesListRefresh.setOnRefreshListener {
+            mainScreenViewModel.refreshFiles()
+            mainScreenViewModel.sync()
```
Parth, line 51: 1. Is this in the wrong order? Shouldn't we `refreshFiles` after `sync`?
2. This seems to happen on the `ui` thread, doesn't it?

clients/android/app/src/main/java/app/lockbook/loggedin/mainscreen/MainScreenFragment.kt
```diff
@@ -0,0 +1,138 @@
...
+import app.lockbook.loggedin.popupinfo.PopUpInfoActivity
+import app.lockbook.loggedin.texteditor.TextEditorActivity
+import app.lockbook.utils.FileMetadata
+import app.lockbook.utils.RequestResultCodes.NEW_FILE_REQUEST_CODE
+import app.lockbook.utils.RequestResultCodes.POP_UP_INFO_REQUEST_CODE
+import app.lockbook.utils.RequestResultCodes.TEXT_EDITOR_REQUEST_CODE
+
+class MainScreenFragment : Fragment() {
+
+    lateinit var mainScreenViewModel: MainScreenViewModel
```
Parth, line 27: Intellij says this could be private

clients/android/app/src/main/java/app/lockbook/loggedin/mainscreen/MainScreenViewModelFactory.kt
```diff
@@ -0,0 +1,15 @@
...
+import androidx.lifecycle.ViewModelProvider
+
+class MainScreenViewModelFactory(
+    private val path: String
+) : ViewModelProvider.Factory {
+    @Suppress("unchecked_cast")
+    override fun <T : ViewModel?> create(modelClass: Class<T>): T {
+        if (modelClass.isAssignableFrom(MainScreenViewModel::class.java))
+            return MainScreenViewModel(path) as T
+        throw IllegalArgumentException("Unknown ViewModel class")
```
Parth, line 13: What's happening here?

clients/android/app/src/main/java/app/lockbook/loggedin/newfile/NewFileActivity.kt
```diff
@@ -0,0 +1,47 @@
...
+        uiScope.launch {
+            withContext(Dispatchers.IO) {
+                val json = Klaxon()
+                val fileType = if (file_radio_button.isChecked) {
+                    json.toJsonString(FileType.Document)
+                } else {
+                    json.toJsonString(FileType.Folder)
+                }
+
+                intent.putExtra("fileType", fileType)
```
Parth, line 36: Why did this become JSON? Can't you `putExtra` objects?

clients/android/app/src/main/java/app/lockbook/loggedin/settings/SettingsActivity.kt
```diff
@@ -0,0 +1,349 @@
...
+                    .setDeviceCredentialAllowed(true)
+                    .build()
+
+                biometricPrompt.authenticate(promptInfo)
+            }
+            BIOMETRIC_NONE -> navigateToBiometricSettings(checkedItem)
+        }
+    }
+
+    private fun checkBiometricOptionsCopy(accountString: String) {
```
Parth, line 159: There's too much duplication going on here

clients/android/app/src/main/java/app/lockbook/loggedin/settings/SettingsAdapter.kt
```diff
@@ -0,0 +1,47 @@
...
+        return SettingsViewHolder(view)
+    }
+
+    override fun getItemCount(): Int = settings.size
+
+    override fun onBindViewHolder(holder: SettingsViewHolder, position: Int) {
+        val item = settings[position]
+        holder.cardView.setting_title.text = item
+
+        if (holder.cardView.visibility == View.GONE && !item.contains("Bio")) {
```
Parth, line 32: Instead of making this disappear for users without bio, let's show the options disabled

clients/android/app/src/main/java/app/lockbook/loggedin/listfiles/ClickInterface.kt
SmailBarkouch: onLongClick has an empty body because if I do that it is no longer required to be overrided. The reason I need this is because I use this interface for both settings and files, and in settings I only want to register clicks, but in files I want to register both clicks since I handle them differently. Since onLongClick doesn't need to be overrided, I don't have to handle it's click for settings.

clients/android/app/src/main/java/app/lockbook/loggedin/mainscreen/MainScreenViewModelFactory.kt
SmailBarkouch: This is boilerplate code that is needed to create an class extended by ViewModel. I use this in MainScreenActivity to create the viewmodel

Parth:
https://github.com/lockbook/lockbook/issues/86 refers to a background task related to sync -- you should remove his from your description

clients/android/app/src/main/java/app/lockbook/loggedin/newfile/NewFileActivity.kt
SmailBarkouch: I can putExtra objects, but in this case I could just keep it as a string because I am never going to deserialize it.

clients/android/app/src/main/java/app/lockbook/loggedin/listfiles/ClickInterface.kt
Parth: What's the benefit of re-using the interface? Seems like you could have some unexpected behavior down the line?

PR #171, opened 2020-07-30
Make AccountService more robust

### LINKED ISSUE #35: Account Creation Improvements
+ Model the successful creation of an account
+ Don't create an account if one already exists
+ Provide a way to clear out an account

### LINKED ISSUE #41: account import improvements

Parth:
+ Closes #35 
+ Also @raayan and @SmailBarkouch have both run into issues where they have imported an account not on a server (they thought they were pointed at QA but they were pointed at PROD or vice versa). This also looks up their public_key and verifies a. they are on the server, b. they have the right public_key (might as well check if we have both values) 
+ Also @SmailBarkouch created deployment scripts here


core/src/repo/document_repo.rs
```diff
@@ -66,9 +66,7 @@ mod unit_tests {
             },
         };
 
-        let config = Config {
-            writeable_path: "ignored".to_string(),
-        };
+        let config = dummy_config();
```
raayan, line 69: 🔥

PR #208, opened 2020-08-11
iOS Tipping Up v9000 and Other Upgrade

raayan:
Tipping up the Swift JSON (de)serializers and preferring a less noisy FFI interaction.


core/src/c_interface.rs
```diff
@@ -160,6 +161,17 @@ pub unsafe extern "C" fn get_file_by_path(
     ))
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn get_file_by_id(
```
Parth, line 165: Identical to the function below it

PR #219, opened 2020-08-16
Android sync settings

### LINKED ISSUE #217: Add visible incremental sync

### LINKED ISSUE #218: Add settings associated with sync
Be able to disable periodic background sync or change the duration of periods between it
Be able to automatically incremental sync after editing
...

SmailBarkouch:
Added incremental sync (close #217  )
Added incremental and background sync settings (close #218 )


clients/android/app/src/main/java/app/lockbook/utils/LogUtil.kt
```diff
@@ -0,0 +1,21 @@
+package app.lockbook.utils
+
+import java.io.BufferedReader
+import java.io.InputStreamReader
+
+class LogUtil {
```
Parth, line 6: What is this

clients/android/app/src/main/java/app/lockbook/utils/Structures.kt
```diff
@@ -51,21 +51,27 @@ data class DecryptedValue(
 
 data class Account(
     val username: String,
-    val keys: String
+    val keys: RSAPrivateKey
+)
+
+data class RSAPrivateKey(
```
Parth, line 57: I am concerned about overflow rsa people are using this library to represent these values: https://docs.rs/num-bigint/0.3.0/num_bigint/

I'm wondering if you even need to serialize this value or if you could just leave it as a json string or something

clients/android/app/src/main/java/app/lockbook/utils/LogUtil.kt
SmailBarkouch: This is a class I created to get logs out of android if we need them. Right now it is unused, but if in future branches we wanted to get the logs we could use this.

clients/android/app/src/main/java/app/lockbook/utils/Structures.kt
SmailBarkouch: I don't know if I could leave just that particular part in a json string, but this should work. This the recommended format that Klaxon wanted me to use, and I am unsure why you believe it can't represent these large values. It just splits the values, doesn't it? I assume it still works since I can execute work successfully

PR #243, opened 2020-08-24
File Backed Logging

### LINKED ISSUE #121: Improve client logging
+ We're using env_logger which allows you to filter out logs based on module and direct to various backends
+ Create a ~sled~ file backed logger
+ Expose something in core that allows clients to log using core logger

raayan:
Logging now goes to `output.log` and std out. It also has colors!

Set the env var `LOCKBOOK_DEBUG` to anything to have `DEBUG` level logs printed to standard out

Closes #121

### DIFF
--- a/core/Makefile
+++ b/core/Makefile
@@ -42,6 +42,10 @@ apple:
 	@cp target/universal/release/liblockbook_core.a ${libs_apple}
 
 
+.PHONY: qa
+qa:
+	API_URL="http://qa.lockbook.app:8000" cargo build
+
 MIN_VER = 22
 jniLibs = '../clients/android/core/src/main/jniLibs'
 libName = 'liblockbook_core.so'
--- a/core/Makefile
+++ b/core/Makefile
@@ -42,10 +42,6 @@ apple:
 	@cp target/universal/release/liblockbook_core.a ${libs_apple}
 
 
-.PHONY: qa
-qa:
-	API_URL="http://qa.lockbook.app:8000" cargo build
-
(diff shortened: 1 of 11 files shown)


core/Makefile
```diff
@@ -42,6 +42,10 @@ apple:
 	@cp target/universal/release/liblockbook_core.a ${libs_apple}
 
 
+.PHONY: qa
+qa:
```
Parth, line 46: I would say move this to `dev_utlis`, but also if you're gonna automate this, you should add `touch lib.rs` otherwise someone is gonna have a bad time

raayan: oh yeah lol I forgot to not put this here

Parth: I've been a fan of adding really specific scripts that configure settings for me `./build_ios_core_pointed_at_qa.sh` and `./build_ios_core_pointed_at_prod.sh` for example

raayan: Yeah it's nbd, removed

Parth:
@SmailBarkouch read and understand this pr, mainly `loggers.rs`. All the core logs are now going to end up in a particular file.

+ You should create an activity that reads this file and shows the content
+ You should be able to access this file regardless of the state of core, so imo surround `InitialLaunchFigureOuter`'s call into core in a `try { } catch ( ) { }` if it fails launch the activity. Otherwise from settings at the bottom expose something that allows you to "View Logs" (which would launch the activity).

raayan:
Push me to the premium build plz

PR #245, opened 2020-08-25
Fixing Logger

raayan:
Fixes the logger so it tries to make sure the writeable path folder exists before writing a file in there.
set `LOCKBOOK_NO_COLOR=1`  to turn off colors. 

Touches up c library usage.

### DIFF
--- a/clients/android/core/src/main/java/app/lockbook/core/core.kt
+++ b/clients/android/core/src/main/java/app/lockbook/core/core.kt
@@ -1,6 +1,6 @@
 package app.lockbook.core
 
-external fun initLogger()
+external fun initLogger(config: String)
 external fun createAccount(config: String, username: String): String
 external fun importAccount(config: String, account: String): String
 external fun exportAccount(config: String): String
@@ -22,5 +22,6 @@ external fun executeSyncWork(config: String, account: String, workUnit: String):
 
 fun loadLockbookCore() {
     System.loadLibrary("lockbook_core")
-    initLogger()
+    // TODO: @smail put the real config here
+    initLogger("JUNK_CONFIG")
 }
--- a/clients/android/core/src/main/java/app/lockbook/core/core.kt
+++ b/clients/android/core/src/main/java/app/lockbook/core/core.kt
@@ -1,6 +1,6 @@
 package app.lockbook.core
 
-external fun initLogger(config: String)
+external fun initLogger()
 external fun createAccount(config: String, username: String): String
 external fun importAccount(config: String, account: String): String
 external fun exportAccount(config: String): String
@@ -22,6 +22,5 @@ external fun executeSyncWork(config: String, account: String, workUnit: String):
 
 fun loadLockbookCore() {
     System.loadLibrary("lockbook_core")
-    // TODO: @smail put the real config here
-    initLogger("JUNK_CONFIG")
+    initLogger()
 }
--- a/clients/apple/lockbook.xcodeproj/xcshareddata/xcschemes/Lockbook-iOS.xcscheme
+++ b/clients/apple/lockbook.xcodeproj/xcshareddata/xcschemes/Lockbook-iOS.xcscheme
@@ -52,13 +52,13 @@
       </BuildableProductRunnable>
       <EnvironmentVariables>
          <EnvironmentVariable
-            key = "RUST_BACKTRACE"
+            key = "LOCKBOOK_DEBUG"
             value = "1"
             isEnabled = "YES">
          </EnvironmentVariable>
          <EnvironmentVariable
-            key = "RUST_LOG"
-            value = "lockbook_core=debug"
+            key = "LOCKBOOK_NO_COLOR"
+            value = "1"
             isEnabled = "YES">
          </EnvironmentVariable>
       </EnvironmentVariables>
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -1,8 +1,11 @@
+mod java_interface;
+
 #[macro_use]
 extern crate log;
 extern crate reqwest;
 
 use crate::client::{ClientImpl, Error};
+use crate::loggers::LoggersError;
 use crate::model::account::Account;
 use crate::model::api::{GetPublicKeyError, NewAccountError};
 use crate::model::crypto::DecryptedValue;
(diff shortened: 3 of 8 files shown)


core/src/lib.rs
```diff
@@ -99,12 +103,22 @@ pub type DefaultFileService = FileServiceImpl<
...
     Unexpected(String),
 }
 
 pub fn init_logger(log_path: &Path) -> Result<(), InitLoggerError> {
-    loggers::init(log_path, env::var("LOCKBOOK_DEBUG").is_ok())
-        .map_err(|err| InitLoggerError::Unexpected(format!("{:#?}", err)))
+    let print_debug = env::var("LOCKBOOK_DEBUG").is_ok();
+    let print_colors = env::var("LOCKBOOK_NO_COLOR").is_err();
+    match loggers::init(log_path, print_debug, print_colors) {
+        Ok(_) => Ok(()),
```
raayan, line 114: HOT PATH!

core/src/lib.rs
```diff
@@ -99,12 +103,22 @@ pub type DefaultFileService = FileServiceImpl<
 
 #[derive(Debug, Serialize)]
 pub enum InitLoggerError {
+    OnTouch(String),
```
Parth, line 106: It's unclear what this means

clients/android/core/src/main/java/app/lockbook/core/core.kt
```diff
@@ -22,5 +22,6 @@ external fun executeSyncWork(config: String, account: String, workUnit: String):
 
 fun loadLockbookCore() {
     System.loadLibrary("lockbook_core")
-    initLogger()
+    // TODO: @smail put the real config here
```
Parth, line 25: You can just remove this line from both sides and revert changes to `java_interface.rs` we can have @SmailBarkouch deal with this in a different PR

core/src/lib.rs
```diff
@@ -99,12 +103,22 @@ pub type DefaultFileService = FileServiceImpl<
...
 }
 
 pub fn init_logger(log_path: &Path) -> Result<(), InitLoggerError> {
-    loggers::init(log_path, env::var("LOCKBOOK_DEBUG").is_ok())
-        .map_err(|err| InitLoggerError::Unexpected(format!("{:#?}", err)))
+    let print_debug = env::var("LOCKBOOK_DEBUG").is_ok();
+    let print_colors = env::var("LOCKBOOK_NO_COLOR").is_err();
+    match loggers::init(log_path, print_debug, print_colors) {
+        Ok(_) => Ok(()),
+        Err(LoggersError::File(_)) => match create_dir_all(log_path) {
```
Parth, line 115: I'm not sure this is the best way to handle this error, you should just check if the folder exists and create it if not.

core/src/lib.rs
Parth: `LockbookFolderCreationError`

Parth: `lib.rs` is *the* contract for `core` described in rust, so let's be verbose here

core/src/lib.rs
```diff
@@ -99,12 +103,22 @@ pub type DefaultFileService = FileServiceImpl<
...
 pub enum InitLoggerError {
+    OnTouch(String),
     Unexpected(String),
 }
 
 pub fn init_logger(log_path: &Path) -> Result<(), InitLoggerError> {
-    loggers::init(log_path, env::var("LOCKBOOK_DEBUG").is_ok())
-        .map_err(|err| InitLoggerError::Unexpected(format!("{:#?}", err)))
+    let print_debug = env::var("LOCKBOOK_DEBUG").is_ok();
+    let print_colors = env::var("LOCKBOOK_NO_COLOR").is_err();
```
Parth, line 112: Do you personally want this? I can't think of a use case where this is useful, especially because we have the file logging going. I'd rather just not have unless you can justify it's existence.

core/src/lib.rs
Parth: But realistically, this is an `UnexpectedError` there's no corrective action we want a user to have to take in this situation

core/src/lib.rs
raayan: Xcode logs don't support color but they do stdout

raayan: That does now make me thing I can tail the file logs for color, but still I'm not going to completely debilitate my ability to use the in editor logs

core/src/lib.rs
raayan: There is some difference in performance:
```
raayanpillai@raayan-mbp ~/dev/lockbook/clients/cli [raayan/fix_clogger] $ rm -r ~/.lockbook
raayanpillai@raayan-mbp ~/dev/lockbook/clients/cli [raayan/fix_clogger] $ time lockbook list
No account! Run init or import to get started!
lockbook list  0.01s user 0.01s system 21% cpu 0.128 total
raayanpillai@raayan-mbp ~/dev/lockbook/clients/cli [raayan/fix_clogger] $ time lockbook list
No account! Run init or import to get started!
lockbook list  0.01s user 0.01s system 47% cpu 0.054 total
```

Also I put this here https://github.com/lockbook/lockbook/pull/245/files/bc16e4ac41b8ab38793835b0773ed315d80b72f4#r476084369 for the sake of discussion, I wasn't planning to die on this hill

clients/apple/lockbook.xcodeproj/xcshareddata/xcschemes/Lockbook-iOS.xcscheme
```diff
@@ -52,13 +52,13 @@
...
          <EnvironmentVariable
-            key = "RUST_BACKTRACE"
+            key = "LOCKBOOK_DEBUG"
             value = "1"
             isEnabled = "YES">
          </EnvironmentVariable>
          <EnvironmentVariable
-            key = "RUST_LOG"
-            value = "lockbook_core=debug"
+            key = "LOCKBOOK_NO_COLOR"
```
raayan, line 60: Also you can see usage of env var here @Parth

Parth: Ah right xcode got it

Issue #189, opened 2020-08-02
Instances of CI flakiness


Parth:
+ https://github.com/lockbook/lockbook/pull/162#partial-pull-merging
+ https://github.com/lockbook/lockbook/runs/863962395#step:5:155
+ https://github.com/lockbook/lockbook/runs/936662458
+ https://github.com/lockbook/lockbook/pull/180/checks?check_run_id=952145048#step:5:289

PR #246, opened 2020-08-26
Android smoke tests

### LINKED ISSUE #119: Add android smoke tests
Currently android uses it's own kotlin `Core.kt` to negotiate the core lib.

Even though it looks like it might, it doesn't check that contract is fulfilled at compile time.

(I'm pretty sure) it doesn't even check at `System.loadLib` time.

It only checks when a specific function is invoked, and then if it fails it fails at runtime.

Add a series of "SmokeTests" one for each `external func`. They don't need to succeed from the rust point of view (it's okay if they return an integer that signals error) because that means it found and ran the function.

SmailBarkouch:
Added tests for each endpoint in kotlin (close #119 )
Added kotlin-interface-tests github actions with its respective docker components
Added a way to view logs in android (you are required to be in debuging mode to view it in settings)
Optimized android docker and android code


clients/android/app/src/test/java/app/lockbook/ExecuteWorkTest.kt
```diff
@@ -0,0 +1,64 @@
...
+        coreModel.insertFile(document).component1()!!
+        val folder =
+            coreModel.createFile(generateAlphaString(), Klaxon().toJsonString(FileType.Folder))
+                .component1()!!
+        coreModel.insertFile(folder).component1()!!
+        val syncWork = coreModel.calculateFileSyncWork().component1()!!
+        for (workUnit in syncWork.work_units) {
+            coreModel.executeFileSyncWork(coreModel.getAccount().component1()!!, workUnit)
+                .component1()!!
+        }
```
Parth, line 54: Should this test not require something?

clients/android/app/src/test/java/app/lockbook/GetAccountTest.kt
```diff
@@ -0,0 +1,55 @@
...
+    @After
+    fun resetDirectory() {
+        Runtime.getRuntime().exec("rm -rf $path")
+    }
+
+    @Test
+    fun getAccountOk() {
+        val coreModel = CoreModel(Config(path))
+        CoreModel.generateAccount(Config(path), generateAlphaString()).component1()!!
+        coreModel.getAccount().component1()!!
```
Parth, line 38: Shouldn't this `require` something

containers/Dockerfile.kotlin_tests
```diff
@@ -0,0 +1,33 @@
...
+ENV DEBIAN_FRONTEND=noninteractive
+ENV ANDROID_HOME /opt/android-sdk-linux
+RUN apt update && apt install -y build-essential openjdk-8-jdk wget curl unzip
+
+# Install rust things
+RUN curl https://sh.rustup.rs -sSf | bash -s -- -y
+ENV PATH="/root/.cargo/bin:${PATH}"
+
+# Install android things
+RUN wget -q https://dl.google.com/android/repository/android-ndk-r21c-linux-x86_64.zip -O android-ndk-r21c-linux-x86_64.zip
```
Parth, line 11: You don't need to install android things

core/Makefile
```diff
@@ -6,11 +6,13 @@ android:
...
-	@rm -rf $(jniLibs)
+	@rm -rf $(jniLibs)/arm64-v8a
+	@rm -rf $(jniLibs)/armebi-v7a
+	@rm -rf $(jniLibs)/x86
+	@rm -rf $(jniLibs)/x86_64
 
 	@echo "Moving new .so's over"
-	@mkdir $(jniLibs)
-	@mkdir $(jniLibs)/arm64-v8a
+	@mkdir -p $(jniLibs)/arm64-v8a
```
Parth, line 15: explain this change?

core/Makefile
```diff
@@ -20,6 +22,45 @@ android:
...
+	@touch src/lib.rs
+	@cargo build --lib --release
+	
+	@echo "Cleaning up old .dylib if they exist"
+	@rm -rf $(jniLibs)/desktop
+	
+	@echo "Moving new .dylib over"
+	@mkdir -p $(jniLibs)/desktop
+
+	@mv target/release/$(macosLibName) $(jniLibs)/desktop/$(macosLibName)
```
Parth, line 36: Any reason you prefer `mv` over `cp`? Still thinking about which I like more

Parth:
Closes #119

core/Makefile
SmailBarkouch: I use mv since I can take that library and not have cargo rebuild everything the next time I call the command. So I just thought I might as well just take it rather than copy it over.

core/Makefile
SmailBarkouch: The -p means it will create the sub directories of the folder you are trying to create if they don't already exist. So I can replace those two commands with just that.

core/Makefile
raayan: I imagine `mv` would make cargo rebuild since the target folder has been modified

core/Makefile
raayan: `-p` is good for things like this

Parth:
This should fix #189 as well

PR #264, opened 2020-09-05
Server reconnects to Postgres if connection closed (Fixes #81)

### LINKED ISSUE #81: Have server model complexity associated with dropped / no connections
Two issues:

1. Currently servers just die if they can't connect. 
2. uring their lifespan they don't notice dropped connections until it causes a failure.

tvanderstad:


raayan (approved):
LGTM

PR #326, opened 2020-09-29
API_URL at runtime and Apple Usage Meter

raayan:
Now when you run the corelib binaries you want `API_URL` to be defined in your environment. This allows for a configurable endpoint and removes unnecessary compilation. We could also pass this in by function parameter but that would require a larger code touch


raayan:
Redoing this whole thing

PR #329, opened 2020-10-06
Android launch with core

### LINKED ISSUE #210: Better handle nullable types
ex. getExtra returns a nullable type, handle it more appropriately

### LINKED ISSUE #276: Android app should use core in InitialLaunchFigureOuter
`InitialLaunchFigureOuter` should not use a shared preference to determine core status. It should use the new `get_db_state` method in `lib.rs`. This method tells you whether the user's state is:
+ `Empty` -> Show welcome screen
+ `MigrationRequired` -> Call `migrate_db` (this updates any local state). You should show a popup with a spinner that says "Migrating local state..." in a dialog while this happens. When this is done you can call `get_db_state` again.
+ `StateRequiresCleaning` -> the db is busted, you have to tell the user to uninstall and re sync. We will do everything in our ability  to prevent this from ever happening. But should we need to, the client should be prepared.

SmailBarkouch:
Now you are determined to be logged in with the core. You're state can correctly be identified. (close #276 )
Better handled nullables (close #210 )

Parth, I changed the snackbar message for the migration based on the review, the screenshot is old thats why it doesn't show.


Parth (changes requested):

clients/android/app/src/main/java/app/lockbook/loggedin/listfiles/FileModel.kt
```diff
@@ -27,28 +29,31 @@ class FileModel(path: String) {
...
+                is CalculateWorkError.NoAccount -> {
                     Timber.e("No account.")
                     _errorHasOccurred.postValue("Error! No account!")
                 }
-                is CoreError.CouldNotReachServer -> {
+                is CalculateWorkError.CouldNotReachServer -> {
                     Timber.e("Could not reach server despite being online.")
                     _errorHasOccurred.postValue(
-                        UNEXPECTED_ERROR_OCCURRED
+                        UNEXPECTED_CLIENT_ERROR
```
SmailBarkouch, line 46: Shouldn't be using that constant.

SmailBarkouch: Get rid of internet checks, just let them know they have no internet when they try to do something that does.

clients/android/app/src/main/java/app/lockbook/loggedin/listfiles/FileModel.kt
```diff
@@ -27,28 +29,31 @@ class FileModel(path: String) {
...
                     )
                 }
-                is CoreError.Unexpected -> {
+                is CalculateWorkError.Unexpected -> {
                     Timber.e("Unable to calculate syncWork: ${error.error}")
-                    _errorHasOccurred.postValue(
-                        UNEXPECTED_ERROR_OCCURRED
+                    _unexpectedErrorHasOccurred.postValue(
+                        UNEXPECTED_ERROR
```
SmailBarkouch, line 52: Shouldn't be posting the constant.

clients/android/app/src/main/java/app/lockbook/utils/Converters.kt
```diff
@@ -6,13 +6,100 @@ import com.beust.klaxon.Klaxon
...
+        okTag -> {
+            val ok = jv.obj?.string("content")
+            if (ok != null) {
+                Ok(
+                    when (ok) {
+                        State.ReadyToUse.name -> State.ReadyToUse
+                        State.Empty.name -> State.Empty
+                        State.MigrationRequired.name -> State.MigrationRequired
+                        State.StateRequiresClearing.name -> State.StateRequiresClearing
+                        else -> {}
```
SmailBarkouch, line 49: This should not be empty

clients/android/app/src/main/java/app/lockbook/utils/Converters.kt
```diff
@@ -22,9 +109,35 @@ val createAccountConverter = object : Converter {
...
+            uiErrorTag -> {
+                val error = jv.obj?.obj("content")?.string("content")
+                if (error != null) {
+                    Err(
+                        when (error) {
+                            CreateAccountError.UsernameTaken::class.simpleName -> CreateAccountError.UsernameTaken
+                            CreateAccountError.InvalidUsername::class.simpleName -> CreateAccountError.InvalidUsername
+                            CreateAccountError.CouldNotReachServer::class.simpleName -> CreateAccountError.CouldNotReachServer
+                            CreateAccountError.AccountExistsAlready::class.simpleName -> CreateAccountError.AccountExistsAlready
+                            else -> CreateAccountError.Unexpected("Can't recognize UiError content.")
```
SmailBarkouch, line 123: Make error more specific enough to allow a person to find out where exactly it happened.

PR #332, opened 2020-10-07
API Tied to Account

raayan:
Since accounts really only exist in the context of a specific server (or no server, of course), we should store the destination API url with the account itself. This will enable multi-account down the road and make testing _much_ easier; you'll never be in a situation where the program you have compiled/installed cannot communicate with the server that the current account is associated with.


clients/cli/src/new_account.rs
```diff
@@ -16,9 +16,12 @@ pub fn new_account() {
         .expect("Failed to read from stdin");
     username.retain(|c| c != '\n');
 
+    let api_location =
+        env::var("API_URL").unwrap_or_else(|_| lockbook_core::DEFAULT_API_LOCATION.to_string());
+
     println!("Generating keys and checking for username availability...");
 
-    match create_account(&get_config(), &username) {
+    match create_account(&get_config(), username.as_str(), api_location.as_str()) {
```
Parth, line 24:
```suggestion
    match create_account(&get_config(), &username, &api_location) {
```

PR #336, opened 2020-10-11
Better JSON and Error Types

raayan:


core/src/lib.rs
```diff
@@ -112,24 +108,71 @@ pub type DefaultFileService = FileServiceImpl<
     DefaultFileEncryptionService,
 >;
 
+macro_rules! impl_get_variants {
```
Parth, line 111: Maybe worth dropping some comments here explaining what's going on. Also let's move this to the bottom of this file.

core/src/lib.rs
```diff
@@ -112,24 +108,71 @@ pub type DefaultFileService = FileServiceImpl<
...
+}
+
+#[derive(Debug, Serialize, EnumIter)]
 pub enum InitLoggerError {
-    UnexpectedError(String),
+    Stub, // TODO: Enums should not be empty
 }
 
-pub fn init_logger(log_path: &Path) -> Result<(), InitLoggerError> {
+pub fn init_logger(log_path: &Path) -> Result<(), Error<InitLoggerError>> {
```
Parth, line 163: Could this be an `Error<()>`?

core/src/lib.rs
raayan: sure thing

core/src/lib.rs
raayan: yes we can

PR #350, opened 2020-10-19
Rework Crypto Model

### LINKED ISSUE #69: Improve Type System For Crypto (Generify That Bad Boy)
`-` @raayan

tvanderstad:
Fixes #69, #365 
* Documents are now `Vec<u8>` (ffi still takes strings)
* Encrypted/Signed things are wrapped with `AESEncrypted<T>`, `RSAEncrypted<T>`, and `RSASigned<T>`
* `AuthService` is gone; requests themselves will be signed so what little logic was left was folded into `CryptoService`
* Client interface rewritten, takes request objects and private key, adds timestamp, signs, and adds core version
* Integration tests rewritten to use new client interface and new test helpers
* Factors common error conditions for requests (invalid auth, client update required, internal error, etc)
* Makes server treat public key as authoritative source of identity (it's needed for auth anyway; better to lookup username than double-check that requests username and pk don't match)
* Makes time `i64` instead of `u64` so there's no error condition for when the user's clock is set before 1970
* Makes all rust tests use `--release` which accelerates key generation dramatically, cuts total time by several minutes
* Relates requests to their responses, errors, http methods, and endpoints using associated types and associated constants, which simplifies use of the core's api client
* Documents are automatically compressed

Todo:
- [x] fix CLI @Parth 
- [x] fix linux client @steverusso 
- [x] fix c# ffi @tvanderstad 
- [x] fix kotlin ffi @SmailBarkouch 
- [x] fix swift ffi @raayan 
- [x] fix server @tvanderstad 
- [x] fix performance tests @raayan


clients/linux/src/backend.rs
```diff
@@ -220,14 +218,14 @@ impl LbCore {
     pub fn open(&self, id: &Uuid) -> Result<(FileMetadata, String), String> {
         match self.file_by_id(*id) {
             Ok(meta) => match self.read(meta.id) {
-                Ok(decrypted) => Ok((meta, decrypted.secret)),
+                Ok(decrypted) => Ok((meta, String::from_utf8(decrypted).unwrap())),
```
tvanderstad, line 221: @steverusso FYI documents are not necessarily UTF-8 because we support binary, use `String::from_utf8_lossy` to fill in invalid UTF characters with those weird question marks in boxes or whatever and not crash

steverusso: Ah ok, thanks. Fixed :+1:

raayan (approved):

PR #389, opened 2020-11-05
cli: Refactor calculate_usage code.

steverusso:
* The units are in 1000s, changed to 1024s.
* The variable names represent a _single_ unit, so the plurality was dropped.
* A tuple is used to reduce the match statement by only getting the necessary parts and calculating afterwards instead of calculating and printing each in place.

### DIFF
--- a/clients/cli/src/calculate_usage.rs
+++ b/clients/cli/src/calculate_usage.rs
@@ -5,46 +5,40 @@ use crate::UNEXPECTED_ERROR;
 use lockbook_core::{get_usage, Error as CoreError, GetUsageError};
 
 const BYTE: u64 = 1;
-const KILOBYTES: u64 = BYTE * 1000;
-const MEGABYTES: u64 = KILOBYTES * 1000;
-const GIGABYTES: u64 = MEGABYTES * 1000;
-const TERABYTES: u64 = GIGABYTES * 1000;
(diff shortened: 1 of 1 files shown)


Parth:
I had it this way initially, but apparently by powers of two are called `kibibyte` and powers of 10 are called `kilobyte`: https://en.wikipedia.org/wiki/Kilobyte

image

If this is wrong and other CLI's in linux do something different then I could be convinced to do it that way.

steverusso:
Wow, I did not know this...

In the man page of `du`, I found the following:
> The  SIZE  argument  is  an  integer and optional unit (example: 10K is 10*1024).  Units are K,M,G,T,P,E,Z,Y (powers of 1024) or KB,MB,... (powers of 1000).  Binary prefixes can be used, too: KiB=K, MiB=M, and so on.

So, the issue right now for this PR is do we want to either:
1. adjust the unit names (from `KB` to `KiB` for example) and keep the size as `1024`
2. adjust the unit sizes (from `1024` to `1000`) and keep the names.

`du` seems to default to using `1000`, and something like `ls` seems to default to using `1024`. Your call.

Parth:
I vote:
> adjust the unit sizes (from 1024 to 1000) and keep the names.
but we can ask standup

PR #414, opened 2020-11-21
Core and Server Folder Errors

### LINKED ISSUE #401: Move a folder into itself puts file tree in a bad state

### LINKED ISSUE #402: Prevent users from deleting their root folder

### LINKED ISSUE #428: lockbook import-private-key message typo
If you use `lockbook import-private-key` incorrectly or leave it blank, it will output a message:
```
To import an existing Lockbook, pipe your account string into this command, eg. pbpaste | lockbook import or xclip -selection clipboard -o | lockbook import
```
The message doesn't take into account the new command name, `lockbook import-private-key`

SmailBarkouch:
This PR is centered around adding core and server errors for folders (mainly root).

Added 2 new errors for clients:
- `FolderMovedIntoItself` restricts a user from moving a folder into one of its children or itself
- `CannotDeleteRoot` restricts a user from deleting root

Added supplemental errors into core for FileService

Added 4 new errors to `api.rs` for server.
- `CannotMoveRoot`
- `CannotMoveIntoDescendant`
- `CannotRenameRoot`
- `CannotDeleteRoot`

Added a two new error to `FileError` in server
- `IllegalRootChange`
- `FolderMovedIntoDescendants`

Made supplemental integration tests for all new errors added.

Todo:
- [x] fix CLI @SmailBarkouch  
- [x] fix linux client @steverusso 
- [x] fix c# ffi @Parth  
- [x] fix kotlin ffi @SmailBarkouch 
- [x] fix swift ffi @raayan 
- [x] check delete and move SQL is right @tvanderstad 

fixes #401
fixes #402
fixes #428


Makefile
```diff
@@ -151,8 +151,16 @@ performance_bench_report: is_docker_running
 dev_stack_run: server
 	HASH=$(hash) docker-compose -f containers/docker-compose-integration-tests.yml --project-name=integration-tests-$(hash) down
 	HASH=$(hash) docker-compose -f containers/docker-compose-dev-stack.yml --project-name=dev-stack-$(hash) down
+	HASH=$(hash) docker-compose -f containers/docker-compose-semi-dev-stack.yml --project-name=dev-stack-$(hash) down
 	HASH=$(hash) docker-compose -f containers/docker-compose-dev-stack.yml --project-name=dev-stack-$(hash) up
 
+.PHONY: semi_dev_stack_run
```
Parth, line 157: there are ways to tell a docker-compose file to not start a service, or to start a particular image defined in the command. Investigate ways to remove this redundant (and hard to maintain docker-compose file). I think in the future we're going to consolidate all the compose files.

server/src/file_service.rs
```diff
@@ -33,10 +34,21 @@ pub async fn change_document_content(
...
 
     let (old_content_version, new_version) = result.map_err(|e| match e {
-        file_index_repo::FileError::DoesNotExist => ChangeDocumentContentError::DocumentNotFound,
-        file_index_repo::FileError::IncorrectOldVersion => ChangeDocumentContentError::EditConflict,
-        file_index_repo::FileError::Deleted => ChangeDocumentContentError::DocumentDeleted,
-        _ => {
+        FileError::DoesNotExist => ChangeDocumentContentError::DocumentNotFound,
+        FileError::IncorrectOldVersion => ChangeDocumentContentError::EditConflict,
+        FileError::Deleted => ChangeDocumentContentError::DocumentDeleted,
+        FileError::FolderMovedIntoDescendants
```
Parth, line 40: @tvanderstad / @SmailBarkouch are all of these supposed to be 500's? `ParentDoesNotExist` for example? Is there a way to model this so that `FileError`s aren't being grouped like this?

core/src/service/file_service.rs
```diff
@@ -526,6 +541,10 @@ impl<
             .map_err(DeleteFolderError::MetadataError)?
             .ok_or(DeleteFolderError::CouldNotFindFile)?;
 
+        if file_metadata.id == file_metadata.parent {
+            return Err(DeleteFolderError::CannotDeleteRoot);
+        }
+
         if file_metadata.file_type == Document {
             return Err(DeleteFolderError::DocumentTreatedAsFolder);
```
Parth, line 549: This is where you should add unit tests for the core side errors

Parth: You've tested server side stuff in integration tests (which is good)

server/src/file_service.rs
SmailBarkouch: I forgot to ask about this during stand-up, i'm confused on what you mean. Can you elaborate?

core/tests/move_file_tests.rs
```diff
@@ -259,4 +259,103 @@ mod move_document_tests {
...
+                },
+            ),
+            Err(ApiError::<MoveFolderError>::Api(
+                MoveFolderError::CannotMoveRoot
+            ))
+        );
+    }
+
+    #[test]
+    fn move_folder_into_itself() {
```
Parth, line 304: Can you also write a test that makes a folder structure like `root -> a -> b -> c` and then try moving `a` into `c`? I think it will fail

tvanderstad (approved):
✅

Makefile
```diff
@@ -151,8 +151,16 @@ performance_bench_report: is_docker_running
...
+semi_dev_stack_run: is_docker_running
+	HASH=$(hash) docker-compose -f containers/docker-compose-integration-tests.yml --project-name=integration-tests-$(hash) down
+	HASH=$(hash) docker-compose -f containers/docker-compose-dev-stack.yml --project-name=dev-stack-$(hash) down
```
raayan, line 160: Why are we downing these?

SmailBarkouch: I followed the same set up for `dev stack run`. I think the logic behind it is to make sure that these containers aren't running and thus blocking ports.

raayan: Something we'll be cleaning up with the tech-debt @Parth

PR #431, opened 2020-11-27
Modularize (that bad boy)

raayan:
Removes `server` dependency on `core` in lieu of new library, `models` that both `core` and `server` share

**Before**
```zsh
$ cat Cargo.lock | grep package | wc -l
    251
$ cargo build
   Compiling lockbook-server v0.1.0 (/Users/raayanpillai/dev/lockbook/server)
    Finished dev [unoptimized + debuginfo] target(s) in 2m 03s
```

**After**
```zsh
$ cat Cargo.lock | grep package | wc -l
    203
$ cargo build
   Compiling lockbook-server v0.1.0 (/Users/raayanpillai/dev/lockbook/server)
    Finished dev [unoptimized + debuginfo] target(s) in 1m 37s
```


raayan:
Will re-open after #350 lands

PR #453, opened 2020-12-13
Cleanup infra

### LINKED ISSUE #135: Make CI even better round 2
+ [ ] Are the integration tests being run in parallel?
+ [ ] Can we express things like `lint` `fmt` and `test` as build-time tasks rather than things that run within a docker container?
+ [ ] Crypto in the release build seems way faster than debug, can we use this at test time? Can we measure the difference?

Parth:
# Paying down infra tech debt

+ [x] Eliminate copy paste across `docker-compose.yml`'s
+ [ ] One touch partial & full dev stacks
+ [x] Do more things at `docker build` rather than `docker run` to benefit from docker caching
+ [x] Use multi-stage builds for things like lint, fmt and unit tests
+ [x] Use multi-stage builds for things like kotlin-core and csharp-core
+ [x] Eliminate the use of volumes
+ [x] Express dependencies correctly in github-actions
+ [x] Run all integration tests in parallel but against the same stack containers
+ [x] Eliminate server lint errors
+ [x] Reduce flakiness of auth
+ [x] Reduce flakiness of local_changes_repo
+ [x] Perform better garbage collection
+ [x] Be smarter about the dockerfile of `linux`, `server`, and `cli`

image

closes #135


Makefile
```diff
@@ -14,152 +17,150 @@ exorcise:
...
 .PHONY: cli
 cli: is_docker_running
-	docker build -f containers/Dockerfile.cli . --tag cli:$(hash)
+	docker build --target cli-build -f containers/Dockerfile.cli . --tag cli:$(hash) --build-arg HASH=$(hash)
 
 .PHONY: cli_fmt
 cli_fmt: cli
-	@echo The following files need formatting:
-	docker run cli:$(hash) cargo +stable fmt -- --check -l
+	@echo The following files need formatting:compute_usage
```
tvanderstad, line 63: what's this??

containers/Dockerfile.android
```diff
@@ -1,38 +1,29 @@
-# TODO to significantly speedup this build make it multistage
-
-FROM ubuntu:20.04
+ARG HASH
```
tvanderstad, line 1: does it affect build caching to put this here vs right before its used?

Parth: Nah it has to go here, doesn't effect the cache you can see when I fix the other comment that android is not rebuilt from scratch

Makefile
Parth: fixed

tvanderstad (approved):
✅

PR #457, opened 2020-12-15
cli: added auto save

### LINKED ISSUE #184: Have CLI detect incremental changes to a file being edited
means file doesn't need to be closed to be saved / synced

Figure out if you need to either
- pull the file to check for changes, and then update it in lockbook
- or have an event driven check for changes

SmailBarkouch:
CLI now has autosave.

Whenever a user wants to edit a new or existing file, it creates a temporary file in `/tmp/` (not sure if this is the case for everything, but it doesn't matter). When a user saves that file, whether it be automatically through vscode, or `:w` in vim, CLI will pickup those changes and save it to core. CLI watches the temporary file for changes every 5 seconds, and if there is a change it will save it.

fixes #184


clients/cli/src/edit.rs
```diff
@@ -121,3 +129,57 @@ pub fn edit(file_name: &str) {
         )
     });
 }
+
+pub fn save_file_to_core(
```
Parth, line 133: Let's move this to `utils.rs`. Consider renaming it to `save_temp_file_contents`.

clients/cli/src/utils.rs
```diff
@@ -183,3 +187,59 @@ pub fn print_last_successful_sync() {
         println!("Last successful sync: {}", duration);
     }
 }
+
+pub fn set_up_auto_save(
+    watch_file_metadata: FileMetadata,
+    watch_file_location: String,
+) -> Hotwatch {
+    let mut hot_watch = Hotwatch::new_with_custom_delay(core::time::Duration::from_secs(5))
+        .unwrap_or_else(|err| {
```
Parth, line 196: Wonder if we need to fail in this scenario? Maybe this can return an `Option<Hotwatch>` and `stop_auto_save` only gets called if it's not `None`

Parth (changes requested):
2 minor changes

Parth (approved):

PR #460, opened 2020-12-20
core: Refactor to use map_err where possible.

steverusso:


raayan (commented):
This looks fine and is cool by me but I vaguely remember having a conversation with @Parth about preferring `match` statements to `.map_err` or `.and_then`

Parth:
@raayan are you thinking of scala `match` being the only option when you have to `await`?

raayan:
@Parth no i remember when you said something against map_err and and_then specifically

PR #470, opened 2020-12-24
cli: support copying in folders

### LINKED ISSUE #320: Support copying folders into lockbook

SmailBarkouch:
fixes #320


clients/cli/src/copy.rs
```diff
@@ -24,75 +25,232 @@ pub fn copy(path: PathBuf, import_dest: &str, edit: bool) {
...
-        let content_to_import = fs::read_to_string(&path).unwrap_or_else(|err| {
-            exit_with(
-                &format!("Failed to read file: {}", err),
-                COULD_NOT_READ_OS_FILE,
-            )
-        });
+        copy_file(&path, import_dest, edit, false)
+    } else {
+        recursive_copy_folder(&path, import_dest, true);
+        exit_with(&format!("imported folder to: {}", import_dest), SUCCESS)
```
Parth, line 31: The norm with this sort of thing is that they just exit silently when they succeed, think about `mv` / `cp` from your system.

clients/cli/src/copy.rs
```diff
@@ -24,75 +25,232 @@ pub fn copy(path: PathBuf, import_dest: &str, edit: bool) {
...
-        } else {
-            import_dest.to_string()
-        };
+    if metadata.is_file() {
+        copy_file(&path, import_dest, false, true);
+    } else {
+        let children_paths: Vec<DirEntry> = fs::read_dir(path)
+            .unwrap_or_else(|err| {
+                exit_with(
+                    &format!("Failed to read folder children: {}", err),
```
Parth, line 49:
```suggestion
                    &format!("Unable to list children of folder: {}, error: {}", path, err), // something like this
```

clients/cli/src/copy.rs
```diff
@@ -24,75 +25,234 @@ pub fn copy(path: PathBuf, import_dest: &str, edit: bool) {
...
+                        }
+                    )
+                };
+
+                recursive_copy_folder(&child_path, &lockbook_child_path, false);
+            }
+        } else if let Err(err) = create_file_at_path(&get_config(), &import_dest) {
+            match err {
+                CoreError::UiError(CreateFileAtPathError::FileAlreadyExists) => {
+                    println!("Input destination {} not available within lockbook, use --edit to overwrite the contents of this file!", import_dest)
```
Parth, line 119: You say this here regarding `--edit`, but on line 43 (`copy_file(&path, import_dest, false, true)`) you pin the value of `edit` to `false`. Isn't this a contradiction?

You should probably carry that value forward.

clients/cli/src/copy.rs
```diff
@@ -24,75 +25,237 @@ pub fn copy(path: PathBuf, import_dest: &str, edit: bool) {
...
+                    None => exit_with(
+                        "Import target does not contain a file name!",
+                        UNEXPECTED_ERROR,
+                    ),
+                }
+            } else {
+                import_dest.to_string()
+            };
+
+            let file_metadata = match create_file_at_path(&get_config(), &import_dest_with_filename)
```
steverusso, line 160: Instead of calling `get_config` each time you need `Config`, this can be stored in a variable at the beginning of the function.

clients/cli/src/copy.rs
```diff
@@ -24,75 +25,237 @@ pub fn copy(path: PathBuf, import_dest: &str, edit: bool) {
...
+                    &format!("Failed to read file: {}", content_err),
+                    COULD_NOT_READ_OS_FILE,
+                )
+            }
+        }
+        (_, Err(path_err)) => {
+            if is_folder_copy {
+                println!(
+                    "Failed to get absolute path from {}, OS error: {}",
+                    path_err, path_err
```
steverusso, line 251: `path_err` printed for both slots where I think the first value should be `import_dest`?

steverusso (commented):
Nice job so far!

clients/cli/src/copy.rs
```diff
@@ -24,75 +25,237 @@ pub fn copy(path: PathBuf, import_dest: &str, edit: bool) {
...
+                        });
+
+                    format!(
+                        "{}{}",
+                        import_dest,
+                        if ends_with_slash {
+                            format!("{}/{}", parent_name, child_name)
+                        } else {
+                            format!("/{}/{}", parent_name, child_name)
+                        }
```
steverusso, line 100: This can be "simplified" with the following:
```                                                                                                                            
                    format!(                                                                                                                                                                                       
                        "{}{}{}/{}",                                                                                                                                                                                  
                        import_dest,                                                                                                                                                                               
                        if ends_with_slash { "" } else { "/" },                                                                                                                                                    
                        parent_name,
                        child_name                                                                                                                                                                       
                    )
```

clients/cli/src/copy.rs
```diff
@@ -24,75 +25,237 @@ pub fn copy(path: PathBuf, import_dest: &str, edit: bool) {
...
+                } else {
+                    format!(
+                        "{}{}",
+                        import_dest,
+                        if ends_with_slash {
+                            child_name.to_string()
+                        } else {
+                            format!("/{}", child_name)
+                        }
+                    )
```
steverusso, line 111: Likewise, this can be simplified with:
```
                    format!(                                                                                                                                                                                       
                        "{}{}{}",                                                                                                                                                                                    
                        import_dest,                                                                                                                                                                               
                        if ends_with_slash { "" } else { "/" },                                                                                                                                                    
                        child_name                                                                                                                                                                                 
                    )
```

steverusso (commented):
Just a couple more cleanups which now form two _very_ similar patterns.

clients/cli/src/copy.rs
```diff
@@ -8,10 +8,12 @@ use lockbook_core::{
```
steverusso, line 8: nitpick: Let's organize imports by grouping and in the following order:
1. std
2. third party libraries
3. lockbook_core
4. crate stuff

clients/cli/src/copy.rs
```diff
@@ -23,76 +25,243 @@ pub fn copy(path: PathBuf, import_dest: &str, edit: bool) {
...
+                        if ends_with_slash { "" } else { "/" },
+                        child_name
+                    )
+                };
+
+                recursive_copy_folder(&child_path, &lockbook_child_path, config, edit, false);
+            }
+        } else if let Err(err) = create_file_at_path(config, &import_dest) {
+            match err {
+                CoreError::UiError(CreateFileAtPathError::FileAlreadyExists) => {
```
steverusso, line 122: I think all of these match cases that specify `CoreError::UiError` throughout would be more readable as nested under a `UiError` match (as follows), but it's up to you.
```
match err {
    CoreError::UiError(err) => match err {
        CreateFileAtPathError::Whatever => ...,
        ...
    }
    CoreError::Unexpected(msg) => ...,
}
```

clients/cli/src/copy.rs
```diff
@@ -1,98 +1,235 @@
...
+        recursive_copy_folder(&path, &import_path, &config, edit);
+    }
+}
+
+fn recursive_copy_folder(path: &PathBuf, import_dest: &str, config: &Config, edit: bool) {
+    if path.is_file() {
+        match copy_file(&path, import_dest, config, edit) {
+            Ok(msg) => println!("{}", msg),
+            Err(err) => {
+                println!("{}", err.msg);
```
Parth, line 64:
```suggestion
                eprintln!("{}", err.msg);
```

Parth (approved):

PR #474, opened 2020-12-25
core: Create a macro for the 'Error::Unexpected(format!())' pattern.

steverusso:
The `Unexpected` variant of core's `Error` enum takes a `String`. In _every_ instance that `Error::Unexpected` is used, it is provided a `String` that is constructed via the `format!` macro, making the pattern `Error::Unexpected(format!(...))` appear quite often.

This PR creates and employs a macro with the same signature as the `format!` macro, but that returns the repetitive Unexpected error building pattern, so:
* `Error::Unexpected(format!(...))` is now just `unexpected!(...)`, and
* this still allows for a simple string literal to be passed as well: `unexpected!("error message")`

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -63,6 +63,12 @@ pub enum Error<U: Serialize> {
 }
 use Error::UiError;
 
+macro_rules! unexpected {
+    ($base:literal $(, $args:tt )*) => {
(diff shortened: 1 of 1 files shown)


raayan (approved):
Nice touch

PR #478, opened 2020-12-28
android: better drawing UI and erasing

### LINKED ISSUE #430: Make erasing more effective
Erasing should work when you make a fast stroke across the screen.

### LINKED ISSUE #495: Create a better UI for drawing
This new drawing screen should...
- always fullscreen
- have a toolbar off to the side with options for color, utensil, and size
- have the ability to tap away the toolbar

SmailBarkouch:
This PR gives the user a better UI for drawing and more effective erasing.

Now, all of the tools in the drawing activity are set towards the left, with options to change the color, size, and utensil. The activity now is fullscreen, and allows the user to tap the open space to hide the toolbar.

Photos:
Screenshot_20210103-143814_Lockbook

Screenshot_20210103-143820_Lockbook

Screenshot_20210103-143831_Lockbook

The the logic behind calculating if a stroke is erased, which is essentially determining if your eraser crosses a stroke, is much more accurate. I incorporate the logic previously used, which is helpful to determine if the eraser crosses two stroke points spread apart, with the new logic helping determine if two far apart eraser points crossed a stroke.

fixes #430 
fixes #495


Parth (changes requested):
It's nice that this toolbar fades in, can we also make it fade out?

Also the changes I asked for regarding file extensions, do you plan to do that on a separate branch?

Parth (approved):

PR #498, opened 2021-01-04
project: start a contributing guidelines file

steverusso:


Parth (approved):
This looks good, some non blocking questions / comments:

+ Do projects ever make this the place to be like "if you want to contribute to X platform see docs Y"?
+ Could this be a good place to try to codify who to tag on what type of review?
+ Maybe a good place to describe what our CI does and if you encounter X failure what it means? CI would probably need to be more reliable, though I wonder if the failures are already clear enough.

steverusso:
I think the first two would be great additions, not just by platform but even a bit more general (to include the other categories).

PR #502, opened 2021-01-05
core: expose human readable strings that described when we last synced and how many bytes we're using

### LINKED ISSUE #490: In core, expose get human duration for last synced

### LINKED ISSUE #500: In core, expose calculate usage (logic taken from CLI) to core

SmailBarkouch:
A readable form of `get_usage` and `get_last_synced` is now exposed for all clients to use through `api.rs`.

The name of these endpoints are:
- `calculate_usage`: returns the total amount of bytes being used with the highest measurement (like megabytes). You may also request the exact amount in bytes.
- `calculate_last_synced`: returns a user friendly amount of time since you last synced

Both of these endpoints return a String if successful (Result<String, ______>)

close #500
close #490


core/src/lib.rs
```diff
@@ -814,6 +816,18 @@ pub fn get_last_synced(config: &Config) -> Result<i64, Error<GetLastSyncedError>
     }
 }
 
+pub fn calculate_last_synced(config: &Config) -> Result<String, Error<GetLastSyncedError>> {
```
Parth, line 819: this should probably be `get_last_synced_human_string`

core/src/lib.rs
```diff
@@ -814,6 +816,18 @@ pub fn get_last_synced(config: &Config) -> Result<i64, Error<GetLastSyncedError>
...
+    let last_synced = get_last_synced(config)?;
+
+    if last_synced != 0 {
+        let duration = Duration::milliseconds(DefaultClock::get_time() - last_synced);
+        println!("{}, {}", DefaultClock::get_time(), last_synced);
```
Parth, line 824: no `println` s in core, only log messages. probably should just delete this line

core/src/lib.rs
```diff
@@ -844,6 +858,42 @@ pub fn get_usage(config: &Config) -> Result<Vec<FileUsage>, Error<GetUsageError>
...
+pub const MEGABYTE: u64 = KILOBYTE * 1000;
+pub const GIGABYTE: u64 = MEGABYTE * 1000;
+pub const TERABYTE: u64 = GIGABYTE * 1000;
+
+pub const KILOBYTE_PLUS_ONE: u64 = KILOBYTE + 1;
+pub const MEGABYTE_PLUS_ONE: u64 = MEGABYTE + 1;
+pub const GIGABYTE_PLUS_ONE: u64 = GIGABYTE + 1;
+pub const TERABYTE_PLUS_ONE: u64 = TERABYTE + 1;
+
+pub fn calculate_usage(config: &Config, exact: bool) -> Result<String, Error<GetUsageError>> {
```
Parth, line 872: You should move this out of this file, this file has a very narrow purpose. In fact the other call probably shouldn't be here either. I would create a `usage_service` and move the bodies of this function and the other usage function to those places.

Parth (changes requested):

Parth:
Can you split out the `android` component of this PR into a separate PR, the tag of this can remain `core` then. Also maybe make issues for each platform instead of checklist on this PR. 

Also try to improve the title of this PR. It's very vague.

clients/cli/src/utils.rs
```diff
@@ -173,17 +168,15 @@ pub fn edit_file_with_editor(file_location: &str) -> bool {
...
     if atty::is(atty::Stream::Stdout) {
-        let last_updated = get_last_synced(&get_config())
-            .expect("Failed to retrieve content from FileMetadataRepo");
-
-        let duration = if last_updated != 0 {
-            let duration = Duration::milliseconds(DefaultClock::get_time() - last_updated);
-            duration.format_human().to_string()
-        } else {
-            "never".to_string()
+        let last_updated = match get_last_synced_human_string(&get_config()) {
```
Parth, line 171: This can prob be a `map_err` + `exit_with`

Parth (approved):

PR #517, opened 2021-01-14
ci: be more aggressive about cleanup, let docker decide container names

Parth:
+ Clean up more types of docker artifacts
+ FIx typo that was causing integration tests to not run
+ That typo was introducing the "naming conflict" that was causing lots of flakiness
+ Eliminate flags and env vars that were providing no value


raayan (approved):

Parth:
Container names should not have been removed from this PR, the container names are how we get exit codes from the test runs. They have been re-added in #531

PR #519, opened 2021-01-14
cli: removing a folder will only show the number of documents being removed

### LINKED ISSUE #505: Only display the number of documents when confirming the removal of a folder in CLI

SmailBarkouch:
Removing a folder will not show the total number of folders and documents being removed, just the documents

fixes #505


Parth (changes requested):
Can you add a `--force` or `-f` flag that skips this check for automation purposes?

I believe you can look at the `get-usage` command for an example of how to do that.

Parth (approved):

PR #583, opened 2021-02-06
android: fix trailing thin lines in drawing

SmailBarkouch:
In this PR I fix a common problem when drawing. If you remove your pen quickly enough of the screen after a stroke, the difference in pressure between two points will cause a thin line to appear. This thin line will be much smaller than the lines before it. 

I fixed this with the help of @Parth, utilizing rolling averages (of **N = 5**) to smooth it out at the end.

### Before
<details>

Screenshot_20210205-213706_Lockbook

</details>

### After
<details>

Screenshot_20210205-213653_Lockbook

</details>


Parth (changes requested):
Can be simpler, faster, more flexible: https://stackoverflow.com/a/16757630/1060955 that is the easiest (an approximation), there are ways to be exact iirc you can explore those too if you find it interesting, approx is prob fine for us

Parth (approved):

PR #593, opened 2021-02-10
android: use metadata version in file descriptions to show when the file was updated synced

### LINKED ISSUE #592: Android is utilizing get human duration for each file and disregarding metadata version

SmailBarkouch:
This PR, I fix a previous error where I used get last synced human string, and instead utilize the metadata version for the date.

fixes #592

### DIFF
--- a/core/src/java_interface.rs
+++ b/core/src/java_interface.rs
@@ -411,7 +411,7 @@ pub extern "system" fn Java_app_lockbook_core_CoreKt_createFile(
 }
 
 #[no_mangle]
-pub extern "system" fn Java_app_lockbook_core_CoreKt_getLastUpdatedFile(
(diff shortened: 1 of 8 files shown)


core/src/java_interface.rs
```diff
@@ -407,6 +410,24 @@ pub extern "system" fn Java_app_lockbook_core_CoreKt_createFile(
     )
 }
 
+#[no_mangle]
+pub extern "system" fn Java_app_lockbook_core_CoreKt_getLastUpdated(
```
Parth, line 414: You should call this something like `convertToHumanDuration` or something like that

Parth (changes requested):

Parth (approved):

PR #605, opened 2021-02-27
android: faster erasing by caching bounds of strokes

SmailBarkouch:
In this PR, erasing is made faster by caching the bounds of strokes that are and have been drawn. These stroke bounds are used during erasing to see if erase stroke intersects it. If it does, we will do our normal procedures for checking if the line itself has been crossed, but if it doesn't it means that that there is no point doing the normal erasing check.


clients/android/app/src/main/java/app/lockbook/ui/HandwritingEditorView.kt
```diff
@@ -286,6 +344,7 @@ class HandwritingEditorView(context: Context, attributeSet: AttributeSet?) :
 
     private fun moveTo(point: PointF, pressure: Float) {
         lastPoint.set(point)
+        strokesBounds.add(RectF(point.x, point.y, point.x, point.y))
```
Parth, line 347: Doesn't this mean that the size of `strokeBounds` will be roughly equal to the size of `points`. Don't we want the size of `strokeBounds` to be equal to the number of strokes?

Parth (changes requested):

Parth (approved):

PR #606, opened 2021-02-28
windows: build for arm64

tvanderstad:


Parth:
Perhaps we want to do something like this:

```toml
[package]
name = "rustylib"
version = "0.1.0"
authors = ["Roberto Huertas <roberto.huertas@outlook.com>"]
edition = "2018"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[lib]
name = "rustylib"
# this is needed to build for iOS and Android.
crate-type = ["staticlib", "cdylib"]

# this dependency is only needed for Android.
[target.'cfg(target_os = "android")'.dependencies]
jni = { version = "0.13.1", default-features = false }
```

clients/windows/lockbook/lockbook.csproj.user
```diff
@@ -6,5 +6,30 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Debug|x64'">
```
Parth, line 6: prob not what we need?

server/Cargo.lock
```diff
@@ -150,10 +150,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f4ebf134c444f81a8f49f315792cede0fcfb489d8a434d14a60c4e06ae6e1624"
 dependencies = [
  "dirs",
- "reqwest",
+ "reqwest 0.10.6",
```
Parth, line 153: maybe you should `cargo build` across:
+ [ ] cli
+ [ ] linux
+ [ ] server

so that the lock files update in this PR rather than someone else's pr

Parth (changes requested):

Parth (approved):

PR #638, opened 2021-03-19
apple: periodic task refinements + bottom bar improvements

### LINKED ISSUE #602: Bottom bar efficiency
Clients are using `get-status` to infer what files are dirty. This utilizes the network by default, this is expensive on mobile phones. Waking up the radio is something we should do sparingly.

image

The ask: create `check_with_server` flag for get status. Can leave the existing code path for desktop and cli (no flag == true) or something like that. Phones should tune how often they check depending on OS recommendations.

### LINKED ISSUE #631: Bottom bar has hardcoded elements

Parth:
# App lifecycle refinements

+ [x] Sync automatically on app launch, app close, app re-open
+ [x] Sync every 30 minutes
+ [x] Reset sync timer while a document is being edited
+ [x] Have core expose a local only calculate work endpoint that can be strategically called after file operations
+ [x] Don't hardcode last synced

# Massively more energy efficient

Initially Upon opening the app (a sync is performed).

image

After about 30 seconds, while editing a document:

image

# issues

+ fixes #602 
+ fixes #631


raayan (approved):
SEND

PR #655, opened 2021-03-27
android: sync snackbar messages don't always appear

SmailBarkouch:
In this PR, I resolved a long standing error involving sync messages. Whenever a user would try to sync using the swipe to refresh functionality, the sync info message did not always appear. This was due android itself lacking a queue functionality. So when snackbars appeared in succession, like they did with syncing, then it did guarantee that the message would always be shown. Luckily, a library we already used called snackprogressbar (which can be forked in need be), implemented this functionality already. So I just utilized it and fixed the bug.

fixes: #567

### DIFF
--- a/clients/android/app/src/main/java/app/lockbook/model/AlertModel.kt
+++ b/clients/android/app/src/main/java/app/lockbook/model/AlertModel.kt
@@ -21,7 +21,7 @@ object AlertModel {
                 }
             })
         }
-        Timber.e("STEP 15")
+
         snackBar.show()
     }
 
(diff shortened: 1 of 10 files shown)


clients/android/app/src/main/java/app/lockbook/model/AlertModel.kt
```diff
@@ -6,11 +6,13 @@ import androidx.appcompat.app.AlertDialog
 import app.lockbook.R
 import app.lockbook.util.UNEXPECTED_ERROR
 import com.google.android.material.snackbar.Snackbar
+import timber.log.Timber
 
 object AlertModel {
     fun notify(view: View, msg: String, onFinishAlert: OnFinishAlert) {
+        Timber.e("STEP 13")
```
Parth, line 13: Rm

Parth (changes requested):

Parth (approved):

PR #663, opened 2021-03-31
android: fixed bad allocation error while drawing

### LINKED ISSUE #642: drawingview may crash on configuration changes due to bad allocation
```
A/OpenGLRenderer: Encountered EGL error 12291 EGL_BAD_ALLOC during rendering
A/libc: Fatal signal 6 (SIGABRT), code -1 (SI_QUEUE) in tid 3461 (RenderThread), pid 2476 (app.lockbook)
```

```
E/Surface: getSlotFromBufferLocked: unknown buffer: 0x754d271300
A/OpenGLRenderer: Encountered EGL error 12291 EGL_BAD_ALLOC during rendering
```

SmailBarkouch:
In this PR, I fix a long time issue involving the drawing view. If you keep rotating and getting in and out of the app while in the drawing screen, then you may crash due to this bug. This was due to the improper handling of the drawing thread whenever it was being ended. I ended up having to get rid of all the old thread logic and replaced it with something that is more concise and cleaner the long run. Its important to note, **I do not entirely know what caused this error**, this was only present on my tablet, my mobile device (Google Pixel 3) had nothing of the sort. What I did know is that it happened while the thread was closing and I must have improperly handled it, possibly in a way that only offended the tablet (the log messages were different too), that of which was solved after the new logic was implemented.

fixes #642

### DIFF
--- a/clients/android/app/src/main/java/app/lockbook/ui/DrawingView.kt
+++ b/clients/android/app/src/main/java/app/lockbook/ui/DrawingView.kt
@@ -20,6 +20,7 @@ import app.lockbook.util.ColorAlias
 import app.lockbook.util.Drawing
 import app.lockbook.util.Stroke
 import kotlinx.android.synthetic.main.activity_drawing.*
+import java.lang.IllegalStateException
 import java.util.*
 import kotlin.math.pow
 import kotlin.math.roundToInt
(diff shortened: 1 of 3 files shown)


clients/android/app/src/main/java/app/lockbook/ui/DrawingView.kt
```diff
@@ -567,6 +586,32 @@ class DrawingView(context: Context, attributeSet: AttributeSet?) :
             } finally {
                 holder.unlockCanvasAndPost(canvas)
             }
+
+            frameTime = (System.nanoTime() - frameStartTime) / 1000000L
+
+            if (frameTime < MIN_TIME_PER_FRAME) {
+                try {
+                    Thread.sleep(MIN_TIME_PER_FRAME - frameTime)
```
Parth, line 594: Thread.sleep is a code smell could you explain what you’re trying to do here?

Parth (changes requested):

clients/android/app/src/main/java/app/lockbook/ui/DrawingView.kt
SmailBarkouch: My bad, I forgot to mention that I limited the frames per second to 60 rather than letting it process as many frames as it wants. I did this just cause most devices wouldn't have any use for all the frames it makes, but this can easily be changed so I can max it out at 120 or something. I can also make a setting for this if you want me to and have it so the user can adjust it.

Parth (approved):

PR #666, opened 2021-04-03
linux: end user default settings and enable syntax highlighting

Parth:
+ By default the file tree opens without showing debugging related columns (id & type)
+ Use `SourceView`'s syntax highlighting based on file name:
image

### DIFF
--- a/clients/linux/src/account.rs
+++ b/clients/linux/src/account.rs
@@ -12,8 +12,8 @@ use gtk::{
     Stack as GtkStack, WrapMode as GtkWrapMode,
 };
 use sourceview::prelude::*;
-use sourceview::Buffer as GtkSourceViewBuffer;
 use sourceview::View as GtkSourceView;
+use sourceview::{Buffer as GtkSourceViewBuffer, LanguageManager};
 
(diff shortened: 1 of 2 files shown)


clients/linux/src/account.rs
```diff
@@ -364,19 +365,23 @@ impl Editor {
         let cntr = GtkScrolledWindow::new(None::<&GtkAdjustment>, None::<&GtkAdjustment>);
         cntr.add(&stack);
 
+        let highlighter = LanguageManager::new();
```
steverusso, line 368: possible nit: I think a good pattern for constructors is if the variable isn't used for something during initializing, just define in the return value:
```
Self {
    ....
    highlighter: LanguageManager::new(),
    ....
}
```

steverusso (changes requested):

clients/linux/src/account.rs
Parth: fixed in: 1119c4177b7c3f0e40f73f18da95eaeef2dbd2e0

steverusso (approved):

PR #678, opened 2021-04-06
server: PagerDuty Change on startup and Alert on error!

raayan:
Server will now POST a [PagerDuty Change](https://lockbook.pagerduty.com/change-events/01BUJK3HTO5HDVFZ5M5ESEP2JH) when the server starts up.

When an `error!` is hit, it will trigger a [PagerDuty Alert](https://lockbook.pagerduty.com/incidents/PDT5GNQ) (these are de-duped by a combination of `BUILDHASH-FILE-LINE`) 

When building `lockbook-server` if you pass `SERVER_BUILD` as an environment variable, the dist will be build with whatever you put in there (typically `$(git rev-parse --short HEAD)`) to tag the server with a build. (This has been added to the automated server dist).


server/src/loggers.rs
```diff
@@ -49,7 +64,143 @@ pub fn init(log_path: &Path, log_name: String, std_colors: bool) -> Result<Dispa
...
+}
+
+fn notify<T: serde::Serialize + std::marker::Send + std::marker::Sync + 'static>(
+    api_key: &String,
+    handle: &Handle,
+    event: Event<T>,
+) {
+    let events = EventsV2::new(api_key.to_string(), Some("lockbook-server".to_string())).unwrap();
+
+    futures::executor::block_on(async {
```
tvanderstad, line 199: * first of all can we use tokio's block on instead of futures? no need for an extra dependency
* second of all, are you sure this is okay? I expect this to crash when called from an async function. sync -> async -> sync -> async is not allowed because creating an sync runtime is a blocking operation which is not ok in async functions
* shouldn't this be an async function anyway? I know this makes life hard because [traits don't support async functions yet last I checked](http://smallcultfollowing.com/babysteps/blog/2019/10/26/async-fn-in-traits-are-hard/). I struggled making our dependency injection pattern work using that crate, which is why server doesn't have dependency injection right now

raayan: I don't think this is creating an async runtime but rather giving async work back to tokio via `handle`

Earlier with the sync version of this library I was experiencing those crashes since it was trying to start an async runtime before running the same async code

Yeah maybe it being async _does_ make sense but that would be violating all of the logging framework stuff. However, if we could have a "fire-and-forget" style piece of async code, that would be ideal since that's kinda what we want out of PagerDuty

https://stackoverflow.com/a/62536772

Parth (approved):

Issue #688, opened 2021-04-13
Deleting a file that is open should close that file


Parth:

Parth:
3

PR #700, opened 2021-04-15
linux: a deleted file does not keep its editor open

SmailBarkouch:
fixes: #688

### DIFF
--- a/clients/linux/src/app.rs
+++ b/clients/linux/src/app.rs
@@ -551,9 +551,14 @@ impl LbApp {
         if d.run() == GtkResponseType::Yes {
             for f in &file_data {
                 let (_, id, _) = f;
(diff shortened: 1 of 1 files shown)


clients/linux/src/app.rs
```diff
@@ -552,6 +552,7 @@ impl LbApp {
             for f in &file_data {
                 let (_, id, _) = f;
                 self.core.delete(&id)?;
+                self.messenger.send(Msg::CloseFile);
```
steverusso, line 555: This loop deletes each file that is selected in the file tree. The `CloseFile` message only needs to be sent once to close the currently open file, so that action should take place outside of (above) this loop.

steverusso (changes requested):

steverusso (approved):

PR #704, opened 2021-04-16
linux: settings use defaults on missing fields

### LINKED ISSUE #690: If a field is missing upon reading the setting file it should fallback to the default specified for that field
If you add a new field, existing users of the linux app will not have that new field in that file. Creating the settings object will decide the default value of the missing field arbitrarily (even though a default is specified in `settings.rs`). Could explicitly handle missing fields to solve this problem.

SmailBarkouch:
fixes #690


steverusso (approved):
Awesome!

PR #705, opened 2021-04-17
linux: renaming a file updates the header and window title

### LINKED ISSUE #687: Linux rename should effect more things
When you rename a file a few things refresh with the new name, but not all do

SmailBarkouch:
fixes #687

### DIFF
--- a/clients/linux/src/app.rs
+++ b/clients/linux/src/app.rs
@@ -604,15 +604,13 @@ impl LbApp {
                     let acctscr = &lb.gui.account;
                     acctscr.sidebar.tree.set_name(&id, &name);
 
-                    match lb.core.file_by_id(meta.id) {
+                    match lb.core.file_by_id(id) {
                         Ok(f) => {
-                            lb.gui.win.set_title(&f.name);
-                            lb.gui.account.set_search_field_text(&lb.core.full_path_for(&f));
+                            lb.gui.win.set_title(&name);
+                            acctscr.set_search_field_text(&lb.core.full_path_for(&f));
                             lb.messenger.send(Msg::RefreshSyncStatus);
                         }
-                        Err(err) => {
-                            lb.messenger.send_err("getmting renamed file", err);
-                        }
+                        Err(err) => lb.messenger.send_err("getting renamed file", err)
                     }
                 }
                 Err(err) => match err.kind() {
--- a/clients/linux/src/app.rs
+++ b/clients/linux/src/app.rs
@@ -603,10 +603,10 @@ impl LbApp {
                     d.close();
                     let acctscr = &lb.gui.account;
                     acctscr.sidebar.tree.set_name(&id, &name);
+                    lb.gui.win.set_title(&name);
 
                     match lb.core.file_by_id(id) {
                         Ok(f) => {
-                            lb.gui.win.set_title(&name);
                             acctscr.set_search_field_text(&lb.core.full_path_for(&f));
                             lb.messenger.send(Msg::RefreshSyncStatus);
                         }
--- a/clients/linux/src/app.rs
+++ b/clients/linux/src/app.rs
@@ -605,14 +605,13 @@ impl LbApp {
                     acctscr.sidebar.tree.set_name(&id, &name);
 
                     match lb.core.file_by_id(meta.id) {
-                        Ok(renamed_meta) => {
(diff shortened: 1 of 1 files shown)


clients/linux/src/app.rs
```diff
@@ -603,7 +603,18 @@ impl LbApp {
                     d.close();
                     let acctscr = &lb.gui.account;
                     acctscr.sidebar.tree.set_name(&id, &name);
-                    lb.messenger.send(Msg::RefreshSyncStatus);
+
+                    match lb.core.file_by_id(meta.id) {
+                        Ok(renamed_meta) => {
```
steverusso, line 608: This var name can just be something like `f`.

clients/linux/src/app.rs
```diff
@@ -603,7 +603,18 @@ impl LbApp {
...
+                            lb.gui.win.set_title(&renamed_meta.name);
+                            lb.gui.account.set_search_field_text(&lb.core.full_path_for(&renamed_meta));
+                            lb.messenger.send(Msg::RefreshSyncStatus);
+                        }
+                        Err(err) => {
+                            d.close();
```
steverusso, line 614: I don't think we need to `d.close()` here? We do so [on 603 above](https://github.com/lockbook/lockbook/pull/705/files#diff-c333b08e1a270da5874d2c32f455ac27d18f7854b40b6d733d50e8e6439347c7R603) before the match statement.

steverusso (changes requested):

clients/linux/src/app.rs
```diff
@@ -603,7 +603,18 @@ impl LbApp {
...
+                            lb.messenger.send_err("renaming file", err);
```
steverusso, line 615: This wouldn't be an error `"renaming file"`, it would be an error like `"getting renamed file"`.

steverusso (changes requested):

clients/linux/src/app.rs
```diff
@@ -603,7 +603,17 @@ impl LbApp {
                     d.close();
                     let acctscr = &lb.gui.account;
                     acctscr.sidebar.tree.set_name(&id, &name);
-                    lb.messenger.send(Msg::RefreshSyncStatus);
+
+                    match lb.core.file_by_id(meta.id) {
```
steverusso, line 607: I made a variable for the id, so we can just use it here:
```suggestion
                    match lb.core.file_by_id(id) {
```

clients/linux/src/app.rs
```diff
@@ -603,7 +603,17 @@ impl LbApp {
...
+                        Ok(f) => {
+                            lb.gui.win.set_title(&f.name);
+                            lb.gui.account.set_search_field_text(&lb.core.full_path_for(&f));
```
steverusso, line 610: I made a variable for the account screen, so we can just use it here:
```suggestion
                            acctscr.set_search_field_text(&lb.core.full_path_for(&f));
```

clients/linux/src/app.rs
```diff
@@ -603,7 +603,17 @@ impl LbApp {
...
+                            lb.messenger.send(Msg::RefreshSyncStatus);
+                        }
+                        Err(err) => {
+                            lb.messenger.send_err("getmting renamed file", err);
```
steverusso, line 614: Let's just keep this on one line now since it's one statement, and also "getmting" -> "getting".

clients/linux/src/app.rs
```diff
@@ -603,7 +603,17 @@ impl LbApp {
                     d.close();
                     let acctscr = &lb.gui.account;
                     acctscr.sidebar.tree.set_name(&id, &name);
-                    lb.messenger.send(Msg::RefreshSyncStatus);
+
+                    match lb.core.file_by_id(meta.id) {
+                        Ok(f) => {
+                            lb.gui.win.set_title(&f.name);
```
steverusso, line 609: The name is already stored in the `name` variable. Let's use that here, and then move this statement out of (above) this match statement.

steverusso (changes requested):

steverusso (approved):
This looks good now!

Just a note for possible future improvement: we get the new metadata (which I assume is unchanged except for the name) only to use it to get the full path for the newly named file. I'm sure we could use the only metadata variable (although no guarantees it would work into the future), but maybe a better solution would be return it from the `LbCore::rename_file` function, or even just return the full path.

PR #707, opened 2021-04-20
linux: auto save doesn't require restart

SmailBarkouch:


Parth (changes requested):
This is a good start but I think it's going to prove to be a bad time as you take on future work, mainly: #695 (auto-sync)

The problem I see is: we have to keep track of the aliveness of the thread, currently this is just based on whether auto_save should happen, but in the future it will also be based on whether auto_sync should happen, and other optional auto_things.

So this is what I suggest, and you could consider doing auto_sync right now too (or separate pr whatever) is: 

+ rename `auto_save.rs` to `background_work.rs` (and other corresponding shared state concurrency things) 
+ `spawn` the thread regardless of settings, even if all the auto stuff is off, having a background thread that sleeps for 1 second, acquires a lock, checks a boolean and goes back to sleep is probably only slightly more expensive than a log statement.
+ the strategy where the channel updates the mutex values is good, but don't be spawning threads in here, it's going to be hard to reason about how many threads are running (another reason to just spawn 1 thread right at the start)
+ inside the `loop` in `background_work` (formerly `auto_save`) is where you should see if you're actually supposed to save or not

If you decide to do auto-sync now, read through this: https://github.com/lockbook/lockbook/blob/parth/lifecycle-docs/docs/lifecycle.md

PR #710, opened 2021-04-24
infra: admin cli

### LINKED ISSUE #601: Create an admin cli
+ [ ] Find orphaned files in s3
+ [ ] Delete a user

Parth:
# Admin CLI

Creates the skeleton for an Admin CLI, this is where we'll put logic that would traditionally be behind a customer support / product manager UI. Since our team is technical this can just be a CLI. As we need to take administrative action in the backend we'll add that logic here, gives us a place to store useful queries, and perform operations between s3 & postgres.

Currently I've added a script that will nuke an account out of existence. This is not the close my account / mark as compromised operation that will come later. This is more of a "we ran integration tests against prod by accident".

# Enhanced testing

Integration tests can now access postgres & s3 clients. This lets you assert arbitrary things, while using core to do heavy lifting. For instance core creates accounts very effectively, it's annoying to create dummy private keys, and dummy root folders, etc. But this sort of testing framework can allow someone to assert that deleting a document deletes it out of s3.

# Top level folder cleanup

Move a number of misc items out of the top level directory to sub directories

# Step 1 for migrating to aws.

@raayan is writing some scripts to set us up on aws with ease. We will leverage these scripts to do load testing and be able to spin up infrastructure quickly given a series of debian boxes.

closes #601


server/admin/src/delete_account.rs
```diff
@@ -0,0 +1,58 @@
...
+                eprintln!(
+                    "Failed to delete file in s3: {}, error: {:#?}",
+                    file.id, err
+                )
+            })
+            .is_err();
+
+            if problem {
+                ok = false;
+            }
```
tvanderstad, line 53: relatable

server/admin/src/main.rs
```diff
@@ -0,0 +1,54 @@
...
+use tokio_postgres::Client as PostgresClient;
+
+#[derive(Debug, PartialEq, StructOpt)]
+#[structopt(about = "A utility for a lockbook server administrator.")]
+enum Subcommands {
+    /// Purge a user, and all their files from postgres & s3
+    ///  *Note: This is intentionally left unexposed to give the user experience of deleting a user more
+    /// thought. This includes thinking about being able to mark themselves as compromised and indicate to
+    /// collaborators that certain files are potentially compromised. This could also involve us reaching out
+    /// to services like Strip / Apple / Google and terminating open subscriptions.
```
tvanderstad, line 21: Stripe

tvanderstad (approved):

server/admin/src/main.rs
Parth, line 21:
```suggestion
    /// to services like Stripe / Apple / Google and terminating open subscriptions.
```

tvanderstad (approved):

PR #711, opened 2021-04-24
linux: implement auto sync and auto save setting change does not require restart

### LINKED ISSUE #695: Auto sync with setting to disable

SmailBarkouch:
Auto sync will occur if these criteria are all met:
- Auto sync is enabled in settings
- A file has not been edited in the last 60 seconds
- Last sync was more than 90 seconds ago
- The current file is not dirty

fixes #695 
fixes #707


clients/linux/src/app.rs
```diff
@@ -80,11 +80,16 @@ impl LbApp {
             messenger: m,
         };
 
+        let background_work = lb_app.state.borrow().background_work.clone();
+
+        thread::spawn(move || BackgroundWork::init_background_work(background_work));
+
         if s.borrow().auto_save {
```
Parth, line 83: We check to see if we should send true.

We should just send the value of the boolean regardless of true or false.

clients/linux/src/background_work.rs
```diff
@@ -0,0 +1,105 @@
...
+        SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .expect("Time went backwards")
+            .as_millis()
+    }
+
+    pub fn init_background_work(state: Arc<Mutex<Self>>) {
+        loop {
+            thread::sleep(Duration::from_secs(1));
+            let mut bks = state.lock().unwrap();
```
Parth, line 31: bgs?

clients/linux/src/background_work.rs
```diff
@@ -0,0 +1,105 @@
...
+            let m = bks.messenger.clone();
+
+            if bks.auto_save_state.is_active {
+                bks.auto_save_state.auto_save(&m);
+            }
+
+            if bks.auto_sync_state.is_active {
+                let last_change = bks.auto_save_state.last_change;
```
Parth, line 39: Organizing the struct (the data) by splitting out `AutoSyncState` and `AutoSaveState` is good. But when you also delineate the logic like that, you have to do things like this. 

If you were to move the logic into the main struct, you could access these members directly from a `auto_sync` function.

This function would then become a cleaner:

```
if field.is_active {
    do_field();
}

if other_field.is_active {
    do_other_field();
}
```

clients/linux/src/background_work.rs
```diff
@@ -0,0 +1,105 @@
...
+        Self {
+            is_active: false,
+            last_sync: 0,
+        }
+    }
+
+    pub fn auto_sync(&mut self, m: &Messenger, last_change: u128, last_save: u128) {
+        let current_time = BackgroundWork::current_time();
+
+        let time_between_edit = current_time - last_change;
```
Parth, line 96: Imagine a timeline.

The word `between` makes my brain think that the start point and endpoint of a subsection of the timeline are non-fixed values.

The word `since` makes my brain think that the endpoint is fixed and the start point is a non-fixed value. I think you want the word `since` and not `between`.

clients/linux/src/background_work.rs
```diff
@@ -0,0 +1,105 @@
...
+        let time_between_syncs = current_time - self.last_sync;
+        let is_file_clean = last_save as i128 - last_change as i128 > 0;
```
Parth, line 98: could this not just be `last_save` > `last_change`?

clients/linux/src/background_work.rs
```diff
@@ -0,0 +1,105 @@
...
+
+        if time_between_syncs > 60000 && (time_between_edit > 90000 && is_file_clean) {
```
Parth, line 100: It's okay to not make these guys settings just yet, but I would hoist these up (along with the auto_save ones I wrote) as constants somewhere (probably `background_work.rs`). And I would leave a `TODO` saying that we should expose these as settings in the future if we or other people want that flexibility.

Parth (changes requested):
A good foundation, and a good second step, some small change requested. Also I understand the desire for a clean start, but you may be better off getting comfortable with `git reset` and `git push --force` rather than opening new PRs.

clients/linux/src/background_work.rs
Parth: I think we actually want something a bit more complex that's going to feel better for the user:

Right now if there haven't been edits in a while we sync every 60 seconds. This is too often.

But we don't want to wait 30 minutes after an edit to sync.

So what we should do is:
+ if more than 90 seconds have elapsed after an edit, sync
+ then sync every 30 minutes after

Which gives us sort of two modes, an initial strategically placed sync to pickup the edit the user just made, and a passive background sync so when the user picks up the app it's always up to date.

clients/linux/src/background_work.rs
```diff
@@ -0,0 +1,120 @@
...
+                bgs.auto_save();
+            }
+
+            if bgs.auto_sync_state.is_active {
+                bgs.auto_sync();
+            }
+        }
+    }
+
+    pub fn auto_sync(&mut self) {
```
Parth, line 43: I have a feeling you don't need `self.auto_sync_state.synced_last_edit` to exist.

Here's how I would probably model this function without that:

```
let unsaved_changes = last_change > last_save;
if unsaved_changes {
    return;
}

let unsynced_changes = last_save > last_sync;
let time_since_sync = ...
if unsynced_changes {
    if time_since_sync > Self::SYNC_AFTER_EDIT_DELAY {
        sync();
        return;
    } else {
        return;
    }
}

// There are no unsynced or unsaved changes at this point
if time_since_last_sync > TIME_BETWEEN_SYNCS {
    sync;
    return;
else {
    return;
}
```

You could declare all these helper variables up top and use if elses to avoid early returns, see which one is cleaner.

clients/linux/src/background_work.rs
```diff
@@ -0,0 +1,120 @@
...
+        let current_time = Self::current_time();
+
+        let time_since_edit = current_time - self.auto_save_state.last_change;
```
Parth, line 46: In all the references to `time_since_...` or `time_between_...`, we should probably call all these variables `millis_since...` or `millis_between_...`

clients/linux/src/background_work.rs
```diff
@@ -0,0 +1,120 @@
...
+        {
+            // There are changes since we last saved
+            self.auto_sync_state.synced_last_edit = false;
+            self.auto_save_state.last_save = current_time;
+            self.messenger.send(Msg::SaveFile);
+        }
+    }
+
+    pub const TIME_BETWEEN_SYNCS: u128 = 1800000;
+    pub const SYNC_TIME_AFTER_EDIT: u128 = 90000;
```
Parth, line 82: SYNC_AFTER_EDIT_DELAY

clients/linux/src/background_work.rs
```diff
@@ -0,0 +1,120 @@
...
+    pub const SAVE_TIME_AFTER_EDIT: u128 = 1000;
```
Parth, line 83: SAVE_AFTER_EDIT_DELAY

Parth (changes requested):

Parth (approved):

PR #712, opened 2021-04-24
linux: gtk is initialized before settings and lbcore

### LINKED ISSUE #703: If an error occurs before the gtk application has been initiated, it will crash
If at start up, an error happens due to the settings, or core cannot be initialized, it will attempt to create a dialog. But since the GTK application has not been initiated, it will crash.

```
thread 'main' panicked at 'GTK has not been initialized. Call `gtk::init` first.', 
```

Note, this may be tied to debian crashes.

SmailBarkouch:
If an error occurred while initializing `Settings` or `LbCore`, we call `launch_err` to post a dialog for this error. The problem is, you can't post a dialog if gtk+ is not initialized, and we initialize gtk+ on the very last line of the main function with `gtk_app.run(&[])`. So if an error happens, it will crash when attempting to run the dialog. So `gtk::init().unwrap()` has been added so we can successfully run dialogs before the `GtkApp` itself is run.

fixes #703

### DIFF
--- a/clients/linux/src/main.rs
+++ b/clients/linux/src/main.rs
@@ -38,6 +38,7 @@ use crate::settings::Settings;
 
 fn main() {
     let data_dir = get_data_dir();
+    gtk::init().unwrap();
 
     let core = match LbCore::new(&data_dir) {
         Ok(c) => Arc::new(c),
--- a/clients/linux/src/main.rs
+++ b/clients/linux/src/main.rs
@@ -37,8 +37,9 @@ use crate::backend::LbCore;
 use crate::settings::Settings;
 
 fn main() {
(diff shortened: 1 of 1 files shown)


Parth (approved):

clients/linux/src/main.rs
```diff
@@ -38,6 +38,7 @@ use crate::settings::Settings;
 
 fn main() {
     let data_dir = get_data_dir();
+    gtk::init().unwrap();
```
steverusso, line 41: Just make this the first line then followed by a blank line.

steverusso (changes requested):

steverusso (approved):

Issue #715, opened 2021-04-28
Don't background sync if there's no account


Parth:
either don't start the thread, or have the thread do nothing when there's no account

PR #717, opened 2021-04-29
linux: background sync and save wont start if there is no account

SmailBarkouch:
fixes: #715

### DIFF
--- a/clients/linux/src/app.rs
+++ b/clients/linux/src/app.rs
@@ -85,7 +85,7 @@ impl LbApp {
             .send(Msg::ToggleAutoSave(s.borrow().auto_save));
         lb_app
             .messenger
-            .send(Msg::ToggleAutoSync(s.borrow().auto_save));
+            .send(Msg::ToggleAutoSync(s.borrow().auto_sync));
 
         let lb = lb_app.clone();
         receiver.attach(None, move |msg| {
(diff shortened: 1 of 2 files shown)


clients/linux/src/background_work.rs
```diff
@@ -25,17 +26,23 @@ impl BackgroundWork {
...
             thread::sleep(Duration::from_secs(1));
             let mut bgs = state.lock().unwrap();
 
-            if bgs.auto_save_state.is_active {
-                bgs.auto_save();
-            }
-
-            if bgs.auto_sync_state.is_active {
-                bgs.auto_sync();
+            match core.has_account() {
```
Parth, line 34: Thought of a better way to do this too. Instead of checking this each loop, you could create a sort of finite state machine. You keep checking every one sec until an account is available, when it is that loop terminates and the next loop starts. No need to check again.

But there's an even better way to do this. The app is already doing a check to see if an account is there or not, if it's there it shows the main scree, otherwise it shows the intro screen. That should be a warning for you to not also do that check yourself.

[This](https://github.com/lockbook/lockbook/blob/master/clients/linux/src/app.rs#L1008) is where that decision is made. I would do this the following way:

Have `Gui` hang onto the `Messenger` it gets (the same way `Editor` does). And create a new message: `AccountScreenShown`. I would then subscribe to this new message in `app.rs`. Follow the convention of creating a helper function, in that helper function is where you should launch the thread.

So in essence the thread will only get launched after the `Gui` says "Hey we launched the account screen, we're passed onboarding".

Parth (changes requested):

Parth (approved):

PR #718, opened 2021-04-29
core, linux, cli: use closures to sync with the same algorithm across platforms

### LINKED ISSUE #716: Linux stops syncing if a single error is encountered, rather than trying to resolve it first

SmailBarkouch:
In this PR, I modified the sync algorithm to accept closures. This gets rid of the re-implementations of this sync function in the rust based clients. 

I also got rid of the sync spinner in place of a progress bar.

## Screenshots
<details>

</details>

fixes #716


Parth (changes requested):
I thought of a better way to go about this:

we have the same logic implemented in rust 3 different ways. It's also a reasonably complicated section of code with a few things going on. We should extend the implementation in core so that `cli` and `linux` don't need to duplicate it.

Rework this so that the implementation in `core` is the canonical one, it can accept an optional closure that it will call every time it successfully syncs a unit of work (pass the new `file_metadata` or something like that). It will also let it know it's current total and completed work (for progress bars).

The callers of this function shouldn't have to worry about things like setting last synced, and retrying errors. They'll just setup a closure that recv's this `Progress` object which contains whatever they care about, either % complete or which file is about to be synced and then call sync_all with that closure.

The `linux` app will probably use that closure to send things down a channel.

And `cli` will probably use that closure to print to `stdout`.

Parth (changes requested):
2 Changes while I experiment to see what's possible with `c_interface`:

+ Can eliminate `SyncState` and cleanup some code
+ Explore exposing `sync` as a macro, see the `add!` example from [this SO](https://stackoverflow.com/questions/24047686/default-function-arguments-in-rust). If this is too annoying, just create a `sync` and `sync_with_progress` helper `fn`. This should eliminate many instances of `sync(&db, None)`

Parth:
I have verified that this is possible via: https://www.nickwilcox.com/blog/recipe_swift_rust_callback/

Parth:
Looks pretty easy in C# land too

Parth (approved):

PR #719, opened 2021-04-30
server: billing part 2

Parth:
# [Project Initiation Document](https://github.com/lockbook/lockbook/blob/parth/billing2/docs/billing.md)

# Progress

+ [x] Admin CLI
+ [x] Determine pricing structure
+ [x] Overview doc
+ [x] Create db scripts for new tables
+ [x] Deprecate `usage_ledger` in favor of a column in `files`
+ [x] `get-usage` returns total usage + your current tier
+ [ ] Detect when a write will exceed tier space
+ [ ] Cli
+ [ ] Integrate stripe creates
+ [ ] Integrate stripe webhooks
+ [ ] General integration tests
+ [ ] Stripe testing
+ [ ] Linux
+ [ ] MacOS
+ [ ] Windows
+ [ ] Apple setup in connect
+ [ ] iOS / macOS UI work
+ [ ] Apple webhook
+ [ ] Apple testing
+ [ ] Android UI work
+ [ ] Google setup in play
+ [ ] Google webhook
+ [ ] Google testing

# Closes

#719


server/server/src/file_index_repo.rs
```diff
@@ -856,3 +902,90 @@ pub async fn delete_account(
...
+pub async fn get_account_data_cap(
+    transaction: &Transaction<'_>,
+    public_key: &RSAPublicKey,
+) -> Result<u64, GetDataCapError> {
+    match transaction
+        .query(
+            "SELECT bytes_cap
+            FROM account_tiers
+            WHERE account_tier_id =
+                (select account_tier from accounts where public_key = $1);",
```
tvanderstad, line 915:
```suggestion
                (SELECT account_tier FROM accounts WHERE public_key = $1);",
```

server/server/src/file_index_repo.rs
```diff
@@ -856,3 +902,90 @@ pub async fn delete_account(
...
+
+pub async fn get_file_usages(
+    transaction: &Transaction<'_>,
+    public_key: &RSAPublicKey,
+) -> Result<Vec<FileUsage>, UsageCalculateError> {
+    let result = transaction
+        .query(
+            "SELECT id, document_size FROM files
+            WHERE owner = (SELECT name FROM accounts WHERE public_key = $1)
+                and not is_folder;",
```
tvanderstad, line 971:
```suggestion
                AND NOT is_folder;",
```

server/server/src/file_index_repo.rs
```diff
@@ -172,7 +189,8 @@ pub async fn change_document_content_version(
                 content_version =
                     (CASE WHEN NOT old.deleted AND old.metadata_version = $2 AND NOT old.is_folder
                     THEN CAST(EXTRACT(EPOCH FROM NOW()) * 1000 AS BIGINT)
-                    ELSE old.content_version END)
+                    ELSE old.content_version END),
+                document_size = $3
```
tvanderstad, line 193: this should be a conditional set just like the content version one is. yeah, you have to copy and paste that whole `CASE` thing (the 3 lines above) because SQL doesn't have good abstractions. otherwise document size will be updated even if this fails (for instance because their version is out of date)

server/server/src/file_service.rs
```diff
@@ -78,18 +78,6 @@ pub async fn change_document_content(
         return Err(None);
     };
 
-    usage_repo::track_content_change(
```
tvanderstad, line 81: it looks to me like deleted files would still be counted towards usage. probably do one of:
* call file_index_repo::change_document_version_and_size here
* update the size to 0 in file_index_repo::delete_file
* exclude deleted files in addition to folders in file_index_repo::get_file_usages

tvanderstad (changes requested):

server/server/src/file_index_repo.rs
Parth: a915883

travis-madhive (approved):

tvanderstad (approved):

PR #732, opened 2021-05-10
linux: create a new file by selecting a parent in the file tree

### LINKED ISSUE #419: Improve UI/UX around creating "New" file
Currently, there's a solitary "New" dialog that uses `create_at_path` which will create a directory if there is a trailing slash and a document otherwise.
* Instead, let's separate this into two explicit avenues: creating a document and creating a folder. This way, a user knows for sure which is being created and doesn't have to worry about forgetting a slash or something. Additionally, these New dialogs should take a parameter that is the location where the new thing is intended to be created. This way, the input field can be prefixed with the directory under which the user is creating a new file. This is purely for UX and will be leveraged once... :arrow_down: 
* Add 'New Document' and 'New Folder' items to the file tree's popup menu. Send the selected files ID with the Msg so that the input fields can be prefixed with its path. These items should only be enabled if there is exactly one file selected.

### LINKED ISSUE #733: Creating a new file should be done by selecting a file in the filetree

SmailBarkouch:
In this PR, I removed the ability to create a new file from a path and added the ability to create a new file from the parent that is selected in the file tree. Users simply right click a folder or document, and enter the name of the folder or document they want to create. The file will then be created in the selected folder, or under the same parent of the selected document.

## Example
<details>

Screenshot_20210510_134748
</details>

fixes #733 
fixes #419


clients/linux/src/app.rs
```diff
@@ -306,12 +305,82 @@ impl LbApp {
...
+        d.set_default_size(300, -1);
+        d.get_content_area().add(&lbl);
+        d.get_content_area().add(&entry);
+        d.add_button("Ok", GtkResponseType::Ok);
+        d.set_default_response(GtkResponseType::Ok);
+
+        let parent = match self.gui.account.sidebar.tree.get_selected_uuid() {
+            Some(id) => {
+                let file = self.core.file_by_id(id)?;
+                let parent = if let FileType::Folder = file.file_type {
```
steverusso, line 337:
```rust
Ok(match file.file_type {                                                                                                                                                                          
    FileType::Document => file.parent,                                                   
    FileType::Folder => file.id,                                                         
})
```
...is more readable imo.

clients/linux/src/app.rs
```diff
@@ -99,7 +99,7 @@ impl LbApp {
 
                 Msg::AccountScreenShown => lb.account_screen_shown(),
 
-                Msg::NewFile(path) => lb.new_file(path),
+                Msg::NewFile(name) => lb.new_file(name),
```
steverusso, line 102: Shouldn't `name` be something like `file_type` instead?

steverusso (changes requested):

PR #758, opened 2021-05-23
core: RSA -> Elliptic Curves (secp256k1)

### LINKED ISSUE #181: Reduce the size of account_string
Account string is pretty large, the QR code generated seems error prone and sometimes hard to fit on one screen

Parth:
[A good primer on elliptic curves and why you would use them over rsa](https://blog.cloudflare.com/a-relatively-easy-to-understand-primer-on-elliptic-curve-cryptography/).

[This particular curve is the same curve used in Bitcoin](https://en.bitcoin.it/wiki/Secp256k1).

Making keys smaller allows people to write them down, using the same curve as bitcoin allows us to use cool things like the [bip 39 word lists](https://github.com/rust-bitcoin/rust-bip39/blob/master/src/language/english.rs) and other innovations from this cryptosystem. 

This is 1 of many several breaking changes we're going to make in rapid succession as put the final touches on our data model.

Fixes #181 

Old private key experience: 
<details>

image

</details>

New private key experience:
<details>

image

</details>


clients/android/app/src/main/java/app/lockbook/util/CoreData.kt
```diff
@@ -47,34 +47,18 @@ enum class FileType {
...
-    val n: List<Int>,
-    val e: List<Int>,
-    val d: List<Int>,
-    val primes: List<String>
-)
-
-data class RSAPublicKey(
-    val n: List<Int>,
-    val e: List<Int>
+    val privateKey: List<Int>,
```
tvanderstad, line 61: is `List<Int>` really Kotlin's best analog to Rust's `[u8; 32]`?

core/libs/models/src/account.rs
```diff
@@ -8,5 +8,59 @@ pub type ApiUrl = String;
...
+}
+
+impl Account {
+    pub fn public_key(&self) -> PublicKey {
+        PublicKey::from_secret_key(&self.private_key)
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
+pub struct SecretKeyHolder(pub [u8; 32]);
```
tvanderstad, line 22:
```suggestion
```

core/libs/models/src/crypto.rs
```diff
@@ -8,7 +8,7 @@ use uuid::Uuid;
 pub type AESKey = [u8; 32];
 pub type DecryptedDocument = Vec<u8>;
 pub type EncryptedDocument = AESEncrypted<DecryptedDocument>;
-pub type EncryptedUserAccessKey = RSAEncrypted<AESKey>;
+pub type EncryptedUserAccessKey = AESEncrypted<AESKey>;
```
tvanderstad, line 11: what's the idea for this change?

Parth: With ECDH 2 parties knowing each others public key can generate a shared secret without talking to one another. The concept of encrypting something with Elliptic Curve is to encrypt it with this shared secret.

tvanderstad: shouldn't it be encrypted with ECDH instead of AES then?

Parth: The shared secret is an arbitrary `u8; 32` we use it with AES as a key to encrypt. When we do this the resulting value is AESEncrypted

tvanderstad (approved):

PR #768, opened 2021-05-24
linux: fixed sync crashes and selection is preserved

### LINKED ISSUE #689: If you sync a move it doesn't refresh the file tree

SmailBarkouch:
This bug originated from [this](https://github.com/lockbook/lockbook/pull/737). Whenever the file tree was populated, using the `fill` method from the `FileTree` struct, it triggered the `on_selection_change` closure. This closure is meant to modify the availability of the pop up items (the options that appear when you right click a file) since a new file was selected. But it assumed (using an `.unwrap()`) that there was at least one file in the tree. The problem was, when `fill` is used, the filetree was cleared and added back all its children. Usually this was just called once when the app was opened, which was before the `on_selection_change` closure was added to the tree. Now it was being used while the closure was already there. The fix was to just get rid of the unwrap and allow for an early return.

fixes #689

### DIFF
--- a/clients/linux/src/filetree.rs
+++ b/clients/linux/src/filetree.rs
@@ -425,28 +425,27 @@ impl FileTreePopup {
         let tsel = t.get_selection();
         let tmodel = t.get_model().unwrap();
 
-        let is_root = tsel.iter_is_selected(&match tmodel.get_iter_first() {
-            Some(iter) => iter,
-            None => return, // early return since tree is empty
-        });
+        if let Some(iter) = tmodel.get_iter_first() {
+            let is_root = tsel.iter_is_selected(&iter);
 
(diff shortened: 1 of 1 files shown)


clients/linux/src/filetree.rs
```diff
@@ -419,12 +425,16 @@ impl FileTreePopup {
         let tsel = t.get_selection();
         let tmodel = t.get_model().unwrap();
 
+        let is_root = tsel.iter_is_selected(&match tmodel.get_iter_first() {
```
Parth, line 428: you should put this in an `if let` rather than returning early. You can put the rest of this fn in the if let, because that's what's actually happening here

Parth (changes requested):

Parth (approved):

PR #773, opened 2021-05-27
android: drawingview's white and black color values switch depending on device theme

### LINKED ISSUE #668: switch white and dark in light mode and dark mode
In a drawing, when you are in dark mode you should render white as black and black as white. This is so strokes that are drawn don't disappear when the mode is switched.

SmailBarkouch:
In this PR, white and black will switch values in the `drawingview` depending on whether the device is in darkmode or lightmode. This means that if you are in lightmode and drawing with the black `ColorAlias`, it will become the white `ColorAlias` in darkmode. Effectively allowing strokes to visible regardless of the theme.

Recorded Example:
<details>

https://user-images.githubusercontent.com/20663038/119876015-696dca80-bef5-11eb-9c6a-46ca5fb2f4a4.mp4
</details>

fixes #668


Parth (approved):
v nice

PR #776, opened 2021-05-28
docs: ux

tvanderstad:


docs/design-ux/editor.md
```diff
@@ -0,0 +1,37 @@
+# Editor
+When a user has a document selected, that file is open for editing in an editor. The name of the document is displayed above the document and is editable (editing renames the document).
```
Parth, line 2: Depending on the UI platform, this could be challenging, I agree it's worth doing.

docs/design-ux/editor.md
```diff
@@ -0,0 +1,37 @@
...
+
+An icon near the document's name indicates one of the following statuses:
+* the document is presented exactly as it appears on disk
```
Parth, line 5: I wonder if

> all changes to the document have been merged into the version on disk

Is too complicated an idea to map cleanly to an icon that's going to exist on every platform. 

> there are unsaved changes to the document

could map to an :writing_hand: icon

> the document is presented exactly as it appears on disk

could map to :heavy_check_mark: 

emoji could be a good way to bound which icons we use within the app. Partly because we don't currently have the resources to author good icon sets, but mainly because they have meaning that people already understand.

docs/design-ux/editor.md
```diff
@@ -0,0 +1,37 @@
...
+* all changes to the document have been merged into the version on disk
+* there are unsaved changes to the document
+
+At most 5s after the user completes an edit on this document, the document is saved. At most 60s after the user completes an edit on any document, a sync occurs and any open documents are reloaded from disk.
+
+On platforms that present an editor side-by-side with navigation, the navigation is collapsible.
+
+## Markdown
+At most 1s after the user completes an edit, the document is syntax-highlighted and the active style is set in accordance with the syntax highlighting at the cursor's position. The syntax highlighting makes links clickable to open the link with the user's default browser.
```
Parth, line 14: If you're bringing up links, maybe we want to also think through embedded drawings like we saw from the other app during our call.

Could also be worth exploring the links story:
+ relative links
+ permanent links (ids)
+ be able to answer which documents link to this one

docs/design-ux/editor.md
```diff
@@ -0,0 +1,37 @@
...
+* undo and redo
+    * strokes added and removed
+    * modifications made by platform-specific features (e.g. iPadOS's "duplicate")
+* pan & zoom
+    * cannot pan to where none of the canvas is visible
+    * cannot zoom out further than would reveal the whole canvas regardless of pan
+    * cannot zoom in more than 10x
+* a touch input toggle (off by default)
+
+Drawings are rendered on a fixed-size canvas. By default, the width of the canvas is exactly the screen's width. The height of the canvas is 2x the width. The non-canvas area is rendered as a solid neutral color distinct from the canvas background color. The canvas background color is determined by the system's light/dark theme setting and drawing theme's white/black colors. In dark mode, the white and black colors in the drawing are swapped (including in the toolbar).
```
Parth, line 37: Yeah roughly my drawing thoughts, but missing:

There's two modes I consider while thinking about our ideal drawing state:

+ a mode that lets me insert a drawing into a markdown document
+ a mode that lets me model a notebook

We could generalize this to: "a folder whose extension is `.nb` or something like that. But also down to laser focus on a drawing centric use case to prove out the idea first.

While on the topic of drawings, I think the ability to annotate pdfs & images is something a lot of users will expect.

docs/design-ux/launch.md
```diff
@@ -0,0 +1,8 @@
+# Launch
+When users launch the app, it checks if there is a lockbook account on the device. If so, the app proceeds to naviation. Otherwise, it prompts for sign-up or import.
+
+If the user opts to create an account, they enter a name and click to submit. If the operation succeeds, the app proceeds to navigation. Otherwise, an unobstructive message appears to represent the failure and the user can retry.
```
Parth, line 4: I think prior to creating an account the user should be made aware of a few things. I think the right way to think about this is that this will be a *ULTRA* readable TOS. It should be clear that they're signing up to manage their own key, they should probably import into another device asap (for redundancy) or save the account string somewhere. etc.

Parth (changes requested):
Some more thoughts

docs/design-ux/launch.md
tvanderstad: todo: succinctly explain "TOS" aka self-managed keys are not recoverable

docs/design-ux/navigation.md
```diff
@@ -0,0 +1,16 @@
+# Navigation
+Once a user is signed in, they are presented with a navigation menu. On some platforms it will be side-by-side with an editor.
+
+The navigation view shows the user's file tree. By default, the root (but no other folders) is expanded. The root is not collapsible. Files display their name and relative last modified time. Users are able to sort by name or last modified, ascending or descending. Folders are displayed above files. The default sort is by name.
```
tvanderstad, line 4: "Folders are displayed above files." only when sorted by name

docs/design-ux/navigation.md
```diff
@@ -0,0 +1,16 @@
...
+
+Users can fuzzy search for their files by title. Fuzzy search results are presented in order of their match strength with the search string. On platforms with keyboards, the user can navigate the search results with the arrow keys or select one of the top results with a displayted hotkey (e.g. cmd-1, cmd-2, cmd-3 for the top 3 results).
```
tvanderstad, line 6: "the user can navigate the search results with the arrow keys" and select with enter

docs/design-ux/navigation.md
```diff
@@ -0,0 +1,16 @@
...
+
+Users can create, move, rename, and delete files from the navigation view.
+
+There is a status bar at the bottom of the navigation. It displays the first applicable status of:
+* offline (if offline)
+* number of dirty files (if any)
+* storage warning (if you are >90% usage)
+* relative last synced time
+
+While a manually requested sync is in progress, a progress bar shows the progress. The bar does not move backwards.
```
tvanderstad, line 16:
```suggestion
While any sync is in progress, a progress bar shows the progress. The bar does not move backwards.
```

docs/design-ux/navigation.md
```diff
@@ -0,0 +1,16 @@
+# Navigation
+Once a user is signed in, they are presented with a navigation menu. On some platforms it will be side-by-side with an editor.
+
+The navigation view shows the user's file tree. By default, the root (but no other folders) is expanded. The root is not collapsible. Files display their name and relative last modified time. Users are able to sort by name or last modified, ascending or descending. Folders are displayed above files. The default sort is by name.
+
+Users can fuzzy search for their files by title. Fuzzy search results are presented in order of their match strength with the search string. On platforms with keyboards, the user can navigate the search results with the arrow keys or select one of the top results with a displayted hotkey (e.g. cmd-1, cmd-2, cmd-3 for the top 3 results).
+
+Users can create, move, rename, and delete files from the navigation view.
```
tvanderstad, line 8: creating a document automagically fills a name and brings you to the editor

docs/design-ux/editor.md
```diff
@@ -0,0 +1,37 @@
...
+On platforms that present an editor side-by-side with navigation, the navigation is collapsible.
+
+## Markdown
+At most 1s after the user completes an edit, the document is syntax-highlighted and the active style is set in accordance with the syntax highlighting at the cursor's position. The syntax highlighting makes links clickable to open the link with the user's default browser.
+
+The editor non-intrusively spell-checks the user's document.
+
+The editor supports undo and redo, including:
+* text added or removed
+* text pasted
```
tvanderstad, line 20: text cut

docs/design-ux/editor.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Drawing
+There is a toolbar at the top of editor which features:
```
tvanderstad, line 23:
```suggestion
There is a toolbar at the left of the editor which features:
```

docs/design-ux/editor.md
```diff
@@ -0,0 +1,37 @@
...
+* pen size control (thickness range is from 1 pixel to 20 pixels at full pen pressure)
+* a stroke eraser
+* an erase-all
+* undo and redo
+    * strokes added and removed
+    * modifications made by platform-specific features (e.g. iPadOS's "duplicate")
+* pan & zoom
+    * cannot pan to where none of the canvas is visible
+    * cannot zoom out further than would reveal the whole canvas regardless of pan
+    * cannot zoom in more than 10x
```
tvanderstad, line 34:
```suggestion
    * cannot zoom in more than 20x
```

docs/design-ux/editor.md
```diff
@@ -0,0 +1,37 @@
...
+* pen color controls (8 colors from drawing theme or, if drawing has no theme, 8 colors from app theme)
+* pen size control (thickness range is from 1 pixel to 20 pixels at full pen pressure)
+* a stroke eraser
+* an erase-all
+* undo and redo
+    * strokes added and removed
+    * modifications made by platform-specific features (e.g. iPadOS's "duplicate")
+* pan & zoom
+    * cannot pan to where none of the canvas is visible
+    * cannot zoom out further than would reveal the whole canvas regardless of pan
```
tvanderstad, line 33: ???

docs/design-ux/editor.md
tvanderstad: how about some kind of refresh/🔁 type icon, to indicate that there's something on disk that isn't in your editor? do you think having ✍️ and 🔁 show up next to each other at the same time is too bad of a way to indicate that there are unsaved changes in the editor *and* unloaded changes on disk at the same time?

docs/design-ux/editor.md
tvanderstad, line 20:
```suggestion
* text pasted or cut
```

docs/design-ux/editor.md
tvanderstad, line 33:
```suggestion
    * cannot zoom out more than 5x (20% zoom)
```

Parth (approved):

PR #778, opened 2021-05-29
core: reduce core to it's spiritual essence

### LINKED ISSUE #741: Core cleanup: organize tests
+ there are non-core integration tests that can probably be relocated to server and made more expressive (since in that location they have access to things like postgres & s3)
+ some macros were created to dry up things, but now the macros are duplicated in a few places, sort this out

### LINKED ISSUE #774: core cleanup: be more selective about our use of DI
# Context

When we first started we weren't really sure which components we wanted to be able to have competing implementations of, what we wanted to mock out and test, and how our team would be working with core.

Now that we know a bit more about all those things we can phase out DI out of the majority of core, leaving it place for: `ApiClient`, `Clock` and additionally perhaps for sources of randomness. We can also make the atomic building block of core a `fn` rather than a `trait + struct`. So the DI'd bits are only required for `fn`s that require them, not entire collections of `fn`'s. 

A brain dump about how core will grow in the future and why we can do this rather fearlessly:

First, in principle, especially with core (the heart of our project) we should keep the codebase small enough that we can make it represent the current source of truth, not intended future states (premature optimizations) or artifacts of history (tech debt). For example, I'd very much like to support multiple crypto implementations later, but we shouldn't have generified these bits until that's a reality. As soon as we stopped using sled, we should have ripped it out entirely, we left it in out of fear that we may want to go back (that's what git is for), we should leave competing implementations in when we feel like there's value to both.

Second, (as @steverusso and many new readers of code bring up), the current DI strategy hurts readability, though no one suggested quite removing DI entirely, I think it's a move (among other options discussed below) that maximally improves readability of our code.

More practically, these are the things I think we'll want to do in core in the medium-long term:

## The future of performance within core

Core is essentially a tool to interact with values on disk. It's modeled as a collection of independent endpoints that do various operations. An area for speedup could be to share state across such operations in memory. Let's explore some ways we can model this:

### Dynamic Dispatch + DI

This is something we explored with @steverusso and in most normal contexts this is probably the move. Basically the cached bits live in the module which is responsible for them (files in `FileIndexRepo`). What makes me not like this is:

1. Forces calls to happen via a `lb` handle
    a. requires some boilerplate for ffi/jni but it's limited to those files so acceptable
    b. requires the caller to hang on to a piece of data that they perhaps do not care about. Would require managing a lifecycle for this handle
2. Performance penalties associated with dynamic dispatch (core is intended to be used by a wide range of hardware). Still though, this is orders of magnitude faster than file io so nbd in the grand scheme of things.
3. I'm not sure I want to couple my logic in that way with the data (my main concern)
4. @tvanderstad's main concern (i think): moves a bunch of operations whose answers are known at compile time to runtime (which impl). Now the whole reason I feel like getting rid of DI in core is because all the traits only have one impl. So in this scenario these questions are easy to answer at compile time by a human anyways.

### Static Dispatch + DI

I think in a static dispatch DI world where code is strung together via generics, it forces you to move this state into some sort of `Context` struct. This addresses point 2, 3 & 4 (to some extent), but consumers of core as a library still have to deal with this pointer. This makes explicit to some extent that there is some state being dealt with within core. This context struct would be passed around as a fn parameter (and can be scoped if desired). So this solution doesn't really care what the DI strategy overall is. DI in this world is just bringing together pieces of logic, that logic can take whatever arguments it needs to perform the tasks it's trying to perform.

Even this variable has several properties that we can express at compile time. And I'd be interested in exploring the `lazy_static!` crate to fully constrain the lifecycle of the context variable (the context would be allocated on the stack I believe). It would make interacting with the crate very straightforward, but would introduce a bit of magic behavior which could trip up people importing lb for the first time. Maybe we could be obnoxious and mention it in every fn doc or something.  

## The future of additional complexity in core

The following is a problem present in just about every client (apart from CLI): the state in memory corresponds to an older on-disk state, drift between these two states can cause some bad race conditions. State more clearly:

If I open doc `A` for editing on lockbook-desktop and am typing away. And lockbook-cli does a sync which updates doc `B`, when I save my doc in lockbook-desktop, it will overwrite this change. Now this particular issue will be solved by #558. But there's all sorts of pathological race conditions that can arise if multiple clients are sharing a lockbook working directory.

For a while I thought having multiple lockbook directories was a suitable solution to this problem, however 2 use cases made me think otherwise:

+ CLI being out of sync with the desktop app strikes me as a poor UX. I edit a file in desktop, I have to sync to see it in vim? I'll need to be online for that, I already have the data. Do I copy and paste this? Sounds awful.
+ In a future where people are saving photos and larger files or just generally have large lockbook they may want to select which folders get synced to what devices. They may conclude that at most, the laptop instance of lockbook should take up 2gb. It would be unfortunate if the CLI replicated this data.

So multiple folders are not the move. The way `git`, `cargo` and friends solve this problem is via file locks. There are poor implementations (which require you to go manually delete lock files) and there are better process-level implementations which clear the lock if your process panics, etc.

While the linux client is open, it will likely grab a file lock on the whole lb working directory (because of the variety of info it's showing), and if you try to use the CLI it will tell you that you have to close X PID (desktop) before you can do that, and then just wait (so you don't lose work). 

This may grow to be more fine grained in the future (lock per file id for instance) where only specific operations lock the whole directory (like sync). This sort of logic does not depend on DI & the lock handle would likely end up inside `Context` whether that lives in static space or is hung on to by the consumer of the core library.

Parth:
# Create a [workspace](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html)

+ Creates a single `.lock` file, a shared cache for all projects
+ Tooling will understand the relationships between the components of our project a bit better
+ Will be able to speed up CI (building cli & core will become 1 step), but we'll do this later
+ Fixed broken benchmarks that @raayan authored, now they all get compiled as part of the same target

# Remove DI where it is not needed

+ See #774 for context (fixes #774) 
+ A large component of the lines deleted

# Create a `test_utils` dev-dep

+ Cleanup all the test related helpers and make them available for all tests, but only tests
+ fixes #741


containers/Dockerfile.core
```diff
@@ -1,5 +1,5 @@
 ARG HASH
-FROM rust AS core-build
+FROM rust:1.52 AS core-build
```
Parth, line 2:
```suggestion
FROM rust AS core-build
```

tvanderstad (approved):

PR #782, opened 2021-05-30
core: make file names secrets

### LINKED ISSUE #763: Encrypt file names in a way that server can maintain integrity

Parth:
# `SecretFileName`

Make file names secret in a way that server can still make sure all paths are unambiguous while not knowing anything about the names of files.

How I arrived at this particular strategy:

A naive approach may be to symmetrically encrypt, however `aes-gcm` requires a unique nonce, reusing the nonce leaks a whole bunch of information. So perhaps the encrypted value could be kept alongside a hash of the secret? This would still let the server administrator do unsavory things like: "find all users that have a document named `guns.md`". You could add something to the hash that only a user would know, maybe the hash of their secret key or something like that. This sounds like a nightmare for sharing use cases. Sharees do have exclusive access to the keys that encrypt documents and folders, maybe we can use that as a source of "randomness" that's reproducible that the server doesn't have access to, that everyone who's interested in a particular document does have access to. More specifically if we have a document, and we want to hide it's name, we can use it's parent key as the source of that randomness, that will ensure that if another document is renamed to that name, collisions will be detectable without the server receiving any additional information.

The formal process for hashing things with a key is an `hmac`

So what the server receives:
+ the filename encrypted with the parents access key.
+ an `hmac(parent_key, file_name)`

# `ClientFileMetadata`

 This PR is part of a group of security updates that are going to have core test more things (and trust the server less). Core is going to verify (during sync) that things like delete came from someone that had access to the file. Server will not be able to forge deletion commands or rename commands for instance. As `FileMetadata` becomes more complicated it grows farther from what clients want to be dealing with.

This PR therefore additionally introduces a `ClientFileMetadata` that just contains the fields clients will care about (`name`, `id`, `parent`, etc). This will result in less ffi churn as we update these fields, and will represent the "plain-text" variant of `FileMetadata`.

fixes #763


tvanderstad (dismissed):

core/src/lib.rs
```diff
@@ -550,7 +534,7 @@ pub fn rename_file(
         }
         DocumentRenameError::FileNameNotAvailable => UiError(RenameFileError::FileNameNotAvailable),
         DocumentRenameError::CannotRenameRoot => UiError(RenameFileError::CannotRenameRoot),
-        DocumentRenameError::DbError(_) | DocumentRenameError::FailedToRecordChange(_) => {
+        _ => {
```
tvanderstad, line 537: is that... a wildcard match arm? who are you and what happened to the real parth

core/src/repo/file_metadata_repo.rs
```diff
@@ -639,151 +339,151 @@ mod unit_tests {
...
-
-        let root = insert_test_metadata_root(config, "uhoh");
-        let doc = insert_test_metadata(config, FileType::Document, root.id, "a");
-        let _ = insert_test_metadata(config, FileType::Document, doc.id, "b");
-
-        let probs = file_metadata_repo::test_repo_integrity(config).unwrap();
-        assert_eq!(probs.len(), 1);
-        assert!(probs.contains(&Problem::DocumentTreatedAsFolder(doc.id)));
-    }
+    // #[test]
```
tvanderstad, line 342: I can't say I haven't done one of these

core/src/repo/file_metadata_repo.rs
```diff
@@ -793,38 +493,32 @@ mod unit_tests {
...
-            "child doc",
         );
 
         {
-            let mut children_of_root =
+            let children_of_root =
                 file_metadata_repo::get_and_get_children_recursively(config, root.id).unwrap();
-            children_of_root.sort_by(|f1, f2| f1.name.cmp(&f2.name));
-            assert_eq!(children_of_root, vec![doc.clone(), root.clone()]);
+            // TODO assert specific children here.
```
tvanderstad, line 507: commenting just in case you didn't mean to leave this one for later

core/src/service/mod.rs
```diff
@@ -4,6 +4,8 @@ pub mod drawing_service;
 pub mod file_compression_service;
 pub mod file_encryption_service;
 pub mod file_service;
+pub mod integrity_service;
```
tvanderstad, line 7: (now that's a service I can get behind)

core/src/service/mod.rs
```diff
@@ -4,6 +4,8 @@ pub mod drawing_service;
...
+pub mod path_service;
```
tvanderstad, line 8: (I also celebrate the existence of this mod)

Parth:
image
A good overview of our current status

core/src/repo/file_metadata_repo.rs
Parth: fixed!

core/src/repo/file_metadata_repo.rs
Parth: This is for a future test cleanup

core/src/lib.rs
Parth: haha I'm going to make this the norm in a future PR

tvanderstad (approved):

PR #786, opened 2021-06-05
android: share documents

SmailBarkouch:
In this PR, I added the ability to share documents (including drawings). A user simply selects the files he wants to share, clicks the share icon on the menubar, and is given the standard share screen. If a user selects a folder, it will recursively get every document within that folder.

On android, to share a file you need to provide a `Uri`. That means there is no way to simply give bytes for something like an image, so you need to save it to some place. This is troublesome since it would just be better if it wasn't saved, but I ended up saving it within the local cache. This cache is private and only accessible to the app, and is cleared everytime the user exits lockbook. The images saved there momentarily will be not be visible in the user's gallery.

The wait time to process a document (like generating an image) and saving it to disk is short since bytes and text aren't passed over ffi and then saved by the clients. Instead, they are saved with core using two new endpoints:
- `save_document_to_disk`
- `export_drawing_to_disk`

In these endpoints you provide a location on disk where you want a file to be saved with the contents of a document/image.

### Screen Recordings
<details>

Sharing different 6 drawings and regular documents:

https://user-images.githubusercontent.com/20663038/121783834-e304ea00-cb7e-11eb-8f19-c583fc3bd307.mp4
</details>

### DIFF
--- a/core/src/java_interface.rs
+++ b/core/src/java_interface.rs
@@ -14,9 +14,10 @@ use crate::service::sync_service::SyncProgress;
 use crate::service::usage_service::bytes_to_human;
 use crate::{
     calculate_work, create_account, create_file, delete_file, export_account, export_drawing,
-    get_account, get_all_error_variants, get_children, get_db_state, get_file_by_id,
-    get_local_and_server_usage, get_root, import_account, init_logger, insert_file, migrate_db,
-    move_file, read_document, rename_file, set_last_synced, sync_all, write_document, Error,
+    export_drawing_to_disk, get_account, get_all_error_variants, get_children, get_db_state,
+    get_file_by_id, get_local_and_server_usage, get_root, import_account, init_logger, insert_file,
+    migrate_db, move_file, read_document, rename_file, save_document_to_disk, set_last_synced,
+    sync_all, write_document, Error,
 };
 use basic_human_duration::ChronoHumanDuration;
 use chrono::Duration;
@@ -440,6 +441,7 @@ pub extern "system" fn Java_app_lockbook_core_CoreKt_readDocument(
         Ok(ok) => ok,
         Err(err) => return err,
     };
+
     let id = match deserialize_id(&env, jid) {
         Ok(ok) => ok,
         Err(err) => return err,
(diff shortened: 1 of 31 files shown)


tvanderstad:
>Most of the wait time is from processing the image (roughly 95% to 99% from my tests).
Processing how? e.g. rendering, saving, something else

SmailBarkouch:
> > Most of the wait time is from processing the image (roughly 95% to 99% from my tests).
> 
> Processing how? e.g. rendering, saving, something else

It was from passing the bytes over ffi, but this has since been resolved through two new endpoints. It is much much quicker. My bad, I forgot to update the description to include this new stuff but it is updated now.

core/src/java_interface.rs
```diff
@@ -449,6 +453,104 @@ pub extern "system" fn Java_app_lockbook_core_CoreKt_readDocument(
...
 
+#[no_mangle]
+pub extern "system" fn Java_app_lockbook_core_CoreKt_saveDocumentToDisk(
+    env: JNIEnv,
+    _: JClass,
+    jconfig: JString,
+    jid: JString,
+    jlocation: JString,
+) -> jstring {
+    let config = match deserialize::<Config>(&env, jconfig, "Couldn't successfully get config") {
```
Parth, line 464:
```suggestion
    let config = match deserialize::<Config>(&env, jconfig, "Couldn't deserialize config") {
```

core/src/java_interface.rs
```diff
@@ -449,6 +453,104 @@ pub extern "system" fn Java_app_lockbook_core_CoreKt_readDocument(
...
+        Ok(ok) => ok,
+        Err(err) => return err,
+    };
+
+    let id = match deserialize_id(&env, jid) {
+        Ok(ok) => ok,
+        Err(err) => return err,
+    };
+
+    let location = match jstring_to_string(&env, jlocation, "Couldn't successfully get path") {
```
Parth, line 474:
```suggestion
    let location = match jstring_to_string(&env, jlocation, "Couldn't parse jstring: path") {
```

Parth: Same below

Parth (approved):

Parth:
fixes #785

PR #793, opened 2021-06-21
linux: could not reach server dialog does not appear when computer is suspended

### LINKED ISSUE #760: Get usage with upper bound linux

SmailBarkouch:
Originally, when you suspended your laptop you may have been greeted with a "Could not reach server" dialog. To counter this, we now have a bottom bar where messages will be shown regarding sync. Depending on the result of the sync, as well as the host of other things, a message will be displayed. All in all, this PR is made to be in accordance with [this](https://github.com/lockbook/lockbook/blob/master/docs/design-ux/navigation.md).

fixes #760

### DIFF
--- a/clients/linux/src/app.rs
+++ b/clients/linux/src/app.rs
@@ -130,8 +130,11 @@ impl LbApp {
                     lb.err_dialog(&title, &err);
                     Ok(())
                 }
-                Msg::ErrorStatusPanel(msg) => {
-                    lb.gui.account.status().set_status(msg.as_str(), None);
+                Msg::SetStatus(txt, tool_tip_txt) => {
+                    lb.gui
+                        .account
+                        .status()
+                        .set_status(txt.as_str(), tool_tip_txt.as_deref());
                     Ok(())
                 }
             };
@@ -318,13 +321,14 @@ impl LbApp {
     }
 
     fn refresh_usage_status(&self) -> LbResult<()> {
-        let status = self.core.usage_status()?;
-        if let (Some(txt), _) = status {
-            self.gui
-                .account
-                .status()
-                .set_status(&txt, status.1.as_deref());
-        }
+        spawn!(self.core as c, self.messenger as m => move || {
+            match c.usage_status() {
+                Ok(status) => if let (Some(txt), _) = status {
+                    m.send(Msg::SetStatus(txt, status.1));
+                }
+                Err(err) => m.send_err_status_panel(err.msg())
+            }
+        });
 
         Ok(())
     }
@@ -892,7 +896,7 @@ impl LbApp {
     }
 
     fn show_dialog_usage(&self) -> LbResult<()> {
-        let usage_string = self.core.usage_human_string()?;
+        let usage_string = self.core.get_usage()?.server_usage.readable;
         let usage = usage(usage_string)?;
         let d = self.gui.new_dialog("My Lockbook Usage");
         d.get_content_area().add(&usage);
--- a/clients/linux/src/app.rs
+++ b/clients/linux/src/app.rs
@@ -318,7 +318,10 @@ impl LbApp {
         spawn!(self.core as c, self.messenger as m => move || {
             match c.sync_status() {
                 Ok(txt) => m.send(Msg::SetStatus(txt, None)),
-                Err(err) => m.send_err_status_panel(err.msg())
+                Err(err) => match err.target() {
+                    LbErrTarget::Dialog => m.send_err_dialog("getting sync status", err),
+                    LbErrTarget::StatusPanel => m.send_err_status_panel(err.msg()),
+                }
             }
         });
 
@@ -331,7 +334,10 @@ impl LbApp {
                 Ok(status) => if let (Some(txt), _) = status {
                     m.send(Msg::SetStatus(txt, status.1));
                 }
-                Err(err) => m.send_err_status_panel(err.msg())
+                Err(err) => match err.target() {
+                    LbErrTarget::Dialog => m.send_err_dialog("getting usage status", err),
+                    LbErrTarget::StatusPanel => m.send_err_status_panel(err.msg()),
+                }
             }
         });
 
--- a/clients/linux/src/app.rs
+++ b/clients/linux/src/app.rs
@@ -30,7 +30,7 @@ use crate::background_work::BackgroundWork;
 use crate::editmode::EditMode;
 use crate::error::{
     LbErrKind::{Program as ProgErr, User as UserErr},
-    LbError, LbResult,
+    LbErrTarget, LbError, LbResult,
 };
 use crate::filetree::FileTreeCol;
 use crate::intro::{IntroScreen, LOGO_INTRO};
@@ -38,9 +38,10 @@ use crate::menubar::Menubar;
 use crate::messages::{Messenger, Msg};
 use crate::settings::Settings;
 use crate::util;
-use crate::{closure, progerr, tree_iter_value, uerr};
+use crate::{closure, progerr, tree_iter_value, uerr, uerr_dialog};
 use lockbook_core::model::client_conversion::ClientFileMetadata;
 use std::thread;
+use std::time::Duration;
 
 macro_rules! make_glib_chan {
     ($( $( $vars:ident ).+ $( as $aliases:ident )* ),+ => move |$param:ident :$param_type:ty| $fn:block) => {{
@@ -95,6 +96,7 @@ impl LbApp {
                 Msg::ExportAccount => lb.export_account(),
                 Msg::PerformSync => lb.perform_sync(),
                 Msg::RefreshSyncStatus => lb.refresh_sync_status(),
+                Msg::RefreshUsageStatus => lb.refresh_usage_status(),
                 Msg::Quit => lb.quit(),
 
                 Msg::AccountScreenShown => lb.account_screen_shown(),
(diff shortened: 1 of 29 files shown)


Parth (changes requested):
The "bottom bar" should still only be a submember of the file tree. It seems to stretch across the entire app now, and there's lots of dead space.

clients/linux/src/backend.rs
```diff
@@ -342,16 +361,69 @@ impl LbCore {
...
+                let total_usage = match usage.server_usage.find(' ') {
+                    None => Err(progerr!("Could not parse total usage.")),
+                    Some(index) => usage.server_usage[..index]
+                        .parse::<f32>()
+                        .map_err(|err| progerr!("Could not parse total usage: {:?}", err)),
+                }?;
+
+                if total_usage / data_cap > USAGE_WARNING_THRESHOLD {
+                    return Ok(format!(
+                        "Warning, you are currently over {}% usage.",
```
Parth, line 415: I think this should say:
`n` _b of `m` _b remaining!

When they mouse over it should say:
You are running out of space, go to settings to buy more!

@tvanderstad something to note for UX docs, there's already some stuff spec'd out in billing.md. But it's probably a good time to start thinking about these flows.

clients/linux/src/backend.rs
```diff
@@ -342,16 +361,69 @@ impl LbCore {
...
+                            "<b>{}</b>  file{} not synced.",
+                            n_files,
+                            if n_files == 1 { "" } else { "s" }
+                        ),
+                    });
+                }
+
+                let usage = self.local_and_server_usage()?;
+
+                let data_cap = match usage.data_cap.find(' ') {
```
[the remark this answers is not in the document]
Parth: Then you don't have to parse strings, deal with errors, and deal with strange behavior if we ever change how these strings do their business.

clients/linux/src/backend.rs
```diff
@@ -342,16 +361,69 @@ impl LbCore {
...
+                }?;
+
+                let total_usage = match usage.server_usage.find(' ') {
+                    None => Err(progerr!("Could not parse total usage.")),
+                    Some(index) => usage.server_usage[..index]
+                        .parse::<f32>()
+                        .map_err(|err| progerr!("Could not parse total usage: {:?}", err)),
+                }?;
+
+                if total_usage / data_cap > USAGE_WARNING_THRESHOLD {
```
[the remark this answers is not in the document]
Parth: Actually if you edit a file I see this message. My usage is less than 1% still.
image

clients/linux/src/backend.rs
```diff
@@ -342,16 +361,69 @@ impl LbCore {
...
+                        0 => "✔  Synced.".to_string(),
+                        _ => format!(
+                            "<b>{}</b>  file{} not synced.",
+                            n_files,
+                            if n_files == 1 { "" } else { "s" }
+                        ),
+                    });
+                }
+
+                let usage = self.local_and_server_usage()?;
```
Parth, line 397: I don't think this stuff should end up here because it's not related to the "sync status". I think this may be the trickiest bit of this PR. Here are the tricky bits:

+ At what point is it appropriate to call this?
+ How do we handle competing information?
+ What portions of core do we have to modify to make this a better experience for a linux developer?

Here are some considerations:
+ set_sync_status is called every time a file is edited, this means that we're recomputing usage every time, which involves a round trip to the server. This current version is also [calculating my compression ratio](https://github.com/lockbook/lockbook/blob/master/core/src/service/usage_service.rs#L59), which is *really* expensive.
+ To the same UI element we're going to be sending competing information. Sometimes we'll send that a file is synced, or is dirty or that the user is approaching their cap, etc. There is an ordering of importance (as described in that doc), what's the best way to hang on to these competing pieces of data and show the most relevant one to the user.
+ So you'll likely only want to calculate usage when a sync is finished, you'll likely not want to calculate the compression ratio, you'll definitely not want to parse strings.

When set_sync_status and set_usage_status become two separate entities (because they're called at different times) the logic about what info is shown at the current moment moves somewhere else. Probably somewhere closer to the UI, and as this is a UI decision, it's probably a good one.

This may get pushed out into a separate PR if you find it too tangential, but: the final thing I want you to think about while you sort this out: I don't want to open the app and for a moment see "Synced!" and then see "Syncing...." and then see upto date. If the init step is to kick off a sync, then initially I should see "Syncing...". Ideally the user never sees something for a split second and is left wondering what that said. In the app the only place this sort of complexity and fine tuning is needed is this sync_status / bottom bar region. By fine tuning I mean, if we show a sync progress bar, maybe we delay the sync operation starting by 0.25s and we delay it's completion by 0.25 seconds so we ensure that "Syncing..." stays on screen for at least half a second. This is probably something that should happen everywhere (@tvanderstad)

Parth (changes requested):

core/src/service/usage_service.rs
```diff
@@ -22,22 +21,56 @@ pub const GIGABYTE_PLUS_ONE: u64 = GIGABYTE + 1;
 pub const TERABYTE_PLUS_ONE: u64 = TERABYTE + 1;
```
Parth, line 21: When I get usage I see this:
```
Uncompressed File Size: 19.289 MB
Server Utilization: 6.427 MB
Server Data Cap: 1000.000 KB
```

I think we should probably tweak the limits of the human duration stuff. I should never see 1000.000 KB.

We should see 1.000 MB (tweak the upper bounds here). But I should really see 1 MB (don't show trailing 0's).

You may encounter a lil bit of ugliness while doing this last step: https://stackoverflow.com/questions/59506403/how-to-format-a-float-without-trailing-zeros-in-rust

clients/linux/src/backend.rs
```diff
@@ -356,8 +360,26 @@ impl LbCore {
...
+        }
+
+        Ok((None, None))
+    }
 }
 
 const UNAME_REQS: &str = "letters and numbers only";
 const STATE_REQ_CLEAN_MSG: &str =
     "Your local state cannot be migrated, please re-sync with a fresh client.";
+const USAGE_WARNING_THRESHOLD: f32 = 0.0001;
```
Parth, line 385: This still looks not right

clients/linux/src/app.rs
```diff
@@ -196,9 +208,13 @@ impl LbApp {
             if let Some(msg) = msgopt {
                 lb.gui.intro.sync_progress(&msg)
             } else if let Err(err) = lb.gui.show_account_screen(&lb.core) {
-                lb.messenger.send_err("showing account screen", err);
+                lb.messenger.send_err_dialog("showing account screen", err);
             } else {
                 lb.messenger.send(Msg::RefreshSyncStatus);
```
Parth, line 213: This is a network call, it should happen in it's own thread. This will lock up the UI while they're typing for someone with a slow network connection.

Parth (changes requested):
While you're tweaking these usage bits, have a look at this as well: https://github.com/lockbook/lockbook/issues/760.

In PR #636 I asked you to remove the upper bound of the progress bar because I thought billing would happen continuously. But that is no longer the case and we have an upper limit now. While you're in the usage realm, I would restore this progress bar and that issue.

clients/linux/src/app.rs
```diff
@@ -207,7 +220,10 @@ impl LbApp {
         // used to receive progress updates as indicated above.
         spawn!(self.core as c, self.messenger as m => move || {
             if let Err(err) = c.sync(sync_chan) {
-                m.send_err("syncing", err);
+                match err.target() {
+                    LbErrTarget::Dialog => m.send_err_dialog("syncing", err),
```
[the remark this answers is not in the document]
SmailBarkouch: It depends on the specified `target` of the error. If it is `LbErrTarget::Dialog`, then it will show up in a dialog, and if it is the `LbErrTarget::StatusPanel`, it will show up in the status panel.

Parth (changes requested):

core/src/service/usage_service.rs
```diff
@@ -16,28 +15,62 @@ pub const MEGABYTE: u64 = KILOBYTE * 1000;
...
+    pub data_cap: UsageItemMetric,
+}
+
+#[derive(Serialize)]
+pub struct UsageItemMetric {
+    pub exact: u64,
+    pub readable_exact: String,
+    pub readable: String,
+    pub unit: ByteUnit,
+}
```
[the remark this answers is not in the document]
Parth: Also let's get rid of `readable_exact` and `unit`

clients/linux/src/app.rs
```diff
@@ -124,13 +126,23 @@ impl LbApp {
                 Msg::ToggleAutoSave(auto_save) => lb.toggle_auto_save(auto_save),
                 Msg::ToggleAutoSync(auto_sync) => lb.toggle_auto_sync(auto_sync),
 
-                Msg::Error(title, err) => {
-                    lb.err(&title, &err);
+                Msg::ErrorDialog(title, err) => {
+                    lb.err_dialog(&title, &err);
```
Parth, line 130: You should do these in their own `fn` like the other ones

Parth (approved):

core/src/lib.rs
```diff
@@ -641,34 +640,17 @@ pub enum GetUsageError {
...
         _ => unexpected!("{:#?}", e),
     })
 }
 
-pub fn get_local_and_server_usage(
-    config: &Config,
-    exact: bool,
-) -> Result<LocalAndServerUsages, Error<GetUsageError>> {
-    usage_service::local_and_server_usages(&config, exact).map_err(|e| match e {
+pub fn get_uncompressed_usage(config: &Config) -> Result<UsageItemMetric, Error<GetUsageError>> {
```
Parth, line 652: We shouldn't group these errors

Issue #805, opened 2021-07-05
Add a CLI command for verifying repo integrity


LucaLoncarGoog:
1. Call `test_repo_integrity()`
2. Check if any documents are empty/unreadable, but do not do so at test time

PR #806, opened 2021-07-06
cli: add an integrity check

LucaLoncarGoog:
Add a cli command that will check the integrity of your lockbook repository.

fixes: #805

### DIFF
--- a/clients/cli/src/main.rs
+++ b/clients/cli/src/main.rs
@@ -145,7 +145,7 @@ enum Lockbook {
     #[structopt(name = "whoami")]
     WhoAmI,
 
-    /// Validate Lockbook state
+    /// Find lockbook file structure problems, corrupted or missing files.
     Validate,
 }
 
--- a/clients/cli/src/main.rs
+++ b/clients/cli/src/main.rs
@@ -23,6 +23,7 @@ mod rename;
 mod status;
 mod sync;
 mod utils;
+mod validate;
 mod whoami;
 
 #[derive(Debug, PartialEq, StructOpt)]
@@ -143,6 +144,9 @@ enum Lockbook {
     /// Display Lockbook username
     #[structopt(name = "whoami")]
     WhoAmI,
+
+    /// Validate Lockbook state
+    Validate,
 }
 
 fn main() {
@@ -175,6 +179,7 @@ fn main() {
         Lockbook::Status => status::status(),
         Lockbook::Sync => sync::sync(),
         Lockbook::WhoAmI => whoami::whoami(),
+        Lockbook::Validate => validate::validate(),
         Lockbook::Backup => backup::backup(),
         Lockbook::GetUsage { exact } => calculate_usage::calculate_usage(exact),
         Lockbook::ExportDrawing { path, format } => export_drawing::export_drawing(&path, &format),
--- a/clients/cli/src/validate.rs
+++ b/clients/cli/src/validate.rs
@@ -1,0 +1,66 @@
+use crate::error::CliResult;
+use crate::utils::{get_account_or_exit, get_config};
+use crate::{err, err_unexpected};
+use lockbook_core::model::state::Config;
+use lockbook_core::service::integrity_service::{test_repo_integrity, TestRepoError, Warning};
+use lockbook_core::get_path_by_id;
+use uuid::Uuid;
+
+pub fn validate() -> CliResult<()> {
+    get_account_or_exit();
+
+    let config = get_config();
+
+    let err = match test_repo_integrity(&config, false) {
+        Ok(warnings) => {
+            if warnings.len() == 0 {
+                return Ok(());
+            };
+
+            for w in &warnings {
+                match w {
+                    Warning::EmptyFile(id) => {
+                        let path = get_path_by_id_or_err(&config, *id)?;
+                        eprintln!("File at path {} is empty.", path);
+                    }
+                    Warning::InvalidUTF8(id) => {
+                        let path = get_path_by_id_or_err(&config, *id)?;
+                        eprintln!("File at path {} contains invalid UTF8.", path);
+                    }
+                }
+            }
+            err!(WarningsFound(warnings.len() as i32))
+        }
+        Err(err) => match err {
+            TestRepoError::NoRootFolder => err!(NoRootOps("validate")),
+            TestRepoError::DocumentTreatedAsFolder(id) => {
+                err!(DocTreatedAsFolder(get_path_by_id_or_err(&config, id)?))
+            }
+            TestRepoError::FileOrphaned(id) => {
+                err!(FileOrphaned(get_path_by_id_or_err(&config, id)?))
+            }
+            TestRepoError::CycleDetected(_) => {
+                err!(CycleDetected)
+            }
+            TestRepoError::FileNameEmpty(_) => {
+                err!(FileNameEmpty)
+            }
+            TestRepoError::FileNameContainsSlash(id) => {
+                err!(FileNameHasSlash(get_path_by_id_or_err(&config, id)?))
+            }
+            TestRepoError::NameConflictDetected(id) => {
+                err!(NameConflictDetected(get_path_by_id_or_err(&config, id)?))
+            }
+            TestRepoError::Core(err) => {
+                err_unexpected!("an unexpected error occurred: {:#?}", err)
+            }
+        },
+    };
+
+    Err(err)
+}
+
+fn get_path_by_id_or_err(config: &Config, id: Uuid) -> CliResult<String> {
+    get_path_by_id(config, id)
+        .map_err(|err| err_unexpected!("failed to get path by id: {} err: {:#?}", id, err))
+}
--- a/clients/cli/src/validate.rs
+++ b/clients/cli/src/validate.rs
@@ -13,7 +13,7 @@ pub fn validate() -> CliResult<()> {
 
     let err = match test_repo_integrity(&config, false) {
         Ok(warnings) => {
-            if warnings.len() == 0 {
+            if warnings.is_empty() {
                 return Ok(());
             };
 
--- a/clients/cli/src/validate.rs
+++ b/clients/cli/src/validate.rs
@@ -1,9 +1,9 @@
 use crate::error::CliResult;
 use crate::utils::{get_account_or_exit, get_config};
 use crate::{err, err_unexpected};
+use lockbook_core::get_path_by_id;
 use lockbook_core::model::state::Config;
 use lockbook_core::service::integrity_service::{test_repo_integrity, TestRepoError, Warning};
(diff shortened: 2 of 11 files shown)


clients/cli/src/validate.rs
```diff
@@ -0,0 +1,56 @@
...
+            // for w in warnings {
+            //     warning_ids.push(w.uuid);
+            // }
+            // (warning_ids, warnings)
+            err!(DocTreatedAsFolder(String::from("DEBUG")))
+        }
+        Err(err) => match err {
+            TestRepoError::NoRootFolder => err!(NoRootOps("validate")),
+            TestRepoError::DocumentTreatedAsFolder(id) => {
+                err!(DocTreatedAsFolder(get_path_by_id_or_err(&config, id)?))
```
Parth, line 28: There's a few things going on here:

First you do this: `get_path_by_id_or_err(&config, id)?` which either converts or returns early with an error.

Then an error enum is created with that value: `err!(DocTreatedAsFolder(...))`.

Nothing happens with this error though. You can `return` it (not sure if `?` works here you can try that). But what's a more common pattern [for instance in `move`](https://github.com/lockbook/lockbook/blob/master/clients/cli/src/move_file.rs) is to leave this `match` statement as the last line of this `fn`.

Currently your last line is `Ok(())` which is what's causing this value to be ignored.

All you have to do is delete that line, and this code will compile.

Parth (changes requested):

clients/cli/src/validate.rs
LucaLoncarGoog: I actually do get the same compilation error when I remove `Ok(())` -- I'll look into your other suggestions though

LucaLoncarGoog: Hmm, still seeing this last error:

```
error[E0618]: expected function, found enum variant `$crate::error::ErrorKind::$err`
   --> clients/cli/src/error.rs:173:38
    |
45  |             $( $variants $( ( $( $types ),* ) )?, )*
    |                --------- `$crate::error::ErrorKind::$err` defined here
...
173 |         $crate::error::Error::Simple($crate::error::ErrorKind::$err $( ( $( $args ),+ ) )?)
    |                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^------------------------
    |                                                                |
    |                                                                call expression requires function
    | 
   ::: clients/cli/src/validate.rs:34:17
    |
34  |                 err!(CycleDetected(get_path_by_id_or_err(&config, id)?))
    |                 -------------------------------------------------------- in this macro invocation
    |
    = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
help: `$crate::error::ErrorKind::$err` is a unit variant, you need to write it without the parenthesis
    |
173 |         $crate::error::Error::Simple($crate::error::ErrorKind::$crate::error::ErrorKind::$err )?)
    |                                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

clients/cli/src/validate.rs
```diff
@@ -0,0 +1,71 @@
...
+                    Warning::InvalidUTF8(id) => {
+                        let path = get_path_by_id_or_err(&config, id)?;
+                        warning_message.push_str(&*format!("file at path {} is contains invalid UTF8, ", path));
+                    }
+                }
+            }
+            // Remove the last two characters ', '
+            warning_message.pop();
+            warning_message.pop();
+            return Err(err!(warning_message));
```
LucaLoncarGoog, line 38: This is the last error I get 

```
error[E0599]: no variant or associated item named `warning_message` found for enum `error::ErrorKind` in the current scope
  --> clients/cli/src/validate.rs:38:29
   |
38 |             return Err(err!(warning_message));
   |                             ^^^^^^^^^^^^^^^ variant or associated item not found in `error::ErrorKind`
   | 
  ::: clients/cli/src/error.rs:44:9
   |
44 |         pub enum ErrorKind {
   |         ------------------ variant or associated item `warning_message` not found here

error: aborting due to previous error; 3 warnings emitted
```

I don't really understand the issue, though, since I see strings being passed to this elsewhere (I get the same issue if I remove err!)

Parth: What you may be seeing is usages of `err_unexpected` which take a string.

Makefile
```diff
@@ -157,7 +157,7 @@ performance_bench: performance server db_container
 
 .PHONY: performance_bench_report
 performance_bench_report: is_docker_running
-	docker container cp "$$(docker inspect --format="{{.Id}}" performance-performance-$(hash))":/core/simple-create_write_read.svg .
+	docker container cp "$$(docker inspect --format="{{.id}}" performance-performance-$(hash))":/core/simple-create_write_read.svg .
```
Parth, line 160: seems like you have a rogue search and replace that's changed a few files you didn't intend to change.

Parth (changes requested):

clients/cli/src/validate.rs
LucaLoncarGoog: It seems err_unexpected also wants an error -- what do you think we should do here? I'm not sure it makes sense to add something new to errors.rs, since these warnings aren't exactly errors.  Can we not just return generic errors with arbitrary string messages?

Also, once I get this done, where shall I define integration tests?

Parth: You can `eprintln` arbitrary things. Let's handle the warnings that way. I think the warnings should be newline separated, so you won't have to process the remaining `", "`.

Then you can return one error like `WarningsFound`, because we still want to return a non-zero status code.

Parth: `core/tests` a good place to define expressive integration tests. I think the main things you want to test here is the validations, you don't have to test the CLI bits.

LucaLoncarGoog: Okay, I feel like I'm almost in a pretty good state -- it compiles, but now my only issue is trying to write a test.  When I run existing tests in core/tests/integrity_tests.rs I get

```
called `Result::unwrap()` on an `Err` value: UiError(CouldNotReachServer)
thread 'integrity_tests::test_name_conflict' panicked at 'called `Result::unwrap()` on an `Err` value: UiError(CouldNotReachServer)'
```

core/src/service/path_service.rs
```diff
@@ -148,6 +149,27 @@ pub fn get_all_paths(config: &Config, filter: Option<Filter>) -> Result<Vec<Stri
...
+        let name = file_encryption_service::get_name(&config, &current_metadata)?;
+        path = format!("{}/{}", name, path);
+        current_id = current_metadata.parent;
+        current_metadata = file_metadata_repo::get(config, current_id)?
+    }
+
+    {
+        let name = file_encryption_service::get_name(&config, &current_metadata)?;
+        path = format!("{}/{}", name, path);
+    }
```
Parth, line 167: maybe you'd be better off modeling it this way: https://github.com/rust-lang/rfcs/issues/1313

core/src/service/path_service.rs
```diff
@@ -148,6 +149,27 @@ pub fn get_all_paths(config: &Config, filter: Option<Filter>) -> Result<Vec<Stri
...
+    // Remove the last forward slash.
+    path.pop();
```
Parth, line 169: If you're only using `"{}/{}"` where does the trailing forward slash come from?

Also I think if the last element in a path is a folder, we want that trailing slash.

core/src/service/integrity_service.rs
```diff
@@ -24,7 +25,16 @@ pub enum TestRepoError {
     Core(CoreError),
 }
 
-pub fn test_repo_integrity(config: &Config) -> Result<(), TestRepoError> {
+#[derive(Debug, Clone)]
+pub enum Warning {
+    EmptyFile(Uuid),
+    InvalidUTF8(Uuid),
```
Parth, line 31: One check missing here, could be my fault for not tracking it anywhere:

if we're dealing with a `.draw` attempt to call the corresponding read_drawing sort of function (either in `drawing_service` or `lib.rs` and if the drawing is invalid, report that as it's own warning.

Parth (changes requested):

clients/cli/src/validate.rs
```diff
@@ -0,0 +1,66 @@
...
+            }
+            err!(WarningsFound(warnings.len() as i32))
+        }
+        Err(err) => match err {
+            TestRepoError::NoRootFolder => err!(NoRootOps("validate")),
+            TestRepoError::DocumentTreatedAsFolder(id) => {
+                err!(DocTreatedAsFolder(get_path_by_id_or_err(&config, id)?))
+            }
+            TestRepoError::FileOrphaned(id) => {
+                err!(FileOrphaned(get_path_by_id_or_err(&config, id)?))
```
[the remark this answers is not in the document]
LucaLoncarGoog: What is the value of .parent if the file is orphaned? The default value for Uuid? Which is what? get_path_by_id should handle that, you're right.

core/src/service/integrity_service.rs
```diff
@@ -108,5 +118,25 @@ pub fn test_repo_integrity(config: &Config) -> Result<(), TestRepoError> {
...
+    for file in all.clone() {
+        if file.file_type == Document {
+            let file_content = file_service::read_document(&config, file.id).map_err(Core)?;
+
+            if file_content.len() as u64 == 0 {
+                warnings.push(Warning::EmptyFile(file.id));
+                continue;
+            }
+
+            if String::from_utf8(file_content).is_err() {
```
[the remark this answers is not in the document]
LucaLoncarGoog: What about, like, every other programming language's suffix lol

core/src/service/path_service.rs
LucaLoncarGoog: But I want to check the condition immediately

core/src/service/path_service.rs
LucaLoncarGoog: path is always empty at first, so there is always a trailing '/'.  I could, however, just not do this if it is a folder.

LucaLoncarGoog:
Will augment tests once we fully agree on implementation

clients/cli/src/main.rs
```diff
@@ -143,6 +144,9 @@ enum Lockbook {
     /// Display Lockbook username
     #[structopt(name = "whoami")]
     WhoAmI,
+
+    /// Validate Lockbook state
```
Parth, line 148:
```suggestion
    /// Find lockbook file structure problems, corrupted or missing files. 
```

core/src/service/integrity_service.rs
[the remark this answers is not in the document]
Parth: Yeah I mean as many as you'd like. If there's a better fundamental approach here for how to detect which files are generally encoded in `utf8` feel free to go in that direction instead.

Parth (approved):

PR #821, opened 2021-07-13
server: add API endpoint for reporting server build info

### LINKED ISSUE #820: Server version info
Via an endpoint, have the server return info about it’s current build version / build hash

plv:
* Adding a new API endpoint that returns package/build # and the corresponding git commit hash. These values are determined at build time.
  * Pulling in [`shadow-rs`](https://github.com/baoyachi/shadow-rs) as a dependency for server to compute git hash.

* I'm not sure what the testing story is here, so lmk how/if I can test this out without integrating into clients.

closes #820

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -479,6 +479,9 @@ name = "cc"
 version = "1.0.68"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4a72c244c1ff497a746a7e1fb3d14bd08420ecda70c8f25c7112f2781652d787"
+dependencies = [
+ "jobserver",
+]
 
 [[package]]
 name = "cesu8"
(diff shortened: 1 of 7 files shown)


Parth (approved):
Nice turnaround

There's two places tests end up:
+ `core/tests` -- the bulk of our integration tests are here
+ `server/tests` -- a newer collection of tests that use core, but are primarily interested in asserting properties about postgres/s3.

You could also throw `closes #issuenumber` in the desc, it'll associate this work with its issue and close it automatically when this PR is merged.

Parth:
Our dockerfiles are a lil wonky: https://github.com/lockbook/lockbook/blob/master/containers/Dockerfile.server. Cargo doesn't let us do a "build the dependency only step" (last time I checked), so the suggestion is to move a dummy `main.rs` / `lib.rs` to allow the dependency build cache to saturate.

We've added more and more cruft into this over time as you can see. Maybe there's a clean cargo extension that you can add that "dry"'s this up a bit. 

If you don't feel like solving this particular problem right now you can just move the `build.rs` you created into the container before the deps.

PR #825, opened 2021-07-16
linux: filetree tweaks and icons

### LINKED ISSUE #815: Bring iconography to linux
Places you can add it:
+ small thumbnails based on file types / extensions
+ in context menus for things like rename
+ in the menubar area for things Save it can have a floppy disk
+ settings can have a cog
+ etc

### LINKED ISSUE #816: During a move if you hover over a folder, pop it open
+ if the folder is closed, open it after half a second
+ if you drag away from that folder close it
+ if you drag into children of that folder that are folders, open those too

SmailBarkouch:
closes #815 
closes #816

### DIFF
--- a/clients/linux/src/filetree.rs
+++ b/clients/linux/src/filetree.rs
@@ -1,6 +1,6 @@
 use std::collections::HashMap;
 use std::rc::Rc;
-use std::sync::{mpsc, Arc};
+use std::sync::Arc;
 
 use gdk::{DragAction, DragContext, EventButton as GdkEventButton};
 use gdk::{EventKey as GdkEventKey, ModifierType};
@@ -15,7 +15,7 @@ use gtk::TreeStore as GtkTreeStore;
 use gtk::TreeView as GtkTreeView;
 use gtk::TreeViewColumn as GtkTreeViewColumn;
 use gtk::{
-    render_icon, CellRendererPixbuf, CellRendererPixbufBuilder, IconLookupFlags, IconSize,
+    CellRendererPixbuf, IconLookupFlags, IconSize,
     IconTheme, Menu as GtkMenu,
 };
 use gtk::{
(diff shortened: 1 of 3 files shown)


clients/linux/src/menubar.rs
```diff
@@ -174,17 +190,17 @@ impl Item {
     }
 
     #[rustfmt::skip]
-    fn data() -> Vec<(Self, ItemData)> {
+    fn data() -> Vec<(Self, Option<String>, ItemData)> {
```
steverusso, line 193: Why not do a static string (`&'static str`) type here instead of a `Option<String>`? Seems a bit cluttered

Parth (changes requested):
Okay tried it out locally, here are some thoughts:
+ remove icons from things that aren't the file tree

Can we do the following (maybe within it's own PR):
+ allow the filetree to scroll while doing a move?
+ un-expand things if you leave that folder's region?

Also maybe in it's own PR:
+ would a setting to toggle between symbolic icons require a restart? Maybe there's a good way to express which settings require a restart and which don't.

clients/linux/src/filetree.rs
```diff
@@ -187,13 +217,54 @@ impl FileTree {
...
+
+    fn on_drag_motion(
+        drag_hover_last_occurred: &Rc<RefCell<Option<u32>>>,
+        drag_ends_last_occurred: &Rc<RefCell<Option<u32>>>,
+    ) -> impl Fn(&TreeView, &DragContext, i32, i32, u32) -> GtkInhibit {
+        closure!(drag_hover_last_occurred, drag_ends_last_occurred => move |w, d, x, y, time| {
+            let vadj = w.get_vadjustment().unwrap();
+            let height = w.get_allocated_height();
+
+            if y > height - 50 || y < 50 {
```
Parth, line 238: Add some comments here

Parth (approved):

PR #830, opened 2021-07-27
server: allow un-auth'd requests for build info for monitoring purposes

### LINKED ISSUE #271: Page on call if server is offline
Relying on error logs isn't enough. If someone trips over a power chord over at digital ocean, we should be paged. 

Also page if version is incorrect.

Parth:
closes #271

### DIFF
--- a/.github/workflows/monitor.yml
+++ b/.github/workflows/monitor.yml
@@ -1,0 +1,13 @@
+on:
+  schedule:
+    - cron:  '* * * * *'
+jobs:
+  Server-Status:
+    runs-on: ubuntu-latest
+    env:
+      STATUS_API_URL: https://api.prod.lockbook.net/get-build-information
+      PD_API_KEY: ${{ secrets.PD_API_KEY }}
+    steps:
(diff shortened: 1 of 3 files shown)


plv:
What kind of monitoring purposes? Doesn't removing auth requirement here allow for request flooding of some sort?

Parth:
@plv Checkout ea7fc28 to see my motivation, I want to be paged if the server is ever offline. This is roughly the same number of CPU cycles as if someone were to flood any of the other non-matching endpoint, this request is handled entirely from memory without any IO, so attacking this is similar to attacking `/fjkslfjaslkfjsd`.

tvanderstad (approved):

plv:
SGTM, in the future might be better to provide Github Actions with a secret that it can use to authenticate itself, giving us the best of both worlds.

Parth:
@plv think about this endpoint tho -- evaluating that secret is roughly as much computation as evaluating an env var and returning the answer.

I agree however if we want this health check to talk to pg / s3.

PR #831, opened 2021-07-28
docs: sharing rfc

Parth:


docs/design-ux/sharing.md
```diff
@@ -0,0 +1,104 @@
...
+  lockbook and occassionally want to share a file with another user. Teams are small groups of users who generally want
+  someplace to store personal things and someplace to store team things, which are shared with the whole team.
+  Enterprises are large organizations that want to control how things are shared in sophisticated ways, such as having
+  teams with subteams that have at least as much access. The designs we consider are:
+
+1. Root sharing
+2. Folder sharing
+3. Policy Sharing
+
+## 1. Root Sharing
```
Parth, line 21: Honestly vault-sharing would simplify *SO MUCH* of core, maybe it's the complexity wins we need to get off the ground.

docs/design-ux/sharing.md
```diff
@@ -0,0 +1,104 @@
...
+### For Enterprises
+
+This model works okay for enterprises. In an enterprise, I would be part of multiple subteams e.g. my team, my
+department, and the whole organization. Because each set of sharees needs its own root, I would have one root for each
+team I am on. This makes it easy to share a file and know who it's shared with but difficult to organize across teams.
+For example if my team has technical documents that should be shared with an architecture council and cost estimation
+documents that should be shared with the budget team, we wouldn't be able to organize those in the same folder tree.
+
+## 2. Folder Sharing
+
```
Parth, line 63: I think this doesn't work well for absolute links. Under vault-sharing every file has an absolute link and you could link a `lb://babyloncove/readinglist.md` in slack. In folder sharing, either the links are subjective, or they leak information about *my* folder structure, right?

docs/design-ux/sharing.md
Parth: This also performs well for links, every document has an absolute path within lockbook.

docs/design-ux/sharing.md
tvanderstad: yeah one way or another it's a loss for links

tvanderstad: well, there's always permalinks e.g. `lb://123e4567-e89b-12d3-a456-426614174000`

docs/design-ux/sharing.md
tvanderstad: How can we address the concerns for the individuals use cases? Can we have vault sharing as described + individual document sharing by permalink that shows up in a 'shared with me' section or some such hack? Weird to have 2 sharing models but they're both simple enough that it's still a simplification and we can implement one at a time to get off the ground

Parth: Yeah I think for individual use cases, being able to share one-off documents is probably sufficient. We'll need to give some thought to what the receiver experiences and how they keep track of things that are shared with them outside of vaults. 

Google drive just does the "shared with me" sort of region, the thing that helps you organize that is the context around who shared it with you and when. That's probably easy for us to do.

The "add to my drive" sort of experience is probably what we want to give more thought to. I think, like you're suggesting this will probably be our version of symlinks. A symlink is a doc which points to a UUID of a file. It's just a navigation thing, probably. This is perhaps a new `file_type`, how exciting.

docs/design-ux/sharing.md
```diff
@@ -0,0 +1,120 @@
...
+the whole team. Enterprises are large organizations that want to control how things are shared in sophisticated ways, such
+as having teams with subteams that have at least as much access. The designs we consider are:
+
+1. Root sharing
+2. Folder sharing
+3. Policy Sharing
+
+## 1. Root Sharing
+
+### UX
```
tvanderstad, line 24: Proposition for one-off sharing UX: make a new vault for the group of people you want to share with, move the document to that vault, and replace it with a symlink to that document. Opening a symlink in the UI is the same as opening the document itself, so laypeople can't tell the difference. Symlinks are by UUID, are immutable, and can't point to other symlinks. Maybe shared vaults can't contain symlinks as an additional simplification?

Vaults like this would just sort of accumulate over time, like group chats. Instead of making formal teams like 1password or github, the vault could just be defined in terms of all the sharees, like how if you send a group message to the same 3 people as a previous message, they show up in the same thread. You can manage group members, name the group, etc like in messengers

Parth:
Another proposition for one-off sharing use cases, that would also result in a simplification of our encryption model.

# Teams, folders and vaults

+ Vaults (aka roots) can be shared with people.
+ Vaults are globally unique along with usernames.
+ Once you're added to a vault you can navigate to it from our omnibox.
+ The vaults you have access to will show up in settings, and you can pin them on a per-device basis.

Once a vault is shared with you, you can navigate to it via URL (as in someone slacks you the URL) or check settings. In the future there can be contextual notifications around this, either in the app or to your home screen whatever.

Additionally the first time you navigate to a vault we can consider asking you at that point if you want to pin this vault on this device.

# One off sharing

+ Documents from a vault can be shared with people outside that vault.
+ This will also show up in that same location in settings.
+ Additionally links will work, and will navigate clients to this document.
+ When you're viewing any document you can always make a "shortcut" to it.
+ Shortcuts are just instructions to your client to navigate to a location. They can exist anywhere and lead to any document.
+ If you can decrypt the file, you can see it, if you cannot decrypt the file you won't be able to see it. 

# Access Control

+ Billing can happen per vault, we can choose to have no free tiers for vaults
+ Maybe a vault can just bill it's members for what they write
+ When you share a document maybe you can do it read-only as well as write.

docs/design-tech/sharing.md
```diff
@@ -0,0 +1,117 @@
...
+
+Core will expose four new functions:
+1. `share`
+2. `get_pending_shares`
+3. `set_link`
+4. `delete_pending_share`
+
+### Share
+`share` is the function used to share a file. It accepts a file id, a username,
+and a share mode and it returns only a success or error. It fetches the public
```
steverusso, line 57:
```suggestion
and a share mode, and it returns only a success or error. It fetches the public
```
or
```suggestion
and a share mode and returns only a success or error. It fetches the public
```

docs/design-tech/sharing.md
```diff
@@ -0,0 +1,117 @@
...
+The expected errors are:
+* `NoAccount`
+* `FileNonexistent`
+* `ClientUpdateRequired`
+* `CouldNotReachServer`
+* `UserNonexistent`
+* `FileAlreadySharedWithThatUser`
+
+### Get Pending Shares
+A pending share is a file which is shared with a user but the user doesn't have
```
steverusso, line 71:
```suggestion
A pending share is a file which is shared with a user, but the user doesn't have
```

docs/design-ux/sharing.md
```diff
@@ -0,0 +1,41 @@
...
+* readonly vs read/write access
+* unambiguous relative paths (e.g. cli, markdown rendering)
+* who is billed for shared documents?
+* should users be able to reshare shared things?
+
+## Sharing A File
+To share a file, a user selects the "share" option from a context menu. They are
+prompted to enter a username for a user to share the file with and a share mode
+(read or write). The username is checked immediately - **the client must be
+online to share**. The sharer is always considered the owner of the file is and
```
steverusso, line 18:
```suggestion
online to share**. The sharer is always considered the owner of the file and
```

steverusso (commented):
A few things I caught while reading this over

PR #833, opened 2021-07-28
docs: file-types rfc

Parth:
This RFC tries to figure out how we deal with files we're not going to engineer explicit support for.

### DIFF
--- a/docs/design-ux/file-types.md
+++ b/docs/design-ux/file-types.md
@@ -10,45 +10,15 @@ These are the set of file types (for now) that we will author UI elements for.
 
 # Second party support
 
-For arbitrary file types we have a few options.
+For arbitrary files within clients we'll a configurable process with a temporary file, and watch that file for changes.
 
-## [FUSE](https://en.wikipedia.org/wiki/Filesystem_in_Userspace)
+We'll choose some defaults on a per-platform and binary-availability basis. We can explore using tools like `open`,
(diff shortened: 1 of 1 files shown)


docs/design-ux/file-types.md
```diff
@@ -0,0 +1,54 @@
...
++ Drawings
++ Todo lists
+
+These are the set of file types (for now) that we will author UI elements for.
+
+# Second party support
+
+For arbitrary file types we have a few options.
+
+## [FUSE](https://en.wikipedia.org/wiki/Filesystem_in_Userspace)
```
Parth, line 15: FUSE on linux may be a non-issue because it comes pre-installed / compiled into the compiler.

steverusso:
I think the first iteration of unsupported files should be the "explicit file sync" option, but only on a single file basis when a user double clicks (or "opens") one unsupported file at a time. No option to do this on a folder or multi-file level (at least for now).

In my mind, this gets Lockbook to the launch stage, and if there is a better way implemented in the future, it would still look very similar to this from the user's perspeciive: I double click a file, edit with the default application for that type on my system, and Lockbook secures my changes.

I also think a FUSE tool should eventually be a standalone tool in the Lockbook suite, perhaps initially for slightly more technical users. However, I think going this route for unsupported files for all of the clients would be a relatively heavy undertaking right now (compared to the above).

PR #834, opened 2021-08-01
linux: links

SmailBarkouch:
In this PR, you can now have links in your lockbook documents. This means you can link to documents and folders on lockbook, as well as arbitrary files or website links. 

To add a link to a lockbook file you simply follow the format of a regular markdown link, and inside the parenthesis you write out the absolute path of the lockbook file, prefixing it with "lb://". You will know the format is valid once the markdown link is bolded red. These are some valid examples:
```
[here](lb://options/info.md)
[there](lb://school/assignment-due-today.md)
```

To link to websites or local files, make sure to include the protocol. Some valid examples:
```
[google](https://google.com)
[lockbook](https://github.com/lockbook/lockbook)
```


clients/linux/src/account.rs
```diff
@@ -361,6 +366,49 @@ impl Editor {
         let gspell_view = gspell::TextView::get_from_gtk_text_view(textview).unwrap();
         gspell_view.basic_setup();
 
+        textarea.connect_button_press_event(
+            closure!(m => move
+                |w, evt| {
```
steverusso, line 371: ?

clients/linux/src/account.rs
```diff
@@ -361,6 +366,49 @@ impl Editor {
...
+                        }
+                    }
+                }
+
+            }
+
+            Inhibit(false)
+        }
+        )
+        );
```
steverusso, line 410: ?

steverusso (changes requested):

clients/linux/src/account.rs
```diff
@@ -369,16 +413,43 @@ impl Editor {
...
+        let language = format!("{}/custom.lang", language_specs);
+
+        if !Path::new(&language_specs).exists() {
+            std::fs::create_dir(&language_specs).map_err(LbError::fmt_program_err)?;
+            std::fs::write(language, CUSTOM_LANG).map_err(LbError::fmt_program_err)?;
+        }
+
+        Ok(language_specs)
+    }
+
```
steverusso, line 452: I don't like seeing file i/o code mixed in with gui code unless it's absolutely necessary. Is this something that can be done when the app is initialized on startup?

clients/linux/src/account.rs
```diff
@@ -11,22 +11,26 @@ use gtk::{
...
 use crate::backend::{LbCore, LbSyncMsg};
-use crate::closure;
 use crate::editmode::EditMode;
-use crate::error::LbResult;
+use crate::error::{LbErrTarget, LbError, LbResult};
 use crate::filetree::FileTree;
 use crate::messages::{Messenger, Msg, MsgFn};
 use crate::settings::Settings;
+use crate::util::gui::LEFT_CLICK;
 use crate::util::{gui as gui_util, gui::RIGHT_CLICK};
```
steverusso, line 27: Please give the imports a once over in the future. Right here, LEFT_CLICK can just be appended to the util import right below it.

steverusso (changes requested):

Parth (changes requested):
This is great, long awaited, works pretty well.

The links do need to start with roots tho, and we should probably change the behavior of that top bar. I think it's a direction @steverusso is not going to be a fan of, but the motivation is:

One of the constraints of sharing is that every document must have an absolute / universal path. Sharing will likely work similarly to github organizations / slack workspaces / 1password vaults.

So if you're chatting with a team, in slack, or github, you'll be able to share absolute paths to documents in your shared vaults.

These documents will start with the vaultname. In the future you'll have multiple root folders representing these vaults. An individual may have different vaults with different security profiles.

Details are all up in the air as we iterate on the RFC, but I'm pretty certain I want every document to have an unambiguous and absolute path, so let's head in that direction.

Next steps along the link journey will probably include:

1. Inspired by convo with @steverusso, let's take a look at how we can have the lockbook application respond to system uri requests that start with `lb://`. The individual would be able to create links / shortcuts / keybindings to specific documents within lockbook. That'd be sick.
2. Relative links, I'm probably okay with using a different schema like `lbrel://` for this.
3. Id links, same deal `lbid://`
4. image links -- probably this is a step along the way of markdown rendering within lockbook

clients/linux/src/app.rs
```diff
@@ -357,6 +360,23 @@ impl LbApp {
         Ok(())
     }
 
+    fn markdown_lb_link_exec(&self, scheme: String, uri: String) -> LbResult<()> {
```
Parth, line 363: Everything should take `&str`

clients/linux/src/account.rs
```diff
@@ -361,6 +365,45 @@ impl Editor {
...
+
+                    let svb = w.get_buffer().unwrap().downcast::<GtkSourceViewBuffer>().unwrap();
+                    let maybe_selected = svb.get_text(&start, &end, false);
+                    let index = iter.get_line_index();
+
+                    if let Some(text) = maybe_selected {
+                        let uri_regex = Regex::new(r"\[.*]\(([a-zA-z]+://)(.*)\)").unwrap();
+
+                        for capture in uri_regex.captures_iter(text.as_str()) {
+                            let whole = capture.get(0).unwrap();
```
Parth, line 388: can we just exit this closure or something instead of `unwrap` ing.

Parth: I understand that it's never going to fail because of the for loop

Parth (changes requested):
In the URL bar the root is still missing.

Parth (approved):

PR #851, opened 2021-08-13
cli: tmp is cleaned up even if an error occurs 

SmailBarkouch:


Parth (changes requested):
I'd like to cleanup 2 things, some out of the context of this PR:
+ have this guy return a value and let the caller determine whether it gets printed or not.
+ don't have this function say cleaning up temp file when its not guaranteeing that happened

Parth (approved):

Parth:
The description here is wrong I think, it's not when an error occurs that this bug happens. It happens [when it suceeds](https://github.com/lockbook/lockbook/blob/bfc70c3e6e8b3f433c7f00a87e11ff1d954adcdf/clients/cli/src/utils.rs#L261) which perhaps makes it a more serious problem.

So the description should be something like -- don't prematurely exit the program while saving files.

PR #856, opened 2021-08-18
core: better import collision error

### LINKED ISSUE #849: Have a better error for import collisions

SmailBarkouch:
fixes #849

### DIFF
--- a/clients/linux/src/backend.rs
+++ b/clients/linux/src/backend.rs
@@ -241,7 +241,7 @@ impl LbCore {
         import_progress: Option<Box<dyn Fn(ImportExportFileInfo)>>,
     ) -> LbResult<()> {
         import_file(&self.config, parent, PathBuf::from(source), false, import_progress).map_err(map_core_err!(ImportFileError,
-            FileAlreadyExists(path) => uerr_dialog!("File collision at {}.", path),
+            FileAlreadyExists(path) => uerr_dialog!("File already exists at {}.", path),
             DocumentTreatedAsFolder => uerr_dialog!("Invalid import destination, document treated as folder."),
             ParentDoesNotExist => uerr_dialog!("Invalid import destination, parent not found."),
             NoAccount => uerr_dialog!("No account."),
--- a/clients/linux/src/backend.rs
+++ b/clients/linux/src/backend.rs
@@ -30,20 +30,20 @@ macro_rules! match_core_err {
     (
         $err:expr,
         $enum:ident,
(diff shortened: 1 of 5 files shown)


clients/linux/src/backend.rs
```diff
@@ -30,20 +30,20 @@ macro_rules! match_core_err {
     (
         $err:expr,
         $enum:ident,
-        $( $variants:ident => $matches:expr ),+,
+        $( $variants:ident $( ( $val:ident ) )? => $matches:expr ),*,
```
steverusso, line 33: Why the change from `+` to `*`?

SmailBarkouch: I was experimenting and forgot to change it back. My bad.

steverusso: Ah ok :+1:

core/src/service/import_export_service.rs
```diff
@@ -76,7 +80,13 @@ fn import_file_recursively(
             fs::read_dir(disk_path).map_err(CoreError::from)?.collect();
 
         if children.is_empty() {
-            path_service::create_at_path(config, &lockbook_path_with_new)?;
+            path_service::create_at_path(config, &lockbook_path_with_new).map_err(|err| {
+                if err == CoreError::PathTaken {
```
steverusso, line 84: nit: I think a match would be cleaner here.

core/src/service/import_export_service.rs
```diff
@@ -80,13 +80,13 @@ fn import_file_recursively(
             fs::read_dir(disk_path).map_err(CoreError::from)?.collect();
 
         if children.is_empty() {
-            path_service::create_at_path(config, &lockbook_path_with_new).map_err(|err| {
-                if err == CoreError::PathTaken {
-                    CoreError::ImportCollision(lockbook_path_with_new)
-                } else {
-                    err
+            match path_service::create_at_path(config, &lockbook_path_with_new) {
```
steverusso, line 83: I don't understand the change here, why not just match within the map_err?

steverusso: The `map_err` that was there previously: https://github.com/lockbook/lockbook/pull/856#discussion_r693281304

SmailBarkouch: Oh, I thought you meant match the entire thing instead of map_err. I was confused too, but now it makes sense haha.

Parth (approved):

PR #857, opened 2021-08-20
linux: paste and drop images and uris into gtksourceview and have them create links

SmailBarkouch:
In this PR I added the ability to paste and drop images and files into gtksourceview. 

When an image is pasted or dropped it will automatically be imported and a link to it will be added in the document. The same goes for uris.

### DIFF
--- a/clients/linux/src/account.rs
+++ b/clients/linux/src/account.rs
@@ -8,19 +8,26 @@ use glib::SignalHandlerId;
 use gspell::TextViewExt as GtkTextViewExt;
 use gtk::prelude::*;
 use gtk::Orientation::{Horizontal, Vertical};
-use gtk::{Adjustment as GtkAdjustment, Align as GtkAlign, Box as GtkBox, Button as GtkBtn, Clipboard, Entry as GtkEntry, EntryCompletion as GtkEntryCompletion, EntryIconPosition as GtkEntryIconPosition, Grid as GtkGrid, Image as GtkImage, Label as GtkLabel, Menu as GtkMenu, MenuItem as GtkMenuItem, Paned as GtkPaned, ProgressBar as GtkProgressBar, ScrolledWindow as GtkScrolledWindow, Separator as GtkSeparator, Spinner as GtkSpinner, Stack as GtkStack, TextWindowType, WrapMode as GtkWrapMode, TextIter};
+use gtk::{
+    Adjustment as GtkAdjustment, Align as GtkAlign, Box as GtkBox, Button as GtkBtn, Clipboard,
+    Entry as GtkEntry, EntryCompletion as GtkEntryCompletion,
+    EntryIconPosition as GtkEntryIconPosition, Grid as GtkGrid, Image as GtkImage,
+    Label as GtkLabel, Menu as GtkMenu, MenuItem as GtkMenuItem, Paned as GtkPaned,
+    ProgressBar as GtkProgressBar, ScrolledWindow as GtkScrolledWindow, Separator as GtkSeparator,
+    Spinner as GtkSpinner, Stack as GtkStack, TextMark, TextWindowType, WrapMode as GtkWrapMode,
+};
 use sourceview::prelude::*;
 use sourceview::{Buffer as GtkSourceViewBuffer, LanguageManager};
 use sourceview::{View as GtkSourceView, View};
 
 use crate::backend::{LbCore, LbSyncMsg};
 use crate::editmode::EditMode;
-use crate::error::{LbError, LbResult, LbErrTarget};
+use crate::error::{LbErrTarget, LbError, LbResult};
 use crate::filetree::FileTree;
 use crate::messages::{Messenger, Msg, MsgFn};
 use crate::settings::Settings;
 use crate::util::{gui as gui_util, gui::LEFT_CLICK, gui::RIGHT_CLICK};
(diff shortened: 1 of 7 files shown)


clients/linux/src/account.rs
```diff
@@ -361,11 +364,127 @@ impl Editor {
...
+            change_sig_id: RefCell::new(None),
+            cntr,
+            messenger: m.clone(),
+        }
+    }
+
+    fn on_paste_clipboard(
+        m: &Messenger,
+        c: &Arc<LbCore>,
+        lbs: &Rc<RefCell<LbState>>,
```
steverusso, line 404: I tried to organize the code base so that `LbState` would just be manipulated within `LbApp`, and I think so far that has been maintained. This scenario isn't a good reason to change that imo. Instead, I'd adhere to that principle and pass a Message back to LbApp.

steverusso (dismissed):

clients/linux/src/account.rs
```diff
@@ -2,32 +2,36 @@ use std::cell::RefCell;
...
-    ProgressBar as GtkProgressBar, ScrolledWindow as GtkScrolledWindow, Separator as GtkSeparator,
-    Spinner as GtkSpinner, Stack as GtkStack, TextWindowType, WrapMode as GtkWrapMode,
+    ProgressBar as GtkProgressBar, ScrolledWindow as GtkScrolledWindow, SelectionData,
+    Separator as GtkSeparator, Spinner as GtkSpinner, Stack as GtkStack, TargetList, TextMark,
+    TextWindowType, WrapMode as GtkWrapMode,
 };
 use sourceview::prelude::*;
-use sourceview::View as GtkSourceView;
 use sourceview::{Buffer as GtkSourceViewBuffer, LanguageManager};
+use sourceview::{View as GtkSourceView, View};
```
steverusso, line 22: ?

clients/linux/src/app.rs
```diff
@@ -43,11 +43,13 @@ use crate::settings::Settings;
 use crate::util;
 use crate::{closure, progerr, tree_iter_value, uerr, uerr_dialog};
 
+use gdk::{Cursor, WindowExt};
 use glib::uri_unescape_string;
 use lockbook_core::model::client_conversion::ClientFileMetadata;
 use lockbook_core::service::import_export_service::ImportExportFileInfo;
 use std::path::PathBuf;
 
+#[macro_export]
```
steverusso, line 52: Does this macro export need to still be here? Or the one for `spawn` below?

clients/linux/src/app.rs
```diff
@@ -1079,6 +1121,79 @@ impl LbApp {
...
+            .ok_or_else(|| uerr_dialog!("Open a file before pasting!"))?;
+
+        match info {
+            TextAreaDropPasteInfo::Image(bytes) => {
+                let parent_path = self.core.full_path_for(&opened_file.parent)?;
+
+                let gdk_win = self.gui.account.cntr.get_window().unwrap();
+                gdk_win.set_cursor(Cursor::from_name(&gdk_win.get_display(), "wait").as_ref());
+
+                let image_name = format!("img-{}.{}", Uuid::new_v4(), "jpeg");
```
Parth, line 1141: How are you able to assume jpeg?

clients/linux/src/app.rs
SmailBarkouch: It seems it doesn't, will remove it.

clients/linux/src/app.rs
Parth: Apart from ☝️ I'm good with this when @steverusso is good with this.

In the future we'll probably want a progress indicator, but in release mode for huge files (100MB RAW) this is already good enough for now, IMO.

SmailBarkouch: It isn't so much as assuming as it is that pixbuf doesn't expose a way to view the file format of the image it contains. If something is of type pixbuf (like what I get from the clipboard or drop), then it is able to be changed into any of the wide array of file formats it supports. Probably adding a dialog to choose would be good. Another way I could go about this is to get the raw bytes from the clipboard and drop, which should in turn preserve its type, but I would likely need to import a library or something to find out the file format to have the correct extension for the file created from it.

clients/linux/src/account.rs
```diff
@@ -361,11 +403,107 @@ impl Editor {
...
+                    Some(pixbuf) => match pixbuf.save_to_bufferv("jpg", &[]) {
+                        Ok(bytes) => TextAreaDropPasteInfo::Image(bytes),
+                        Err(err) => {
+                            m.send_err_dialog("Dragging bytes", LbError::fmt_program_err(err));
+                            return;
+                        }
+                    },
+                },
+                TEXT_TARGET_INFO => return,
+                _ => panic!("impossible"),
```
Parth, line 473: idk if this is panic worthy, you could just `eprintln`, pop a dialog open, or return early from the closure

clients/linux/src/app.rs
Parth: https://valadoc.org/gdk-pixbuf-2.0/Gdk.Pixbuf.save_to_bufferv.html

Gotcha, okay can we save as png, if the image has an alpha layer, then jpeg will blast it away

SmailBarkouch: Oh, I didn't know it didn't support alpha. Gotcha, will save it as png.

Parth (approved):

PR #864, opened 2021-09-07
apple: fix drawings

Parth:
This whole bit does need to be reworked, but this fixes a regression introduced in the refactor


Parth:
admin merging this

PR #910, opened 2021-11-08
core: restore get dirty local files

Parth:


tvanderstad (approved):
I like to `vec1.iter().chain(vec2.iter()).collect()` but this is fine too

PR #911, opened 2021-11-09
fuzzer: got rid of unwraps and log nonstop in fuzzy tests

SmailBarkouch:


core/tests/exhaustive_sync/trial.rs
```diff
@@ -89,6 +89,7 @@ pub struct Trial {
     pub steps: Vec<Action>,
     pub completed_steps: usize,
     pub status: Status,
+    pub time_of_start: i64,
```
Parth, line 92:
```suggestion
    pub start_time: i64,
```

core/tests/exhaustive_sync/experiment.rs
```diff
@@ -35,7 +37,8 @@ impl Experiment {
         let mut state = experiments.lock().unwrap();
         let experiment = state.pending.pop();
         match experiment {
-            Some(found) => {
+            Some(mut found) => {
+                found.time_of_start = get_time().0;
```
Parth, line 41: I think this internal modification of a `Trial` should just happen in trial when in `execute()`.

Parth: I think the default start time should be `0`

additionally when a mutation is generated there is a fn that will reset it's status and other properties, you should set the time on the mutation back to 0 as well.

Parth (changes requested):
comments

Parth (approved):

PR #921, opened 2021-12-02
linux: populating file tree is quicker

SmailBarkouch:
Originally in the linux client, the filetree was populated using a chain of `get_children` calls, starting from the root. This was an inefficient process as it would require `core` to keep retrieving all the files using io operations, and matching to get the desired parent. 

In this PR, I call the exposed `lib.rs` function, `list_metadatas`, to get all the metadata. I match on these metadata to construct the file tree myself, so I don't make any more `core` calls.

On my local machine, with my lockbook containing `307` files and `14.54 mb` of compressed usage, these were the time differences between the new algorithm and the old one (the time it takes to load the file tree)

Debug Build:
`58524ms` old algorithm
`6ms` new algorithm

Release Build:
`4707ms` old algorithm
`1ms` new algorithm

### DIFF
--- a/clients/linux/src/backend.rs
+++ b/clients/linux/src/backend.rs
@@ -1,3 +1,4 @@
+use std::collections::BTreeMap;
 use std::env;
 use std::path::{Path, PathBuf};
 use std::sync::RwLock;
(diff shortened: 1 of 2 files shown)


clients/linux/src/backend.rs
```diff
@@ -168,6 +168,10 @@ impl LbCore {
         Ok(get_children(&self.config, parent.id)?)
     }
 
+    pub fn get_all_files(&self) -> LbResult<Vec<DecryptedFileMetadata>> {
+        Ok(list_metadatas(&self.config)?)
```
Parth, line 172: This is not the way you want to do this, there already is a get all metas just think about how android is doing it!

Parth (dismissed):

clients/linux/src/backend.rs
Parth: whoops I thought this was lib.rs for a second, nvm

clients/linux/src/filetree.rs
```diff
@@ -347,10 +347,52 @@ impl FileTree {
         self.tree.get_selection().get_selected_rows()
     }
 
+    pub fn vec_into_tree(
+        &self,
+        metadatas: &[DecryptedFileMetadata],
+        parent_metadata: &DecryptedFileMetadata,
+        parent_iter: &TreeIter,
+    ) -> LbResult<()> {
```
steverusso, line 355: This function will always return `Ok(())`, so I don't see the need for a return type. Am I missing something?

SmailBarkouch: You're right!

steverusso (approved):

PR #938, opened 2021-12-23
docs: sync v3

tvanderstad:


docs/design-tech/sync.md
```diff
@@ -1,89 +1,118 @@
...
-Design constraints:
+This data model lends itself to three operations, which correspond to modifications to the three mutable fields of a file:
+* `rename`: change the file's name
+* `move`: change the file's parent
+* `delete`: delete the file (cannot be undone)
 
-+ Secure / Private / Trustless -- cannot depend on server to resolve conflicts
-+ Frictionless -- cannot depend on user intervention to resolve conflicts
-+ Network Independent -- should allow a user to complete as many operations offline as possible
+With that data model in mind, these are the three main invariants that must be true of a user's file tree on all devices at all times:
```
tvanderstad, line 28:
```suggestion
With that data model in mind, these are the four main invariants that must be true of a user's file tree on all devices at all times:
```

docs/design-tech/sync.md
```diff
@@ -1,89 +1,118 @@
...
+* Root Invariant: there must always be exactly one `root` and all modifications to `root`s are invalid. This is straightforward to enforce.
+* Path Conflict Invariant: no two files may have the same name and parent. An example challenge is when a user creates a new file on each of two clients with the same name and parent, then syncs them both.
+* Cycle Invariant: every non-root file must not have itself as an ancestor.  An example challenge is when on one client the user moves folder A into folder B and on the other client moves folder B into folder A, then syncs them both.
```
tvanderstad, line 31:
```suggestion
* Cycle Invariant: every non-root file must not have itself as an ancestor. An example challenge is when on one client the user moves folder A into folder B and on the other client moves folder B into folder A, then syncs them both.
```

docs/design-tech/sync.md
```diff
@@ -1,89 +1,118 @@
...
 
-As discussed in the [data model](data_model.md), every time a file's property changes, the server increments its
-`metadata_version`. Ordering the `file_metadata` by this property is our primary means of filtering out what events a
-client knows about and what new events it has to sync.
+This encryption chain, and its interaction with the invariants, creates programming challenges. Our general pattern is to decrypt keys, names, and files as we read them from disk or as responses from the server so that we can write logic without worrying about encryption. However, most invariant violations (e.g. orphans and cycles) will cause us to fail to be able to decrypt files, which in turn makes it difficult to detect and appropriately resolve invariant violations.
```
tvanderstad, line 39:
```suggestion
This encryption chain, and its interaction with the invariants, creates programming challenges. Our general pattern is to decrypt keys, names, and files as we read them from disk or receive them as responses from the server so that we can write logic without worrying about encryption. However, most invariant violations (e.g. orphans and cycles) will cause us to fail to be able to decrypt files, which in turn makes it difficult to detect and appropriately resolve invariant violations.
```

docs/design-tech/sync.md
```diff
@@ -1,89 +1,118 @@
...
+* implicitly deleted (the file has `false` value for `deleted` but has an explicitly deleted ancestor)
+* pruned (there is no mention of file on disk, although it once existed)
 
-Pushing local changes from core is much more trivial. Core will ask `LocalChangesRepo` what changed and perform the
-corresponding network action. If a file was renamed, for example, it will send a `/rename` request to the server.
+First we'll justify the distinction between deleted and pruned. If the file deletion system is working properly, then after a user deletes a file on a device and syncs that device, that device will prune the file and any other devices that sync will prune that file. What's required to make that happen is that the client which deletes the file remembers that the file exists (and is deleted) until that client syncs its changes. Once the changes are synced, the client can prune the file knowing that the server will remember the deletion forever and will inform the other clients of the deletion. Note that deleted documents can have their contents pruned immediately - it is only important to remember the metadata until the deletion is pushed.
```
tvanderstad, line 76: todo: we can delete file contents for implicitly deleted files, but we might have to re-pull them during a sync

docs/design-tech/sync.md
```diff
@@ -1,89 +1,118 @@
...
+
+### Invariant Violation Resolution
+Even if clients individually enfoce invariants, violations can still occur in the presence of concurrent updates. This is a complicated matter.
+
+First, we have to be concerned about the Cycle Invariant. A cycle can occur if a user has two folders, moves the first folder into the second on one client and moves the second folder into the first on another, then syncs them both. A cycle can also involve more folders, if a first folder is moved into a subfolder of a third folder and the third folder is moved into the first folder concurrently.
+
+We also have to be concerned about the Path Conflict Invariant. A path conflict can occur if two clients create files with the same name in the same folder and then sync. Occurances can also involve a client renaming an existing file that folder, moving a file into a folder, or after a file edit conflict is resolved by saving a copy of the local version with a new name. Additionally, it must be valid for a two files to share a name and parent if one or both of them are deleted.
```
tvanderstad, line 89:
```suggestion
We also have to be concerned about the Path Conflict Invariant. A path conflict can occur if two clients create files with the same name in the same folder and then sync. Occurrences can also involve a client renaming an existing file in that folder, moving a file into a folder, or after a file edit conflict is resolved by saving a copy of the local version with a new name. Additionally, it must be valid for a two files to share a name and parent if one or both of them are deleted.
```

docs/design-tech/sync.md
```diff
@@ -1,89 +1,118 @@
...
+
+There are enough corner cases here to drown in. Rather than employ a team of talented engineers to discover and diagnose them, we aim to implement a robust system which repairs the violations in a user-friendly way as part of the sync routine.
+
+To resolve cycles, we rely on the fact that `base`, `local`, and `remote` all individually do not contain any cycles and design the resolution to preserve that. If the result of merging the file trees has a cycle, it must involve an unsynced local move and a remove the client just pulled from the server. In particular, one of the files in the cycle must have been the subject of an unsynced local move. Our policy to resolve this is to unmove (by referring to `base`) all the locally moved files that are involved in any cycles that result after a file tree merge.
```
tvanderstad, line 95:
```suggestion
To resolve cycles, we rely on the fact that `base`, `local`, and `remote` all individually do not contain any cycles and design the resolution to preserve that. If the result of merging the file trees has a cycle, it must involve an unsynced local move and a remote move the client just pulled from the server. In particular, one of the files in the cycle must have been the subject of an unsynced local move. Our policy to resolve this is to unmove (by referring to `base`) all the locally moved files that are involved in any cycles that result after a file tree merge.
```

docs/design-tech/sync.md
```diff
@@ -1,89 +1,263 @@
...
+
+For `delete`, the client validates that:
+- the file is not the root folder
+
+_Concerns: `create` unnecessarily checks for invalid cycles_
+
+### Local Content Edits
+When a user modifies a document's contents, the client checks that the file exists and is not deleted, then updates the `local` version of the file's contents.
+
+### Upsert File Metadata
```
tvanderstad, line 214: todo: clarify design in terms of two dbs and crash recovery

docs/design-tech/sync.md
```diff
@@ -1,89 +1,263 @@
...
...g the new content version, the new metadata is never saved with the new content version, so clients never request it and the contents are effectively not updated. If the server is interrupted after the new metadata is saved, clients receieve the new metadata with the new content version and the content is effectively updated, but the old version of the file is never cleaned up from object storage. If the update is a deletion, the first step is skipped, but the rest works the same way. The server saves the metadata, then deletes the old version of the content. If the server is interrupted after only saving the metadata, clients will receive the deletion and prune the file when they pull updates, but like with other edits, the old version of the file is never cleaned up from object storage. So, **under this design, the system operates safely but will sometimes leave unnecessary objects in storage.** The last piece of the server recovery design is to figure out how to remove these old content versions from object storage in the event of an interruption. Note that this process is not urgent; it only affects infrastructure costs to the maintainers to have unnecessary objects in storage.
```
tvanderstad, line 168: todo: is this not bold?

docs/design-tech/sync.md
```diff
@@ -1,89 +1,263 @@
...
+## Overview
+Sync v3 is the latest iteration of our logic that syncs users' file trees across devices. It is designed under the following key constraints:
+* The state of a user's file tree must satisfy certain invariants at all times, on all devices (including the server), even in the event of process interruptions and failures.
+* Users' file contents and names must not be readable by the server, but users can share files and folders with other users.
+* Users must be able to view and edit files offline, then sync frictionlessly, even with intervening edits made to files on other devices.
+* The performance of the system must not suffer as a result of typical usage, which presents a number of challenges worth discussion
 
-Understanding the [data model](data_model.md) is a prerequisite to understanding this document.
+These design constraints motivate the following high-level design, which will be flushed out in more detail after a further discussion of the constraints:
+* The client maintains two versions of the file tree: one for the current local version (`local`), and one for the last version known to be agreed upon by the client and server (`base`).
```
Parth, line 10: A potential useful disambiguation: `local` is empty when there are no changes locally. We don't just take up 2x the storage locally.

Resolve if you disagree.

docs/design-tech/sync.md
```diff
@@ -1,89 +1,263 @@
...
 
-Sync is the means by which a user updates a client with either their changes on other devices or peer changes on shared files.
+## Constraints
+### File Tree Invariants
+Before discussing the invariants, it is helpful to understand the data model. For the purposes of Sync v3, each file has:
+* `id`: an immutable unique identifier
+* `file_type`: an immutable flag that indicates if the file is a folder or document. A _file_ is either a _document_, which has contents, or a _folder_, which can have files as children.
+* `name`: the file's name. Clients encrypt and HMAC names before sending them to the server, so while the server cannot read file names, it can test them for equality.
+* `parent`: the file's parent folder (specified by the parent's `id`). The `root` folder is the folder which has itself as a parent (it's name is by convention it's owners username and it cannot be deleted).
+* `deleted`: a boolean flag which indicates whether the file is _explicitly_ deleted. A file whose ancestor is deleted need not be explictly deleted; if not, it is considered _implictly_ deleted. A file that is no longer known to a client at all is considered _pruned_ on that client.
```
Parth, line 21: `need not be explictly deleted` doesn't the server have too delete all children explicitly tho?

docs/design-tech/sync.md
```diff
@@ -1,89 +1,263 @@
...
-+ `DocumentRepo` -- store of every document known locally
-+ `LocalChangesRepo` -- a description of all the changes that have been made locally
-+ `LastSyncedRepo` -- a simple `get/set` value that the sync algorithm uses to help the server filter out events that
-  are already known locally
+This encryption chain, and its interaction with the invariants, creates programming challenges. Our general pattern is to decrypt keys, names, and files as we read them from disk or receive them as responses from the server so that we can write all the remaining logic without worrying about encryption. However, most invariant violations (e.g. orphans and cycles) will cause us to fail to be able to decrypt files, which in turn makes it difficult to detect and appropriately resolve invariant violations.
 
-As discussed in the [data model](data_model.md), every time a file's property changes, the server increments its
-`metadata_version`. Ordering the `file_metadata` by this property is our primary means of filtering out what events a
-client knows about and what new events it has to sync.
+_todo: write and link the sharing design doc._
```
Parth, line 40: You can link to the un-merged doc now, you know what it's URL will be: https://github.com/lockbook/lockbook/blob/master/docs/design-tech/sharing.md

docs/design-tech/sync.md
```diff
@@ -1,89 +1,263 @@
...
 
-## `execute_work`
+Clients store two versions of the file tree, `base` and `local`. In order to save space, we use semantics similar to copy-on-write where the version of a file on `local` is not saved if it is identical to the version on `base`. Clients interpret the absence of a `local` version to mean that `local` and `base` are identical. Additionally, we compress files, both before saving to disk and before pushing to the server.
 
-### `server -> core`
+The metadata for deleted files must be eventually pruned on clients so that the amount of space taken on users' devices scales with the number of not-deleted files rather than the total number of files ever created. Deleted files cannot be immediately pruned because the client needs to remember to push the deletion to the server on the next sync. The server keeps track of all file metadata forever (though it deletes contents of deleted files) so that clients can sync after being offline indefinitely and still receive deletion events without the complexity of the server needing to know which clients exist and which updates they have each received.
```
Parth, line 60: > cannot be immediately pruned because the client needs to remember to push the deletion to the server on the next sync.

however it's content can, correct?

delete is a permanent operation.

Okay so what about in the recursive sense, what if that file was rescued by someone else out of a folder that was deleted by you.

That document will need to be undeleted. Perhaps an area that requires further consideration.

Could be frusterating to delete a file because your phone is out of space, and then the space doesn't free up, and you're confused about why.

Parth: >  Note that deleted documents can have their contents pruned immediately - it is only important to remember the metadata until the deletion is pushed.

okay I see this is covered later, let's see if the folder case is covered

Parth: okay cool, I'll just resolve this one lol

docs/design-tech/sync.md
```diff
@@ -1,89 +1,263 @@
...
+
+I know what you're wondering - does the order of these conflict resolutions matter? Yes it does. Any content conflicts need to be resolved before path conflicts because resolving content conflicts can create new files which can cause path conflicts. Any cycles need to be resolved before path conflicts because resolving cycles can move files which can cause path conflicts. Updates to orphaned files can be ignored at any time because no resolution policy depends on or modifies orphaned files.
+
+_Concern: we don't currently respect "Any cycles need to be resolved before path conflicts because resolving cycles can move files which can cause path conflicts." A failing test case is left as an exercise for the reader._
```
Parth, line 104: > A failing test case is left as an exercise for the reader.
lol that's not how this works

docs/design-tech/sync.md
```diff
@@ -1,89 +1,263 @@
...
+
+If desired, a similar process can be used **to update the editor after a sync: when the user syncs, read the most up-to-date version of the document from disk (including potential changes pulled during a sync), 3-way merge the up-to-date disk version, editor version, and the inital disk version (using the initial disk version as the base), then update the editor to contain the result of the merge.**
+
+#### Client Recovery
+The three key sync-related things that the client stores on disk are file metadata, file contents, and the last synced time. If these are managed properly, then clients are resilient to interruptions. Ultimately, **if updates to data on disk are atomic (e.g. by using an embedded database with transaction support), client recovery is not difficult.** Let's look at some of the problems we would encounter otherwise.
```
Parth, line 145: does this imply an explicit modeling of last synced time?

Parth (approved):
looks good, would be great if you wrapped lines in the source

docs/design-tech/sync.md
travis-madhive: I actually fixed this already anyway; will update

docs/design-tech/sync.md
travis-madhive: yeah in the meantime it is not explicitly deleted

Issue #907, opened 2021-11-07
cleanup core integration tests


Parth:
+ There's a handful of test sets that aggressively copy paste utilities
+ Tests share a module which slows down test run time significantly

PR #942, opened 2022-01-02
core: test cleanup, additional sync tests

tvanderstad:
this PR:
- makes test helpers not macros if they can be not macros. macros are difficult to maintain for a bunch of reasons. they do have the advantage of reporting the line of the test failure concisely but we can use the stack trace for that.
- moves test helpers to test_utils.rs
- standardizes sync tests so that, when applicable, they use the version of functionality exported by core's lib.rs instead of an internal service
- adds test helpers for checking specifically how many local vs server work units there are and what files they correspond to
- adds various integrity check assertions to all sync tests
- makes sync tests more specific rather than running through a long list of operations with assertions interspersed throughout
- adapts or removes tests cases which have been commented out
- adds so many test cases

problems raised + fixed:
- [x] deleting an unsynced file should not result in work for that file
- [x] path conflicts result in server file being renamed instead of local
- [x] cycles involving folders with children result in infinite loops
- [x] cycle resolution also undoes local renames
- [x] deleted files interfere with versions while resolving content conflicts

Closes: https://github.com/lockbook/lockbook/issues/907


Parth:
Currently changing a single test causes all integration tests to recompile. I think this is because all the tests are a member of this module: https://github.com/lockbook/lockbook/blob/master/core/tests/integration_test.rs

For example: https://github.com/lockbook/lockbook/blob/master/core/tests/sync_service_tests.rs#L1

But they just don't need to be: https://github.com/lockbook/lockbook/blob/master/core/tests/exhaustive_sync_check.rs

While you're cleaning all this up, see if you can just remove those lines, and benefit from a much more productive development experience.

Parth (approved):

PR #947, opened 2022-01-05
server: use redis instead of postgres

Parth:
See schema documented [here](https://github.com/lockbook/lockbook/blob/parth/redis-for-real/docs/design-tech/redis-schema.md).

# Implement endpoints

+ [x] `/upsert-file-metadata`
+ [x] `/change-document-content`
+ [x] `/get-document`
+ [x] `/get-public-key`
+ [x] `/get-usage`
+ [x] `/get-updates`
+ [x] `/new-account`
+ [x] `/get-build-info`

# Documentation

+ [x] document redis schema
+ [x] document [redis-utils](github.com/parth/redis-utils)
+ [x] publish [redis-utils](github.com/parth/redis-utils) to crates.io

# Infra work

+ [x] local dev infra work
+ [ ] restore metrics calculated directly through postgres
+ [ ] benchmark and prove it's faster via fuzzer
+ [ ] prod wipe and deploy

# Bonus points

+ [ ] server should `PING` redis on startup
+ [ ] setup LRU document cache


core/libs/models/src/tree.rs
```diff
@@ -0,0 +1,298 @@
+use crate::file_metadata::FileType;
```
tvanderstad, line 1: none of these fn's had unit tests?

core/src/service/import_export_service.rs
```diff
@@ -9,10 +9,11 @@ use lockbook_models::file_metadata::{DecryptedFileMetadata, FileType};
 
 use crate::model::repo::RepoSource;
 use crate::model::state::Config;
-use crate::pure_functions::files;
+
```
tvanderstad, line 12:
```suggestion
```

core/src/service/usage_service.rs
```diff
@@ -1,10 +1,11 @@
 use serde::Serialize;
 
 use lockbook_models::api::{FileUsage, GetUsageRequest, GetUsageResponse};
+use lockbook_models::tree::FileMetaExt;
 
 use crate::model::repo::RepoSource;
 use crate::model::state::Config;
-use crate::pure_functions::files;
+
```
tvanderstad, line 8:
```suggestion
```

core/tests/get_usage_tests.rs
```diff
@@ -105,20 +103,20 @@ mod get_usage_tests {
...
-        let usages = get_usage(&config).unwrap();
-        let mut total_usage = 0;
-        for usage in usages.usages {
-            total_usage += usage.size_bytes;
-        }
+        let usage = get_usage(&config).unwrap_or_else(|err| panic!("{:?}", err));
 
-        assert_eq!(get_usage(&config).unwrap().usages.len(), 3);
-        assert_eq!(total_usage, local_encrypted.len() as u64)
+        println!("{:?}", usage);
```
tvanderstad, line 119:
```suggestion
```

server/server/src/file_service.rs
```diff
@@ -1,213 +1,242 @@
...
+    for change in changes {
+        if let Some((old_parent, _)) = change.old_parent_and_name {
+            if change.id == old_parent {
+                return Err(ClientError(RootImmutable));
             }
-            // prevent turning existing folder into root
-            if upsert.id == upsert.new_parent {
-                return Err(ClientError(FileMetadataUpsertsError::GetUpdatesRequired));
-                // todo: better error
+            if change.id == change.new_parent {
```
tvanderstad, line 70: unclear how this would be resolved by getting updates; seems like this is a case of a folder being moved into itself which shouldn't be allowed even locally

server/server/src/file_service.rs
```diff
@@ -1,213 +1,242 @@
...
+
+        meta.content_version = new_version;
+        meta.metadata_version = new_version;
+
+        pipe.json_set(size(request.id), new_size)?
+            .json_set(file(request.id), meta)
+    });
+    return_if_error!(tx);
+
+    file_content_client::create(
```
tvanderstad, line 205: the metadata update should happen in-between creating the new version of the document and deleting the old version (see https://github.com/lockbook/lockbook/pull/938/files#diff-4a07b1d902803e753277b5b5c01e349a3960b5023e91d000e4a448187bdbccccR168)

tvanderstad (changes requested):

core/libs/models/src/tree.rs
Parth: https://github.com/lockbook/lockbook/blob/master/core/src/pure_functions/files.rs#L448

they just test `apply`

server/server/src/file_service.rs
Parth: I believe this is the same behavior as prior functionality, I think all the get_updates can have better variants for debugability. There are lots of instances of `GetUpdatesRequired` being returned for things that no updating would solve (for example root changes)

server/server/src/file_service.rs
Parth: fixed: 035d9ef

but 

we're gonna wanna do something better for sure, we eat 2 s3 interactions for every doc change, ~50-150ms each.

We can reduce it down to 1 if we process the cleanup asynchronously, throw it on a redis queue and have a background worker churn on them.

We can reduce it down to 0 if we write to redis exclusively, mark the keys as non-volatile until a worker persists them to s3, and then tell redis, if memory is running low you can lru-evict this key.

travis-madhive (approved):

PR #965, opened 2022-01-26
server: feature flags

### LINKED ISSUE #875: implement feature flag system
+ need to be able to toggle off new account sign ups

SmailBarkouch:
In this PR, I exposed a feature flag for creating an account.

The state of all feature flags is stored in redis as a hashmap.  For a new account, lb server simply checks the state of the feature flag, a boolean (on/true or off/false), to determine whether it should accept or deny a request.

The feature flag can be toggled through admin cli, using this command: `lba features new-account [boolean]`.

Since new account requests can now be denied, a new error is exposed. Error needs to be accepted on the following clients:
- [x] Linux (@SmailBarkouch)
- [x] CLI (@SmailBarkouch)
- [x] Android (@SmailBarkouch)
- [x] Apple (@Parth)
- [x] Windows (@tvanderstad)

closes #875


clients/cli/src/error.rs
```diff
@@ -81,6 +81,7 @@ make_errkind_enum!(
     7 => ExpectedStdin,
     8 => NoCliLocation,
     9 => NoRoot,
+    10 => ServerDisabled,
```
Parth, line 84:
```suggestion
    10 => NewAccountsDisabled,
```

clients/cli/src/error.rs
```diff
@@ -132,6 +133,7 @@ impl ErrorKind {
             Self::ExpectedStdin => "expected stdin".to_string(),
             Self::NoCliLocation => "Could not read env var LOCKBOOK_CLI_LOCATION HOME or HOMEPATH, don't know where to place your `.lockbook` folder".to_string(),
             Self::NoRoot => "No root folder, have you synced yet?".to_string(),
+            Self::ServerDisabled => "Server has disabled this feature.".to_string(),
```
Parth, line 136:
```suggestion
            Self:: NewAccountsDisabled => "This server is not accepting new accounts at the moment. Please try again another time.".to_string(),
```

server/admin/src/feature_flags.rs
```diff
@@ -0,0 +1,25 @@
...
+
+pub async fn toggle_new_account_feature_flag(
+    server_state: ServerState,
+    feature_flag: FeatureFlag,
+) -> bool {
+    let mut con = server_state.index_db_pool.get().await.unwrap();
+
+    match feature_flag {
+        FeatureFlag::NewAccount { enable } => {
+            con.hset::<_, _, bool, bool>(
```
Parth, line 14: are you sure these type parameters are needed?

server/server/src/main.rs
```diff
@@ -34,6 +36,19 @@ async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+
+        for flag in feature_flags {
+            if !futures::executor::block_on(con.hexists::<_, _, bool>(FEATURE_FLAGS_KEY, flag))
+                .unwrap()
+            {
+                futures::executor::block_on(con.hset(FEATURE_FLAGS_KEY, flag, true)).unwrap()
+            }
+        }
+    }
+
```
Parth, line 51: Apart from the `block_on` weirdness we just discussed in slack, I view this as some random logic put in our main file. Observe there's an attempt here to move things into fns or have 1-liners.

I do think there will be more things we do at the moment of startup, maybe we want to have a file that has a function called `pre_request_work`, and in there just goes through any startup validation we may want to perform (of which this is the only one).

I think we're going to have similar stuff for stuff our server does in the background (calculate metrics, for instance).

But I think we could also get away from all these stray calls, maybe we don't even need anything in `keys.rs`. Maybe you could (like `content_cache.rs` expose a `feature_flags.rs` which neately encapsulates all this business. I can picture 3 fns:

+ `initialize_flags`
+ `new_accounts_enabled -> bool`
+ `set_new_account_status(bool)`

Yeah I think that's going to solve a lot of the above feedback in a more neat way

clients/linux/src/backend.rs
```diff
@@ -107,6 +107,7 @@ impl LbCore {
             AccountExistsAlready => uerr_dialog!("An account already exists."),
             CouldNotReachServer => uerr_dialog!("Unable to connect to the server."),
             ClientUpdateRequired => uerr_dialog!("Client upgrade required."),
+            ServerDisabled => uerr_dialog!("Server has disabled account creation."),
```
Parth, line 110:
```suggestion
            ServerDisabled => uerr_dialog!("This server is not accepting new accounts at the moment. Please try again another time."),
```

Parth: The error should show up contextually as part of new account error handling

clients/android/app/src/main/res/values/strings.xml
```diff
@@ -212,6 +212,7 @@
     <string name="folder_moved_into_itself">A folder cannot be moved into itself.</string>
     <string name="no_root">No root!</string>
     <string name="cannot_move_root">Cannot move root!</string>
+    <string name="new_account_disabled">This server is not accepting any new accounts at this moment. Please try again another time</string>
```
Parth, line 215:
```suggestion
    <string name="new_account_disabled">This server is not accepting any new accounts at this moment. Please try again another time.</string>
```

Parth (approved):

PR #968, opened 2022-02-03
apple: massively improve markdown editing experience

### LINKED ISSUE #754: Set active style to markdown style on MacOS
steps to reproduce:
1. make a markdown doc in the macos app
2. type `# Anything`
3. wait for markdown render
4. keep typing on the same line

expected behavior:
newly typed text same style as rest of line instantly

actual behavior:
newly typed text is always the same style and you have to wait for markdown render to fix it

### LINKED ISSUE #846: link support in documents
+ [ ] Outside lockbook
+ [ ] Within lockbook

Parth:
# make markdown a B experience on iOS

+ Created a more expressive way to model styles. Previously certain styles would conflict. For instance you could not have a heading which modified the fontsize and then try to put a code block inside it (the code block would become small).
+ Made it so we're the only people who can set attributes. There was a lot of weirdness that happened previously with pasted styles, active cursor styles, etc. All that is ignored and only the parsing engine can tell us how to style our document.
+ I relied more heavily on the default style-expansion-behavior of texteditors. Think about when you're using google docs, and you type within a bolded region. You don't enter plain text, you "inherit" the style of the surrounding characters. By being surgical about when this happens we avoid many style updates which are incredibly expensive.
+ We can now, render documents instantly, regardless of document size, and we generally only experience lag in a handful of select places.

### Notes on performance

+ Our prior bottleneck was document updates, our current bottleneck is parsing the document. This requires some explanation because Down is not the bottleneck, it's taking the info we get from down and determining what styles that translates to. The pipeline we have to follow is something like:
  1. down gives us utf-8 column rows
  2. normalize this to ranges (expensive)
  3. convert this to utf-16 ranges (because that's what `NSStorage` wants
+ It could be good to do this whole process in rust and do something that is taking about 20-30ms per keystroke in less than 5ms.
+ In our pipeline we then determine if we should re-render the whole document, rendering the whole document takes 25ms (for a huge document). I did some tests that tried to answer the time difference between styling 2 spans vs styling `n` spans. And the difference did not seem substantial, diminishing the argument for an incremental parser. But I think if we were to move as much of the pipeline to rust, then we can evaluate what gains there are to be had by modeling the whole problem as an incremental computation problem. This should probably happen after we accumulate some more knowledge about the bottlenecks and API's of the other platforms.
+ Could also be a significant value add to the swift community if we popped this out as it's own package -- a high performance markdown editor. But I don't have the bandwidth for that at the moment. Could also be good to do a rust-based version rather than a down + cmark based version. If someone has the interest and the bandwidth for that, don't hesitate to reach out, I'd be happy to offer support in any way I can.

## What getting to an A- experience requires:

+ Automatic insertion of bullets, numbered lists, tabbing, etc.
+ Slightly faster performance for pathological document sizes.
+ Being able to handle image pastes in some capacity.

## A+

+ requires being able to page novel sized documents somehow

### Progress tracking:

+ [x] Eliminate a bunch of dead code
+ [x] Properly use `NSTextStore`'s `beginEditing()` and `endEditing()` for BIG performance gains
+ [x] Re-write Notepad from the ground up to eliminate more deadcode and strange behaviors
+ [x] Investigate paragraph styles for regions like code block background
+ [x] Fix our use of `utf-8` and `utf-16` to prevent weird characters from showing up at document boundaries
+ [x] Make typesetting more beautiful (maybe separate pr)
    + [x] Bullets properly indented
+ [ ] Make editing feel better
    + [ ] Automatic insertion of bullets on return
    + [x] Be able to click on links

fixes #754 
fixes #846 

handy lil things for future work (ignore):

```swift
print(DebugVisitor().visit(document: (try? Down(markdownString: backingStore.string).toDocument())!))
```

```swift
print("Text start line: \(node.cmarkNode.pointee.start_line), endline: \(node.cmarkNode.pointee.end_line), start_column: \(node.cmarkNode.pointee.start_column), end_column: \(node.cmarkNode.pointee.end_column)")
```

# Reading
+ https://developer.apple.com/library/archive/documentation/StringsTextFonts/Conceptual/TextAndWebiPhoneOS/CustomTextProcessing/CustomTextProcessing.html
+ https://developer.apple.com/videos/play/wwdc2021/10061/
+ https://www.objc.io/issues/5-ios7/getting-to-know-textkit/
+ https://github.com/kyle-n/HighlightedTextEditor
+ https://github.com/lockbook/lockbook/tree/f09eb0756572aae21e1012a0d9697c09d802802e/clients/apple/Notepad
+ https://blog.vinhis.me/2017/04/16/bullet-list-on-ios-with-textkit.html
+ https://christiantietze.de/posts/2017/11/syntax-highlight-nstextstorage-insertion-point-change/

revisit: https://stackoverflow.com/questions/36828331/nstextview-is-too-slow-to-update-while-changing-the-attributes-with-different-ns


Parth:
Early experiments with automatic insertion of bullets: 
https://user-images.githubusercontent.com/821972/154593870-660c9981-8ae8-43ed-adba-f946aed04574.mov

PR #972, opened 2022-02-04
server: metrics based on redis and prometheus

SmailBarkouch:
In this PR, I added metrics to server. We already had metrics before, but since our old db was PostgreSQL, the metrics did not work anymore. Now that we utilize redis, we have to have metrics based on it. 

In this situation, we use prometheus to expose various types of information through the metrics endpoint. This is the information:
- total users
- active users
- total documents
- total bytes


server/server/src/metrics.rs
```diff
@@ -0,0 +1,221 @@
...
+
+// pub struct MetricsPreProcessingInfo<'a> {
+//     pub public_keys: Vec<PublicKey>,
+//     pub ids: HashMap<&'a PublicKey, Uuid>,
+//     pub metadatas: HashMap<&'a PublicKey, EncryptedFileMetadata>
+// }
+
+pub const TWO_DAYS_IN_MILLIS: u128 = 172800000;
+
+pub async fn init(server_state: &Arc<ServerState>) {
```
Parth, line 60: + This doesn't actually have to be `async`.

+ `init`, is a missed opportunity here imo, you should call this `start_metrics_worker`.

server/server/src/metrics.rs
```diff
@@ -0,0 +1,221 @@
...
+#[derive(Debug)]
+pub enum MetricsError {}
+
+// pub struct MetricsPreProcessingInfo<'a> {
+//     pub public_keys: Vec<PublicKey>,
+//     pub ids: HashMap<&'a PublicKey, Uuid>,
+//     pub metadatas: HashMap<&'a PublicKey, EncryptedFileMetadata>
+// }
+
+pub const TWO_DAYS_IN_MILLIS: u128 = 172800000;
```
Parth, line 58: instead of `172800000` you should write out `1000 * 60 * 60 * 24 * 2`, I don't have to wonder then if 172800000 is the right number.

containers/local.env
```diff
@@ -8,6 +8,8 @@ FILES_DB_BUCKET=testbucket
 FILES_DB_ACCESS_KEY=minioadmin
 FILES_DB_SECRET_KEY=minioadmin
 
+DURATION_BETWEEN_METRICS_REFRESH=5000
```
Parth, line 11: Instead of DURATION, you should specify the unit of 5000.

Parth: I don't think `MILLIS` is the right unit to use, I think this should be measured in terms of minutes, and I think 5 minutes is probably the right value for us.

server/server/src/config.rs
```diff
@@ -75,11 +76,29 @@ impl ServerConfig {
...
 
+#[derive(Clone)]
+pub struct MetricsConfig {
+    pub duration_between_metrics_refresh: Duration,
+}
+
+impl MetricsConfig {
+    pub fn from_env_vars() -> MetricsConfig {
+        MetricsConfig {
+            duration_between_metrics_refresh: Duration::from_millis(
```
Parth, line 87: from minutes, like I mentioned before

server/server/src/metrics.rs
```diff
@@ -0,0 +1,221 @@
...
+    let state_clone = server_state.clone();
+
+    tokio::spawn(async move {
+        if let Err(e) = start(state_clone).await {
+            error!("Metrics error: {:?}", e)
+        }
+    });
+}
+
+pub async fn start(server_state: Arc<ServerState>) -> Result<(), ServerError<MetricsError>> {
```
Parth, line 70: good clear fn

server/server/src/metrics.rs
```diff
@@ -0,0 +1,221 @@
...
+        let all_file_ids = get_owned_files_id(&server_state, &public_keys).await?;
+        METRICS_STATISTICS.total_documents.set(
+            all_file_ids
+                .iter()
+                .map(|users_files| users_files.len())
+                .sum::<usize>() as i64,
+        );
+
+        let metadatas = get_file_metadata(&server_state, &all_file_ids).await?;
+        let total_bytes = get_total_size(&server_state, &metadatas).await?;
```
Parth, line 84: At this moment you will have used a massive amount of memory, you'll have all metas for all users in ram.

If our server is at a healthy 10/16gb at this moment, then this will cause our server to [OOM](https://en.wikipedia.org/wiki/Out_of_memory) and crash.

You should re-write this with the following constraints:
+ Don't keep more than the following in memory at once:
    + All public keys
    + All needed memory for at most 1 user

+ Additionally you should sprinkle some sleeps in between operations like this:

```
for key in public_keys_keys {
        let public_key = con.maybe_json_get(&key).await?.ok_or_else(|| {
            internal!(
                "Cannot retrieve public_key despite having a valid key: {:?}",
                key
            )
        })?;

        public_keys.push(public_key)
```

If we tried to compute metrics as quickly as possible we'd put considerable load on redis (think about `MONITOR` and all the calls just being metric calls for like a few seconds.

Thankfully we don't need this to be fast, slower is better.

Parth (changes requested):
good first pass

server/server/src/config.rs
```diff
@@ -75,11 +76,48 @@ impl ServerConfig {
     }
 }
 
+#[derive(Clone)]
+pub struct MetricsConfig {
+    pub minutes_between_metrics_refresh: Duration,
+    pub millis_between_user_metrics: Duration,
+    pub millis_between_getting_pub_key_metrics: Duration,
+    pub millis_between_getting_pub_key_key_metrics: Duration,
```
Parth, line 84: You don't need to specify the type here, `Duration` will do all the conversions. It's almost inaccurate to say `minutes` for a duration object.

server/server/src/config.rs
```diff
@@ -75,11 +76,48 @@ impl ServerConfig {
...
+    pub duration_between_metrics_refresh: Duration,
+    pub duration_between_user_metrics: Duration,
+    pub duration_between_getting_pub_key_metrics: Duration,
+    pub duration_between_getting_pub_key_key_metrics: Duration,
+}
+
+impl MetricsConfig {
+    pub fn from_env_vars() -> MetricsConfig {
+        MetricsConfig {
+            duration_between_metrics_refresh: Duration::from_secs(
```
Parth, line 90: seconds minutes?

server/server/src/metrics.rs
```diff
@@ -0,0 +1,215 @@
...
+        METRICS_STATISTICS.total_users.set(public_keys.len() as i64);
+
+        let mut total_documents = 0;
+        let mut total_bytes = 0;
+        let mut active_users = 0;
+
+        for public_key in public_keys {
+            let mut con = server_state.index_db_pool.get().await?;
+
+            let ids = get_users_owned_files(&mut con, &public_key).await?;
```
Parth, line 82: think about what is implied by the parameters and return values.

I'd call this `get_files`

server/server/src/metrics.rs
```diff
@@ -0,0 +1,215 @@
...
+
+            let metadatas = get_user_file_metadatas(&mut con, &ids).await?;
+            let bytes = calculate_total_document_bytes(&mut con, &metadatas).await?;
+            total_bytes += bytes;
+
+            if is_user_active(&metadatas).await? {
+                active_users += 1;
+            }
+
+            tokio::time::sleep(server_state.config.metrics.duration_between_user_metrics).await;
```
Parth, line 93: the type already contains the word duration. Think deeper about the names of these things

server/server/src/metrics.rs
Parth, line 82:
```suggestion
            let ids = get_owned(&mut con, &public_key).await?;
```

everything else is implied by parameters and return values

server/server/src/metrics.rs
```diff
@@ -0,0 +1,215 @@
...
+pub async fn start(server_state: Arc<ServerState>) -> Result<(), ServerError<MetricsError>> {
+    loop {
+        let public_keys = get_all_public_keys(&server_state).await?;
+        METRICS_STATISTICS.total_users.set(public_keys.len() as i64);
+
+        let mut total_documents = 0;
+        let mut total_bytes = 0;
+        let mut active_users = 0;
+
+        for public_key in public_keys {
```
Parth, line 79: good, readable, simple

server/server/src/metrics.rs
```diff
@@ -0,0 +1,215 @@
...
+            let mut con = server_state.index_db_pool.get().await?;
+
+            let ids = get_users_owned_files(&mut con, &public_key).await?;
+            total_documents += ids.len();
+
+            let metadatas = get_user_file_metadatas(&mut con, &ids).await?;
```
Parth, line 85:
```suggestion
            let metadatas = get_metadatas(&mut con, &ids).await?;
```

server/server/src/metrics.rs
```diff
@@ -0,0 +1,215 @@
...
+        tokio::time::sleep(server_state.config.metrics.duration_between_metrics_refresh).await;
+    }
+}
+
+pub async fn get_all_public_keys(
+    server_state: &Arc<ServerState>,
+) -> Result<Vec<PublicKey>, ServerError<MetricsError>> {
+    let mut con = server_state.index_db_pool.get().await?;
+
+    let mut public_keys_k: AsyncIter<String> = con.scan_match(public_key("*")).await?;
```
Parth, line 111: k?

server/server/src/metrics.rs
```diff
@@ -0,0 +1,215 @@
...
+
+pub async fn get_users_owned_files(
+    con: &mut deadpool_redis::Connection,
+    public_key: &PublicKey,
+) -> Result<Vec<Uuid>, ServerError<MetricsError>> {
+    con.maybe_json_get(keys::owned_files(public_key))
+        .await?
+        .ok_or_else(|| {
+            internal!(
+                "Cannot retrieve owned_files despite having a valid public_key: {:?}",
```
Parth, line 156:
```suggestion
                "Cannot retrieve owned_files for public_key: {:?}",
```

hard to determine what *valid* means if this is ever a message someone gets woken up with this.

server/server/src/metrics.rs
```diff
@@ -0,0 +1,215 @@
...
+    Ok(metadatas)
+}
+
+pub async fn calculate_total_document_bytes(
+    con: &mut deadpool_redis::Connection,
+    metadatas: &[EncryptedFileMetadata],
+) -> Result<i64, ServerError<MetricsError>> {
+    let mut total_size: u64 = 0;
+
+    for metadata in metadatas {
```
Parth, line 186: doing it this way is fine... but you could do:
`for metadata in metadatas.filter_not_deleted()` which will parse the tree and remove everything that's not deleted. Maybe you want to do that in `get_user_file_metadatas`, maybe not because a recently deleted file indicates user activity.

containers/local.env
```diff
@@ -8,6 +8,11 @@ FILES_DB_BUCKET=testbucket
 FILES_DB_ACCESS_KEY=minioadmin
 FILES_DB_SECRET_KEY=minioadmin
 
+MINUTES_BETWEEN_METRICS_REFRESH=5
+MILLIS_BETWEEN_USER_METRICS=50
+MILLIS_BETWEEN_GETTING_PUB_KEY_METRICS=5
+MILLIS_BETWEEN_GETTING_PUB_KEY_KEY_METRICS=5
```
Parth, line 14: I think these could just be one variable: millis_between_redis_calls, and I think 100 for all of these is probably the right place for us to start.

server/server/src/config.rs
SmailBarkouch: You can't do duration from_minutes. So when I retrieve the environment variable (which is in minutes), I just multiply it by 60 to get it in seconds.

server/server/src/metrics.rs
SmailBarkouch: Isn't it important to mention duration? I mean, I should elaborate this is a variable involving time, so I am not necessarily sure what is wrong with it. Is `time_between_user_metrics` better?

server/server/src/metrics.rs
SmailBarkouch: I think filtering deleted files isn't really necessary since they don't really cause any problems in the code. Right? Wouldn't filtering instead just make metrics run a bit longer.

server/server/src/config.rs
Parth: oh didn't see *60

server/server/src/metrics.rs
Parth: you actually do need to do it this way because right now we're counting deleted files towards our total file count.

but you need to do it in a way that deletion still counts towards active users. In the main loop `total_documents += ids.len();` is wrong because it will include deleted files. 

IMO in your main loop you should first calculate if a user is active and then set `metadatas = metadatas.filter_deleted()` and then the rest of your functions can remain unchanged.

SmailBarkouch: Okay that makes sense.

PR #975, opened 2022-02-09
rust: fmt config

Parth:
[reference](https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=)

@steverusso checkout stuff related to mods and imports, feel free to push commits directly to this one change at a time

### DIFF
--- a/clients/cli/src/backup.rs
+++ b/clients/cli/src/backup.rs
@@ -24,10 +24,7 @@ pub fn backup() -> CliResult<()> {
 
     let folders_len = list_paths(&config, Some(FoldersOnly))?.len();
 
-    println!(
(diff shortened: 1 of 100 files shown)


steverusso:
Both [`imports_granularity`](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#imports_granularity) and [`group_imports`](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#group_imports) are unstable. Adding `unstable_features = true` still gives the same warning, which is that unstable features are only available in the nightly channel.

steverusso (approved):
I think this is a great start while we await stability in the imports category.

PR #994, opened 2022-02-27
server: fix metrics errors and add user usage breakdown

### LINKED ISSUE #988: make metrics great again
There was an error thrown by [pager duty](https://lockbook.pagerduty.com/incidents/Q152IEQ35P8FCW), originating from server's metrics service. 

```
Cannot retrieve file usage for id: a554286c-6776-4be4-bd1c-6977d4bcb926
```

SmailBarkouch:
- fixes current problems with server metrics (fixes #988)
- adds an additional metric item for each users individual usage size


server/server/src/metrics.rs
```diff
@@ -159,7 +188,7 @@ pub async fn calculate_total_document_bytes(
     let mut total_size: u64 = 0;
 
     for metadata in metadatas {
-        if metadata.file_type == FileType::Document {
+        if metadata.file_type == FileType::Document && metadata.content_version != 0 {
             let file_usage: FileUsage = con
```
Parth, line 192: need a sleep here

Parth: also we still need to not error out here, a file may be deleted and no longer have a `size` while this function is running, metrics shouldn't abort because of this.

Parth (changes requested):

Parth (approved):

PR #1001, opened 2022-03-08
core: refactor importing files

steverusso:
This PR improves the file importing functionality (touches `core` as well as the `cli` and `linux` clients).

### Why

I've found the current implementation to be limited in the following ways:
1. I want to know when the work of importing a single item has started _and finished_. Right now, there is only the option to be alerted when an item has started.
2. When the process of importing an item is done, a graphical client with a tree might want to add an entry at that point. In the case of `linux` for example, this would require the `FileMetadata` of the newly imported file.

### Changes to Code

To start, `import_file` in lib.rs is now `import_files` and takes _at least one_ `PathBuf` to import. (Same change made for the `import_file` function in `import_export_service`.)

Within `import_export_service::import_file_recursively`, the notable change is that `file_service::create_file` is used instead of `path_service::create_file_at_path`. This removes the need to reason about strings and how a string ends in order to determine things such as file type.

In order to determine the stage of the import process as it relates to each individual item, the closure was modified to take a new enum, `ImportStatus`, as a single argument. This is used to communicate to the caller things like when an item is started, when it is done (with the newly created `FileMetadata`), and if there was an error.

The closure that is passed to receive progress updates is no longer optional. Errors while importing files and communicated via the closure and do not halt the entire import process.

### Changes to UI/UX

During import, I think that the destination of each item is both redundant (because they can deduce where things are going as they're the one who kicked off the import) and unimportant (imo the progress is far more important).

<details><summary>cli::before</summary>

image
</details>

<details><summary>cli::after</summary>

image
</details>

<details><summary>linux::before</summary>

image
</details>

<details><summary>linux::after</summary>

image

and if there are errors:

image
</details>

If there is consensus that the destination path actually is important information, it can easily be added back in, likely with a `path_service` call.

Name conflicts are automatically resolved by incrementing the conflicting name.

<details><summary>Here's an example of repeatedly importing the same file in cli.</summary>

image
</details>

### DIFF
--- a/clients/cli/src/copy.rs
+++ b/clients/cli/src/copy.rs
@@ -1,48 +1,48 @@
+use std::cell::Cell;
+use std::io::Write;
 use std::path::PathBuf;
 
+use lockbook_core::service::import_export_service::ImportStatus;
 use lockbook_core::{
-    create_file_at_path, get_file_by_path, import_file, CreateFileAtPathError, Error as CoreError,
-    Error, GetFileByPathError, ImportFileError,
+    create_file_at_path, get_file_by_path, import_files, CoreError, CreateFileAtPathError,
+    GetFileByPathError, ImportFileError,
 };
+use lockbook_models::file_metadata::DecryptedFileMetadata;
 
 use crate::error::CliResult;
 use crate::utils::{account, config};
 use crate::{err, err_unexpected};
-use lockbook_core::service::import_export_service::ImportExportFileInfo;
-use lockbook_models::file_metadata::DecryptedFileMetadata;
 
 pub fn copy(disk_paths: &[PathBuf], lb_path: &str) -> CliResult<()> {
     account()?;
 
-    let file_metadata = get_or_create_file(lb_path)?;
-
-    let import_progress = |info: ImportExportFileInfo| {
-        println!("importing: {} to {}", info.disk_path.display(), info.lockbook_path);
+    let total = Cell::new(0);
+    let nth_file = Cell::new(0);
+    let update_status = move |status: ImportStatus| match status {
+        ImportStatus::CalculatedTotal(n_files) => total.set(n_files),
+        ImportStatus::Error(disk_path, err) => match err {
+            CoreError::DiskPathInvalid => eprintln!("invalid disk path '{}'", disk_path.display()),
+            _ => eprintln!("unexpected error: {:#?}", err),
+        },
+        ImportStatus::StartingItem(disk_path) => {
+            nth_file.set(nth_file.get() + 1);
+            print!("({}/{}) Importing: {}... ", nth_file.get(), total.get(), disk_path);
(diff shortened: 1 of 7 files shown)


clients/cli/src/copy.rs
```diff
@@ -1,64 +1,64 @@
...
-use lockbook_models::file_metadata::DecryptedFileMetadata;
 
 pub fn copy(disk_paths: &[PathBuf], lb_path: &str) -> CliResult<()> {
     account()?;
 
-    let file_metadata = get_or_create_file(lb_path)?;
-
-    let import_progress = |info: ImportExportFileInfo| {
-        println!("importing: {} to {}", info.disk_path.display(), info.lockbook_path);
+    let total = Cell::new(0);
```
Parth, line 19: Is `Cell` actually needed here? These vars can just be local to the closure right?

clients/cli/src/copy.rs
```diff
@@ -1,64 +1,64 @@
...
 }
 
 fn get_or_create_file(lb_path: &str) -> CliResult<DecryptedFileMetadata> {
     // Try to get a file
     match get_file_by_path(&config()?, lb_path) {
         Ok(file) => return Ok(file),
         Err(err) => match err {
-            Error::UiError(GetFileByPathError::NoFileAtThatPath) => {} // Continue
-            Error::Unexpected(msg) => return Err(err_unexpected!("{}", msg)),
+            lockbook_core::Error::UiError(GetFileByPathError::NoFileAtThatPath) => {} // Continue
```
Parth, line 53: what's the pattern here?

Parth, line 53: what's the pattern here?

clients/linux/src/app.rs
```diff
@@ -694,69 +696,105 @@ impl LbApp {
...
+
+        d.get_content_area().add(&disk_lbl);
+        d.get_content_area().add(&pbar);
+        d.get_content_area().add(&prog_lbl);
 
         const FILE_SCHEME: &str = "file://";
 
-        let mut total = 0;
+        let total = Cell::new(0);
         let progress = Rc::new(RefCell::new(-1));
```
Parth, line 722: maybe we want to be uniform with types and initial values here?

Parth, line 722: maybe we want to be uniform with types and initial values here?

Parth (approved):
some nits, but lgtm, can probably give other "ProgressClosures" similar treatment

clients/cli/src/copy.rs
Parth: Nope it is actually needed here

Parth: nvm

clients/linux/src/app.rs
steverusso: How do you mean? As in `progress` should also be a `Cell`?

PR #1004, opened 2022-03-11
cli: tree printing function

### LINKED ISSUE #872: tree command
drop this crate in to viz a file tree: https://github.com/sighol/tree-rs

vvmirkovic:
Adding cli command to print a pretty file tree structure for a given path. 

fixes #872 

sample output:

```
vvmirkovic
├── first real document.md
├── fldr
│   ├── anotherfldr
│   │   ├── hi
│   │   └── hi2
│   └── tree_test.md
├── folder
├── folder2
├── folder3
├── folder4
├── hello_world.md
├── something_new.md
├── something.md
└── xfolder
    └── empty
```


clients/cli/Cargo.toml
```diff
@@ -18,3 +18,4 @@ chrono = "0.4"
 qr2term = "=0.2.1"
 hotwatch = "0.4.5"
 atty = "0.2"
+ptree = "0.4"
```
Parth, line 21: what made you choose this dep? could be fine but when I look at the repo for signs of popularity i see that it goes to a gitlab with 1 star. widely used software is safer in a few regards than niche software.

Parth: usually this will cause a change to the `.lock` file. it's a security mechanism, you want to check that bad boy in as well.

clients/cli/src/print_tree.rs
```diff
@@ -0,0 +1,30 @@
+use crate::error::CliResult;
```
Parth, line 1: CI seems to claim that your code doesn't compile. Does this work for you locally, or is this just pseudocode?

clients/cli/src/main.rs
```diff
@@ -112,6 +113,12 @@ enum Lockbook {
         path: String,
     },
 
+    /// Print the file tree with a given file as the head
+    PrintTree {
```
Parth, line 117: I think we just want this to be called `Tree`. We may put this in a `debug` subcommand, not sure yet, but `Print` is redundant.

clients/cli/src/main.rs
```diff
@@ -112,6 +113,12 @@ enum Lockbook {
...
+        /// Absolute path of a file !@#$%^&*I don't know what this is:(lockbook list-docs)
+        path: String,
```
Parth, line 119: I don't think you have to worry about only printing out the subtree, just print out the whole business, simpler that way too.

clients/cli/src/print_tree.rs
```diff
@@ -0,0 +1,30 @@
...
+    let cfg = config()?;
+
+    let file_metadata = get_file_by_path(&cfg, file_name).map_err(|err| match err {
+        CoreError::UiError(GetFileByPathError::NoFileAtThatPath) => {
+            err!(FileNotFound(file_name.to_string()))
+        }
+        CoreError::Unexpected(msg) => err_unexpected!("{}", msg),
+    })?;
+
+    // Temp copy and paste solution. To be reworked
```
Parth, line 18: Cool didn't realize it was at this stage of a WIP, I'll let you continue to iterate in peace to get to the point where you have something that's functional. And we can do the next round of feedback at that point.

clients/cli/Cargo.toml
```diff
@@ -18,3 +18,4 @@ chrono = "0.4"
 qr2term = "=0.2.1"
```
Parth, line 18: checkout: https://github.com/lockbook/lockbook/blob/master/docs/CONTRIBUTING.md for a handful of pr heading and description conventions

Parth (changes requested):
looking good, keep it up

clients/cli/src/utils.rs
```diff
@@ -178,6 +179,15 @@ pub fn edit_file_with_editor(file_location: &str) -> bool {
     }
 }
 
+pub fn metadatas() -> CliResult<Vec<DecryptedFileMetadata>> {
+    match list_metadatas(&config()?) {
+        Ok(metadatas) => Ok(metadatas),
+        Err(err) => match err {
+            UnexpectedError(msg) => Err(err_unexpected!("{}", msg)),
```
vvmirkovic, line 186: I think this is likely not the ideal way to handle this.

Parth (changes requested):

clients/cli/src/utils.rs
Parth: Fundamentally, this is fine, if we get an error this is what we want to return.

you can just `map_err` tho, anytime you do something like: `Ok(metadatas) => Ok(metadatas),` it's a sign there's probably a smaller way to write it.

roughly it'll end up being something like:
```
list_metadatas(&config()?)
    .map_err(|err| err_unexpected!("{}", err))
```

Parth (changes requested):

clients/cli/src/tree.rs
```diff
@@ -0,0 +1,53 @@
+use crate::error::CliResult;
+use crate::utils::{account, config};
+use lockbook_core::{get_children, get_root};
+use lockbook_models::file_metadata::DecryptedFileMetadata;
+use lockbook_core::model::state::Config;
+
+fn get_sorted_children(cfg: &Config, node: &DecryptedFileMetadata) -> Vec<DecryptedFileMetadata> {
+    let mut children = get_children(&cfg, node.id).unwrap_or_else(|err| {
```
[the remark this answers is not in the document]
vvmirkovic: Did you mean  to comment on this old commit?

core/libs/models/src/file_metadata.rs
```diff
@@ -70,6 +70,18 @@ impl FileMetadata for EncryptedFileMetadata {
     fn deleted(&self) -> bool {
         self.deleted
     }
+    fn display(&self) -> String {
+        match self.file_type() {
+            FileType::Folder => format!("{}{}", self.id, "/"),
```
Parth, line 75: minor detail I think we should make it clear that these are ids, also perhaps slighly simpler way to write this: 
```suggestion
            FileType::Folder => format!("id: {}/", self.id, "/"),
```

core/libs/models/src/file_metadata.rs
```diff
@@ -70,6 +70,18 @@ impl FileMetadata for EncryptedFileMetadata {
...
+            FileType::Document => format!("{}", self.id),
```
Parth, line 76:
```suggestion
            FileType::Document => format!("id: {}", self.id),
```

core/libs/models/src/tree.rs
```diff
@@ -64,8 +66,16 @@ pub trait FileMetaExt<T: FileMetadata> {
...
+    fn display(&self) -> Result<String, TreeError>;
 }
 
+// Why cant get work :(
+// impl Display for FileMetadata {
+//     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+//         self.display()
+//     }
+// }
+
```
Parth, line 78: Alright good attempt here but what you want to do is something like this 
```
impl<Fm> Display for [Fm]
where
    Fm: FileMetadata,
```

I think something like this will work?

You want to express that you'll provide a static dispatch parameterized implementation based on what they asked for (think duplication in the binary).

I think if you impl'd that you could do away with the `display()` fn you authored entirely and you don't need your function to be a part of the `FileMetaExt` at all, it can just be a self standing sort of implementation. Slightly neater.

Parth (changes requested):

Parth (approved):
Nice job, contribution number 2! Generally you want to slap a fixes #872 in your description to associate the solution with the problem. Let's see if adding it here will automatically close it.

PR #1005, opened 2022-03-11
infra: improve dev server script

steverusso:
* create the provided (or default) directory and `cd` into it when starting `redis` and `minio`
* use `command` to print out which programs are being used, or alternatively, to fail if one of them isn't found
* use `lsof` rather than `nc` for a slightly more robust way of listening for `minio`

### DIFF
--- a/utils/dev/run-server.sh
+++ b/utils/dev/run-server.sh
@@ -2,6 +2,10 @@
 
 set -ea
 
+command -V redis-server
+command -V minio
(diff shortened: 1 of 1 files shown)


Parth (approved):
Do we also want to do the song and dance of detecting if they have `minio` and `redis-server` in their path beforehand?

steverusso:
Hmm, yea we probably should.

PR #1007, opened 2022-03-13
core: fix overflow

tvanderstad:


core/src/service/file_service.rs
```diff
@@ -228,19 +228,7 @@ fn insert_metadata_given_decrypted_metadata(
...
-            {
-                base.deleted = true;
-                metadata_repo::insert(config, source, &base)?;
-            } else {
-                metadata_repo::insert(config, source, &encrypted_metadata)?;
-            }
-        } else {
-            metadata_repo::insert(config, source, &encrypted_metadata)?;
-        }
+        metadata_repo::insert(config, source, &encrypted_metadata)?;
```
tvanderstad, line 231: I am not sure what the purpose of the old code was, but in this case it cause a move to be undone, resulting in a cycle

core/src/service/file_service.rs
```diff
@@ -1908,10 +1896,10 @@ mod unit_tests {
             .unwrap();
         file_service::prune_deleted(config).unwrap();
 
-        assert_metadata_changes_count!(config, 1);
+        assert_metadata_changes_count!(config, 2);
         assert_document_changes_count!(config, 0);
         assert_metadata_count!(config, RepoSource::Base, 2);
-        assert_metadata_count!(config, RepoSource::Local, 2);
```
tvanderstad, line 1914: because a move for a deleted document is no longer undone, the document still has the folder as a parent, so the folder is not pruned

Parth:
fuzzing

PR #1020, opened 2022-03-17
cli: adding notepad support on windows and using pathbufs

vvmirkovic:
Adding notepad support to windows, and replacing string representations of paths with path buffs for better support across different OS


vvmirkovic:
Initial commit has replaced some strings with pathbufs, but only to a limited extent so far

Notepad has been added as a supported editor, and can be opened with cli. However, it still has a bug where notepad won't remember the file location when you use save or save as.

Parth (changes requested):
this is pretty close to being mergable, here are some things that caught my eye:

+ when you have types going into enums, you generally need to either `move` them in or `clone()` them like you've done. This is fine, I think some of those clones could be un-needed but clippy will yell at you if that's the case. But when you have fn's you generally don't need them to take the owned variant, you're generally fine with the borrowed version. This is why most fn's take `&str` instead of `String`. 
+ Windows stuff can get saved in the docs directory like we discussed, if it's easy to set the files as "hidden" to keep their folders clean, great, if not nbd. If it requires unsafe code starting to feel not worth it (making it possible for our app to crash isn't worth it).
+ OS errors being paths seems good to me.
+ `Notepad is only supported on windows! Set LOCKBOOK_EDITOR to a different editor.` I think you can architect this such that you never have to write this. You can make the `Notepad` variant a windows only feature, and then the rust compiler will not force you to have a `Notepad`  variant here.

clients/cli/src/utils.rs
```diff
@@ -184,7 +193,7 @@ pub fn print_last_successful_sync() -> CliResult<()> {
     Ok(())
 }
 
-pub fn set_up_auto_save(id: Uuid, location: String) -> Option<Hotwatch> {
+pub fn set_up_auto_save(id: Uuid, location: PathBuf) -> Option<Hotwatch> {
```
Parth, line 196: We used to accept a `String` here, as a rule of thumb, accepting `String`s in functions is a bad idea when you can instead accept an `&str`. Because you don't actually want ownership of a `String` and you force people to do `.clone()` type business needlessly so, rule of thumb with the old fn should have been `&str`.

If you read the docs about [`PathBuf`](https://doc.rust-lang.org/stable/std/path/struct.PathBuf.html) or [`Path`](https://doc.rust-lang.org/stable/std/path/struct.Path.html), you'll see that `PathBuf` is analogous to `String` (owned regular struct) and `Path` is analogous to an `&str`.

So here you really want to take a `Path`.

If you wanted to really go for excellence here, you could accept an `AsRef<Path>`, this is the common thing to do when dealing with the standard library.

Here is an example of this: https://github.com/lockbook/lockbook/blob/parth/hmdb/core/src/service/log_service.rs#L12

this allows a caller to hand in any of the following types:
+ `Path`
+ `String`
+ `&str`
+ `&String`
+ `PathBuf`
+ `&Path`
+ and much more.

Parth: Applies to all of the below functions.

Parth (changes requested):

Parth:
@tvanderstad can you test this on windows
@SmailBarkouch can you test this on linux?

clients/cli/src/error.rs
```diff
@@ -94,12 +95,13 @@ make_errkind_enum!(
...
-    34 => OsCouldNotDeleteFile(String, IoError),
-    35 => OsInvalidPath(String),
-    36 => OsFileCollision(String),
+    31 => OsCouldNotCreateDir(PathBuf, IoError),
+    32 => OsCouldNotCreateFile(PathBuf, IoError),
+    33 => OsCouldNotWriteFile(PathBuf, IoError),
+    34 => OsCouldNotDeleteFile(PathBuf, IoError),
+    34 => OsCouldNotDeleteFolder(PathBuf, IoError),
+    35 => OsInvalidPath(PathBuf),
+    36 => OsFileCollision(PathBuf),
```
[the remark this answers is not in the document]
steverusso: @Parth no we didn't consolidate, but we did reformat cli errors in general on the hmdb PR. I think it'd be best to leave them alone (revert back to `String`s) so that merging or rebasing will be easier.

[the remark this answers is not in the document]
Parth: Chatted and pathbuf can stay

clients/cli/src/utils.rs
```diff
@@ -96,22 +98,25 @@ pub fn get_editor() -> SupportedEditors {
...
+            editor
         }
     }
 }
 
-pub fn get_directory_location() -> CliResult<String> {
-    let result = format!("{}/{}", env::temp_dir().to_str().unwrap_or("/tmp"), Uuid::new_v4());
+pub fn get_directory_location() -> CliResult<PathBuf> {
+    let mut result = env::temp_dir();
+    result.push(format!("{}", Uuid::new_v4()));
```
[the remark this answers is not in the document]
vvmirkovic: There's no to_string or any equivalent for Uuid (at least I don't see any.  Here is how fmt is implemented, maybe it's insightful:
impl fmt::Display for Uuid {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        fmt::LowerHex::fmt(self, f)
    }
}

clients/cli/src/utils.rs
vvmirkovic: I managed to get a bunch of functions working with Asref instead of the hardcoded types. However, this one you showed complained a lot in particular. This is what it took:
196`<P: 'static + AsRef< Path> + Send>`
200`.watch(location.as_ref().to_path_buf(), move |event: Event| match event {`

vvmirkovic: I used AsRef in a few other functions. Let me know if you think they are good there

clients/cli/src/utils.rs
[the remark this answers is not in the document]
Parth: huh

image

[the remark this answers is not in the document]
Parth: search our code you'll find lots of examples of us to_stringing

clients/cli/src/error.rs
```diff
@@ -70,6 +70,7 @@ macro_rules! make_errkind_enum {
 }
 
 type IoError = std::io::Error;
+type PathBuf = std::path::PathBuf;
```
steverusso, line 73: I would just import `PathBuf` instead of type aliasing it.

steverusso (changes requested):

clients/cli/src/error.rs
Parth: definitely

clients/cli/Cargo.toml
```diff
@@ -16,4 +16,6 @@ serde_json = "1.0.44"
 chrono = "0.4"
 qr2term = "=0.2.1"
 hotwatch = "0.4.5"
-atty = "0.2"
\ No newline at end of file
+atty = "0.2"
+dirs = "4.0.0"
+winapi = "0.3.9"
```
Parth, line 21: are these used anywhere?

clients/cli/src/edit.rs
```diff
@@ -32,35 +31,38 @@ pub fn edit(file_name: &str) -> CliResult<()> {
...
             Err(err) => err.print(),
         }
     } else {
         eprintln!("Your editor indicated a problem, aborting and cleaning up");
     }
 
-    fs::remove_file(&temp_file_path).map_err(|err| err!(OsCouldNotDeleteFile(file_location, err)))
+    fs::remove_file(&file_buf).map_err(|err| err!(OsCouldNotDeleteFile(file_buf.clone(), err)))?;
+    file_buf.pop();
+    fs::remove_dir_all(&file_buf).map_err(|err| err!(OsCouldNotDeleteFolder(file_buf, err)))
```
Parth, line 67: doesn't `remove_dir_all` include `remove_file`?

Parth (changes requested):
editing files doesn't seem to work on macos anymore

Parth:
Taking a look at why macos isn't working: 
```
    let mut result = env::temp_dir();
    result.push(Uuid::new_v4().to_string());
    fs::create_dir(&result)
        .map_err(|err| err_unexpected!("couldn't open temporary file for writing: {:#?}", err))?;
    Ok(result)
```

result exists for sure, and my document is written to this location, but vim opens empty.

`edit_file_with_editor` also gets the right path... That vim command looks suspicious to me. You can debug the command just by printing it out, pretty neat.

Your current implementation sends this to `/bin/sh`:

```"/bin/bash" "-c" "</dev/tty vim" "/var/folders/53/z4dpd1397jj0fdysg3v_rwfc0000gn/T/29436361-f867-4267-b2a3-681c129abdef/todo.md"
```

this does not work

```"/bin/bash" "-c" "</dev/tty"  "vim" "/var/folders/53/z4dpd1397jj0fdysg3v_rwfc0000gn/T/29436361-f867-4267-b2a3-681c129abdef/todo.md"
```

unfortunately this does not work either

```"/bin/bash" "-c" "</dev/tty vim /var/folders/53/z4dpd1397jj0fdysg3v_rwfc0000gn/T/29436361-f867-4267-b2a3-681c129abdef/todo.md"
```

this does work though, this is the old way we did it. We could dive into the details of how `</dev/tty` works and what the right way to express this within `std::process::Command`, or we could just keep the old way for now using `format!` when using `#[cfg(not(target_os = "windows"))]`.

clients/cli/src/edit.rs
SmailBarkouch: Yeah, `remove_dir_all` recursively deletes.

Parth:
superseded by #1064

PR #1080, opened 2022-05-02
core: reproduce + fix infinite loop raised by fuzzer

tvanderstad:


core/src/service/file_service.rs
```diff
@@ -53,12 +53,11 @@ impl Tx<'_> {
...
-        self.insert_metadata_given_decrypted_metadata(
-            config,
-            source,
-            &all_metadata,
-            metadata_changes,
-        )
+        self.insert_metadata_given_decrypted_metadata(source, &all_metadata, metadata_changes)?;
+        if source == RepoSource::Local {
+            self.insert_new_docs(config, &all_metadata, metadata_changes)?;
+        }
```
tvanderstad, line 59: required as side effect of other changes

core/src/service/file_service.rs
```diff
@@ -146,7 +145,7 @@ impl Tx<'_> {
     /// Disk optimization opportunity: this function needlessly writes to disk when setting local metadata = base metadata.
     /// CPU optimization opportunity: this function needlessly decrypts all metadata rather than just ancestors of metadata parameter.
     fn insert_metadata_given_decrypted_metadata(
```
tvanderstad, line 147: no longer inserts new documents

core/src/service/file_service.rs
```diff
@@ -208,6 +197,20 @@ impl Tx<'_> {
         Ok(())
     }
 
+    fn insert_new_docs(
```
tvanderstad, line 200: factored out of `insert_metadata_given_decrypted_metadata`

core/src/service/file_service.rs
```diff
@@ -575,17 +578,16 @@ impl Tx<'_> {
...
             base_metadata_changes,
         )?;
         self.insert_metadata_given_decrypted_metadata(
-            config,
             RepoSource::Local,
             &local_metadata,
             local_metadata_changes,
-        )
+        )?;
+        self.insert_new_docs(config, &local_metadata, local_metadata_changes)
```
tvanderstad, line 590: `insert_metadata_given_decrypted_metadata` can't be responsible for adding new documents because to tell if a document is new, you need to read the current state, but in the event of cycle resolution the file tree will have a cycle in-between the two calls to `insert_metadata_given_decrypted_metadata` here. Also of note, `insert_new_docs` must be called after `insert_metadata_given_decrypted_metadata` so that it can check that the metadata for the docs it is inserting already exists, but it must use the version of `local_metadata` from before any calls to `insert_metadata_given_decrypted_metadata` because it only knows which documents are new based on which documents are missing from the local metadata (and they aren't missing anymore after the metadata changes are applied).

Parth (approved):

PR #1084, opened 2022-05-04
core: implement file tree for hashmap

### LINKED ISSUE #1059: Performance improvements in core
[hmdb](https://github.com/parth/hmdb) integrated into core has resulted in big performance gains. But our use of hmdb is still not really unlocking the full potential of hmdb. In many places we will ask hmdb for all metadata, as a vec, and then linear scan that data for a specific file. We could instead ask hmdb directly for that file. In many places this would turn `O(n)` operations to `O(1)`. In some situations it could turn `O(n^2)` operations into `O(1)` operations.

On a pretty new laptop our benchmarks take several hours to run, the operations we're doing are pretty modest (create 2000 files). It takes 15-30 minutes for core to create 2000 files. There is some fundamental limit to how fast we can do this (no matter what we have to read these files, compress them, encrypt them, and then write them to disk). But even tree only operations on all 2000 files while all metadata are already in memory take several seconds (TODO update after benchmarks complete). A reasonable overarching goal for this quest would be to get the bechmarks to all run in 1-2 hours. Ideally you'd make a set of changes and then kick the benchmarks off overnight. Ideally we reach a point where this isn't an overnight task.

Your first order of business if this is a mission you want to work on is probably to figure out a good way to run benchmarks. On some machine that can stay on for long hours. A laptop that you may need to shove into a backpack for instance is not a good candidate as the benchmarks would be interrupted. Our benchmarks output pretty graphs for before and after, and perform some statistics so you can know how good your changes have been. I'll begin brainstorming problematic areas of our code in this thread after #1053 is merged. But off the top of my head if we `impl`'d `FileMetaExt` on `HashMap<Uuid, FM>` instead of `Vec<FM` I think we'd see 80% of the gains we're looking for ([ref](https://github.com/lockbook/lockbook/blob/master/core/libs/models/src/tree.rs#L51)).

Generally though you should see what the benchmarks are doing as they're inspired by the various ways higher level clients do things. And begin doing research from there.

Another area of inquiry could be around network usage, as #961 suggests. But I think I can tackle that one.

Big improvements in this realm will provide us great value to our clients (anything that imports core), on our server (tree.rs is used in server), and would allow us to test far greater scenarios in our automated fuzzer. If you can allow users to import 2000 files in 10 seconds instead of 10 minutes it changes what people can do with our platform.

I think there's lower hanging fruit, but also on the table is doing async await within core.

vvmirkovic:
Our default "set of files" were always a `Vec<FileMetadata>` (whether encrypted or not). Alongside this we implemented several prototype-grade datastructure operations, which were almost always O(n) or in some cases O(n^2). We then called these very often in many places.

This includes, by default, using `HashMap<Uuid, FileMetadata>` which unravels many `O(n)` and `O(n^2)` operations into `O(1)` operations. `HashMap<Uuid, EncryptedFileMetadata>` is aliased to `EncryptedFiles` (and similar with the decrypted variant).

Benchmarks en-route. Hopefully this will result in big gains across the whole stack.

fixes #1059


steverusso (changes requested):

core/libs/models/src/tree.rs
```diff
@@ -123,133 +164,122 @@ where
...
 
-    /// Returns the files which are documents.
-    fn filter_documents(&self) -> Vec<Fm> {
-        self.iter()
-            .filter(|f| f.file_type() == FileType::Document)
-            .cloned()
+    fn filter_documents(&self) -> HashMap<Uuid, Fm> {
+        self.clone()
+            .into_iter()
+            .filter(|(_, f)| f.file_type() == FileType::Document)
```
steverusso, line 221:
```suggestion
            .filter(|(_, f)| f.file_type() == Document)
```
`FileType`'s variants got imported for this file with this PR. So if we are going to keep the variants, then let's drop the enum prefix just right here for now.

core/libs/test_utils/src/lib.rs
```diff
@@ -258,16 +258,16 @@ pub fn assert_server_work_paths(
...
-                .filter(|f| all_local_files.maybe_find(f.id).is_none())
-                .collect::<Vec<DecryptedFileMetadata>>();
+                .filter(|(id, _)| all_local_files.maybe_find(*id).is_none())
+                .collect::<DecryptedFiles>();
             all_local_files
                 .stage(&new_server_files)
                 .into_iter()
-                .map(|s| s.0)
-                .collect::<Vec<DecryptedFileMetadata>>()
+                .map(|s| (s.0, s.1 .0))
```
steverusso, line 269: Can what's going on here be a little clearer?

steverusso (changes requested):
Also: there's a few places where `insert(id, fm)` is used instead of the `push(fm)` fn. Should we prioritize `push` for readability?

Parth (approved):

steverusso (approved):

PR #1136, opened 2022-06-10
linux: refactor how the data directory is determined

steverusso:
Notably, include `HOMEPATH` as a last resort so that the gtk app can behave normally on windows.


Parth (approved):

Parth:
we do a very similar calculation in CLI, I wonder if there's overlap to be had in a future PR for all the desktop clients.

PR #1145, opened 2022-06-13
core: pre-sharing, pre-search, core data structure finalization

Parth:
With @vvmirkovic's `HashMap` changes, we saw agorithmic speedups in some instances, but regressions in other areas. Those regressions are largely caused by `O(n)` operations being more expensive with `HashMap`s, probably due to the lack of cache locality. This PR addresses all regressions, now core is faster in all situations compared to vecs (10 metas or 2000 metas).

In this PR I profile core and:
+ Stop scanning for root when we know root's id already ([a12dad2](https://github.com/lockbook/lockbook/pull/1145/commits/a12dad26731dd7bc67dc8357cf2e18165abd7a95))
+ Deal in terms of ids instead of `cloning` file metadatas, rework some algorithms to be `O(n)` instead of `O(n2)` ([1e49ed8](https://github.com/lockbook/lockbook/pull/1145/commits/1e49ed8f4323987151145469d14a3ee6b01137ba))
+ Rework the deleted calculation to be faster, deal in terms of ids. Also, for specific allocation heavy functions (`stage`) offer an allocation free, consuming, alternative. Use that where possible ([dc8a69a](https://github.com/lockbook/lockbook/pull/1145/commits/dc8a69a882fdd2351d94d8b93cd7dbd1d5c810d0)).
+ Create a place for core to persist volatile state, currently used for caching crypto primitives. These are things we do not want to put in our Database, but we do not want to recompute them for every core call ([7e52102](https://github.com/lockbook/lockbook/pull/1145/commits/7e52102d9097339d0b20bdedf2e70ac71c8abe15))  

Things I'd like to do in a Separate PR:
+ Eliminate many `config` parameters, now that everything is `impl`'d on `RequestContext`. `RequestContext` contains it's own `config`.
+ Rework `tree.rs` to push allocations up to the user of the trait (functions will either return a reference, or an id).
+ Something that will happen when I'm a little more certain about how I want to go about it: `stage` is the workhorse of `tree.rs` and is used heavily in all parts of the stack. All the tree functions should be flexible in what types they operate on. They should be able to operate on `HashMap`s, or more unique data structures, perhaps like a `StagedMetadata` data structure. Which consumes two trees and retains all the information about sources and is able to answer questions about which file is present where. If the tree functions were then additionally implemented for `StagedMetadata`, `cycle` and `path_conflict` detection (and many other things) would be faster because they, themselves, wouldn't be constantly staging things.
+ Something that is unlikely to happen, but is interesting so I'll mention it anyways: [not all HashMap implementations are slow to iterate](https://crates.io/crates/indexmap) once the first two things are sorted out, if we're still seeking more gains we could look in this direction, in no case do we care about insertion order for any of our maps, if we can get gains by expressing that, could be good. I don't expect us to go down this route anytime soon.

### DIFF
--- a/core/libs/models/src/tree.rs
+++ b/core/libs/models/src/tree.rs
@@ -26,6 +26,10 @@ pub trait FileMetadata: Clone + Display {
     fn is_document(&self) -> bool {
         self.file_type() == Document
     }
+
+    fn is_root(&self) -> bool {
+        self.id() == self.parent()
+    }
 }
 
 #[derive(Debug, Clone, PartialEq)]
@@ -147,7 +151,7 @@ where
     }
 
     fn maybe_find_root(&self) -> Option<Fm> {
-        self.values().find(|f| f.id() == f.parent()).cloned()
+        self.values().find(|f| f.is_root()).cloned()
     }
 
     fn maybe_find(&self, id: Uuid) -> Option<Fm> {
@@ -257,7 +261,7 @@ where
 
         for (id, (f, source)) in files_with_sources.iter() {
             let parent_id = f.parent();
-            if id == &parent_id {
+            if f.is_root() {
                 continue;
             };
             let name = f.name();
--- a/core/libs/models/src/tree.rs
+++ b/core/libs/models/src/tree.rs
@@ -99,7 +99,7 @@ pub trait FileMetaMapExt<Fm: FileMetadata> {
     fn maybe_find_parent(&self, id: Uuid) -> Option<&Fm>;
     fn find_children(&self, id: Uuid) -> HashMap<Uuid, Fm>;
     fn filter_deleted(self) -> Result<HashMap<Uuid, Fm>, TreeError>;
-    fn deleted(&self) -> Result<DeletedStatus, TreeError>;
+    fn deleted_status(&self) -> Result<DeletedStatus, TreeError>;
     fn filter_not_deleted(self) -> Result<HashMap<Uuid, Fm>, TreeError>;
     fn documents(&self) -> HashSet<Uuid>;
     fn parents(&self) -> HashSet<Uuid>;
@@ -204,14 +204,14 @@ where
 
     fn filter_deleted(self) -> Result<HashMap<Uuid, Fm>, TreeError> {
         let mut tree = self;
-        for id in tree.deleted()?.not_deleted {
+        for id in tree.deleted_status()?.not_deleted {
             tree.remove(&id);
         }
 
         Ok(tree)
     }
 
-    fn deleted(&self) -> Result<DeletedStatus, TreeError> {
+    fn deleted_status(&self) -> Result<DeletedStatus, TreeError> {
         let mut confirmed_not_delted = HashSet::new();
         let mut confirmed_deleted = HashSet::new();
         for meta in self.values() {
@@ -267,7 +267,7 @@ where
 
     fn filter_not_deleted(self) -> Result<HashMap<Uuid, Fm>, TreeError> {
         let mut tree = self;
-        for id in tree.deleted()?.deleted {
+        for id in tree.deleted_status()?.deleted {
             tree.remove(&id);
         }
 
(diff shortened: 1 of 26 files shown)


server/server/src/file_service.rs
```diff
@@ -141,19 +141,20 @@ async fn apply_changes(
...
-    for id in implicitly_deleted_ids {
-        if let Some(implicitly_deleted) = metas.maybe_find_mut(id) {
-            implicitly_deleted.deleted = true;
-            implicitly_deleted.metadata_version = now;
-            if implicitly_deleted.is_document() {
-                deleted_documents.push(implicitly_deleted.clone());
+        .deleted;
+
+    for id in deleted_ids {
+        if let Some(deleted) = metas.maybe_find_mut(id) {
```
steverusso, line 150: `deleted` -> `deleted_fm` (clearer var name)

core/libs/models/src/tree.rs
```diff
@@ -179,63 +202,116 @@ where
...
+    fn filter_deleted(self) -> Result<HashMap<Uuid, Fm>, TreeError> {
+        let mut tree = self;
+        for id in tree.deleted()?.not_deleted {
+            tree.remove(&id);
+        }
+
+        Ok(tree)
+    }
+
+    fn deleted(&self) -> Result<DeletedStatus, TreeError> {
```
steverusso, line 214: `deleted_status` clearer method

core/libs/models/src/tree.rs
```diff
@@ -179,63 +202,116 @@ where
...
         for (_, (f, _)) in staged_changes.clone().into_iter() {
             let mut checking = HashMap::new();
             let mut cur = &f;
-            while cur.id() != root.id() && prev_checked.get(&cur.id()).is_none() {
+
+            if !root_found && cur.is_root() {
+                root_found = true;
+            } else if root_found && cur.is_root() {
+                result.push(cur.id());
+            }
```
steverusso, line 312:
```suggestion
            if cur.is_root() {
                if root_found {
                    result.push(cur.id());
                } else {
                    root_found = true;
                }
            }
```

core/tests/sync_service_tests.rs
```diff
@@ -3311,12 +3311,11 @@ fn deleted_path_is_released() {
 
     db1.db
         .transaction(|tx| {
-            tx.insert_metadatum(
-                &db1.config,
-                RepoSource::Local,
-                &files::apply_delete(&tx.get_all_metadata(RepoSource::Local).unwrap(), file1.id)
-                    .unwrap(),
-            )
+            let mut tx = db1.context(tx).unwrap();
```
steverusso, line 3314: `tx` -> `ctx`?

core/tests/integrity_tests.rs
```diff
@@ -56,6 +57,7 @@ fn test_invalid_file_name_slash() {
     let doc = core.create_at_path(&path(&core, "document1.md")).unwrap();
     core.db
         .transaction(|tx| {
+            let mut tx = core.context(tx).unwrap();
```
steverusso, line 60: `tx` -> `ctx`?

core/tests/file_service_tests.rs
```diff
@@ -15,6 +15,7 @@ macro_rules! assert_metadata_changes_count (
     ($core:expr, $total:literal) => {
         assert_eq!(
             $core.db.transaction(|tx| {
+                let tx = $core.context(tx).unwrap();
```
steverusso, line 18: `tx` -> `ctx`?

steverusso (commented):
There are a few files (with many instances) where I think a `RequestContext` variable is named `tx`, whihc imo can be confusing. I think I left suggestions in all the files with these, feel free to resolve to ignore, because besides that, lgtm :+1:

utils/dev/run-server.sh
```diff
@@ -43,4 +43,4 @@ mc mb -p --region=$FILES_DB_REGION filesdb/$FILES_DB_BUCKET
 mc policy set public filesdb/$FILES_DB_BUCKET
 
 echo "Compiling and running lockbook server..."
-cargo run
+cargo run --release
```
steverusso, line 46: Lol

steverusso (approved):

PR #1153, opened 2022-06-21
infra: improve run-server util script

steverusso:
This will run `cargo check` before starting `minio` and `redis-server`.

Additionally, an environment variable (`DATA_DIR`) is used for the data directory, and all script args are passed to `cargo run`, so running release in a directory named "__datadir" can be done with:
```shell
DATA_DIR="./__datadir" ./utils/dev/run-server.sh --release
```

### DIFF
--- a/utils/dev/run-server.sh
+++ b/utils/dev/run-server.sh
@@ -9,7 +9,7 @@ command -V mc
 projRoot=`git rev-parse --show-toplevel`
 
 cd $projRoot/server/server
-cargo check
+cargo build $@
 
 cd $projRoot
 if [ -z "$DATA_DIR" ]; then


utils/dev/run-server.sh
```diff
@@ -8,6 +8,9 @@ command -V mc
 
 projRoot=`git rev-parse --show-toplevel`
 
+cd $projRoot/server/server
+cargo check
+
 dir="$1"
```
Parth, line 14: How would you feel about making the directory an env variable and cargo args `@`?

utils/dev/run-server.sh
```diff
@@ -8,25 +8,27 @@ command -V mc
 
 projRoot=`git rev-parse --show-toplevel`
 
-dir="$1"
-if [ -z "$dir" ]; then
-	dir="/tmp/lbdev"
-fi
+cd $projRoot/server/server
+cargo check
```
Parth, line 12: wondering if this is worthwhile, or in what circumstances you're gonna be facing 2 completely clean builds, but let's roll with it for now

steverusso: This is to solve for: build fails and mino and redis aren't terminated. I also think "does server compile" comes logically before "can minio and redis start."

Parth: yeah I wonder if your first pass at that was better then, `cargo build @` then run that binary

utils/dev/run-server.sh
```diff
@@ -8,25 +8,27 @@ command -V mc
 
 projRoot=`git rev-parse --show-toplevel`
 
-dir="$1"
-if [ -z "$dir" ]; then
-	dir="/tmp/lbdev"
-fi
+cd $projRoot/server/server
+cargo build $@
```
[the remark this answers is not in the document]
steverusso: Lol it comes through weird for me too, but it is indeed an `@` (at sign) character

PR #1179, opened 2022-06-27
core(tests): fix the db_equal assertion to include encrypted docs

steverusso:
As a result of https://github.com/lockbook/lockbook/pull/1175, @parth realized we weren't comparing documents when asserting db equality.


core/libs/test_utils/src/lib.rs
```diff
@@ -340,24 +340,7 @@ pub fn assert_deleted_files_pruned(core: &Core) {
...
-    ));
-    assert!(keys_match(
-        &left.db.local_metadata.get_all().unwrap(),
-        &right.db.local_metadata.get_all().unwrap(),
-    ));
-    assert!(keys_match(
-        &left.db.base_metadata.get_all().unwrap(),
-        &right.db.base_metadata.get_all().unwrap(),
-    ));
+    assert!(dbs_equal(left, right));
```
Parth, line 343: This is certainly nicer, but when a test is failing the prior version would tell us what database was unequal, which was a helpful starting point for debugging. There's also an earlier iteration in the history which would use `assert_eq!` which spits out both sides, intellij would then have a button that would allow you to `find difference`.

Parth (changes requested):

core/libs/test_utils/src/lib.rs
Parth: Not sure exactly what the best way to write it while not repeating ourselves.

steverusso: Ah ok, makes sense. For now I think I should just revert to how it was before.

Parth: cool

Parth (approved):

core/libs/test_utils/src/lib.rs
```diff
@@ -358,6 +358,10 @@ pub fn assert_dbs_eq(left: &Core, right: &Core) {
         &left.db.base_metadata.get_all().unwrap(),
         &right.db.base_metadata.get_all().unwrap(),
     ));
+    assert!(slices_equal_ignore_order(
+        &doc_repo_get_all(&left.config, RepoSource::Local),
```
Parth, line 362: you also want to assert this business for `RepoSource::Base`

Parth (changes requested):

steverusso:
As per [this discussion](https://lockbook.slack.com/archives/CPWR5L7SB/p1656366134983379) over Slack

PR #1181, opened 2022-06-28
server: hotfix existing user subscription info

SmailBarkouch:


Parth:
I piece of data was only inserted when an account was created, this redis call returns an error if a key is missing. This handles the optionality that exists for existing accounts.

This sort of error is not possible with hmdb or postgres

Parth (approved):

PR #1218, opened 2022-07-12
core, android: add file name and content search

### LINKED ISSUE #1165: add a searchbar

SmailBarkouch:
## Motivation

Searching is an essential feature for lockbook. We already had search implemented, but it worked only with file names and returned after scanning all the files. This search implementation will allow a user to get file names and contents while returning the results asynchronously as they come.

## Files of Interest
- `core/.../search_service.rs`: contains all the logic for this new implementation of search
- `clients/android/.../SearchedDocumentViewModel.kt`: all the complex logic on android's end, including the functions called from `java_interface.rs` (`addFileNameSearchResult`, `addFileContentSearchResult`, and `noMatch`)

## General Implementation

This iteration of search works off of channels to supply asynchronous matches. Two channels exist, one dealing with sending and receiving search queries and the other with sending and receiving search results. The search algorithm will start by copying all the contents of `.md` and `.txt` files into memory, storing the data in an `Arc`. Once it has all the data, the search moves off into a separate thread. There it will wait on its channels to receive search queries. Every search query creates a new thread where the results will be asynchronously streamed back to the clients using channels. Every new query stops the old query's thread. All the threads will end when a user wants to end their search.

## Search Algorithm
[This](https://crates.io/crates/sublime_fuzzy) is the search algorithm I used. I chose sublime's algorithm because it seemed to fit how we wanted to search function, with the way it [scored various matches](https://crates.io/crates/sublime_fuzzy). Not necessarily sure what the time complexity for this algorithm is since none of their docs mention it. All though, we can very easily be switched to [this](https://crates.io/crates/fuzzy-matcher) algorithm, which we previously used, if need be.

## Other Additions
Inspired by the Linux client, I added a transparent lockbook logo on the right side of the screen for tablets when no editor is open. See screenshots for specifics.

## Screenshots
<details>

</details>

## Video
<details>

https://user-images.githubusercontent.com/20663038/178838221-908d2deb-4393-4404-a033-8ccdd4c51517.mp4
</details>

fixes #1165


SmailBarkouch:
This pull request was split into #1223 and #1224.

PR #1224, opened 2022-07-15
core: search

### LINKED ISSUE #1225: add fuzzy search for file contents and file names that asynchronously returns results

SmailBarkouch:
## Motivation

Searching is an essential feature for lockbook. We already implemented search, but it worked only with file names and returned after scanning all the files. This search implementation will allow a user to get file names and contents while returning the results asynchronously as they come.

## File of Interest
- `core/.../search_service.rs`: contains all the logic for this new implementation of search

## General Implementation

This iteration of search works off of channels to supply asynchronous matches. Two channels exist, one dealing with sending and receiving search queries and the other with sending and receiving search results. The search algorithm will start by copying all the contents of `.md` and `.txt` files into memory, storing the data in an `Arc`. Once it has all the data, the search moves off into a separate thread. There it will wait on its channels to receive search queries. Every search query creates a new thread where the results will be asynchronously streamed back to the clients using channels. Every new query stops the old query's thread. All the threads will end when a user wants to end their search.

## Search Algorithm
[This](https://crates.io/crates/sublime_fuzzy) is the search algorithm I used. I chose sublime's algorithm because it seemed to fit how we wanted to search function, with the way it [scored various matches](https://crates.io/crates/sublime_fuzzy). Not necessarily sure what the time complexity for this algorithm is since none of their docs mention it. Although, we can very easily switch to [this](https://crates.io/crates/fuzzy-matcher) algorithm, which we previously used.

fixes #1225

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -46,6 +46,7 @@ lazy_static = "1.4.0"
 lockbook-models = { path = "libs/models" }
 lockbook-crypto = { path = "libs/crypto" }
 sublime_fuzzy = "0.7.0"
+crossbeam = "0.8.1"
 
 [profile.release]
 debug = true
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -45,6 +45,7 @@ hmdb = "0.1.12"
 lazy_static = "1.4.0"
 lockbook-models = { path = "libs/models" }
 lockbook-crypto = { path = "libs/crypto" }
+sublime_fuzzy = "0.7.0"
 
 [profile.release]
 debug = true
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -30,9 +30,7 @@ pub use crate::service::usage_service::{bytes_to_human, UsageItemMetric, UsageMe
 
 use std::collections::HashMap;
 use std::path::PathBuf;
-use std::sync::mpsc::{Receiver, Sender};
 use std::sync::{Arc, Mutex, MutexGuard};
-use std::thread::JoinHandle;
 
 use basic_human_duration::ChronoHumanDuration;
 use chrono::Duration;
@@ -50,7 +48,7 @@ use crate::model::errors::Error::UiError;
 use crate::model::repo::RepoSource;
 use crate::repo::schema::{transaction, CoreV1, OneKey, Tx};
 use crate::service::log_service;
-use crate::service::search_service::{SearchRequest, SearchResult, SearchResultItem};
+use crate::service::search_service::{SearchResultItem, StartSearchInfo};
 
 #[derive(Debug, Deserialize, Clone)]
 pub struct Config {
@@ -439,13 +437,9 @@ impl Core {
         Ok(val?)
     }
 
-    #[instrument(level = "debug", skip(self, results_tx, search_rx), err(Debug))]
-    pub fn start_search(
-        &self, results_tx: Sender<SearchResult>, search_rx: Receiver<SearchRequest>,
-    ) -> Result<JoinHandle<()>, UnexpectedError> {
-        let val = self
-            .db
-            .transaction(|tx| self.context(tx)?.start_search(results_tx, search_rx))?;
+    #[instrument(level = "debug", skip(self), err(Debug))]
+    pub fn start_search(&self) -> Result<StartSearchInfo, UnexpectedError> {
+        let val = self.db.transaction(|tx| self.context(tx)?.start_search())?;
         Ok(val?)
     }
 
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -30,7 +30,9 @@ pub use crate::service::usage_service::{bytes_to_human, UsageItemMetric, UsageMe
 
 use std::collections::HashMap;
 use std::path::PathBuf;
+use std::sync::mpsc::{Receiver, Sender};
 use std::sync::{Arc, Mutex, MutexGuard};
+use std::thread::JoinHandle;
 
 use basic_human_duration::ChronoHumanDuration;
 use chrono::Duration;
@@ -48,7 +50,7 @@ use crate::model::errors::Error::UiError;
 use crate::model::repo::RepoSource;
 use crate::repo::schema::{transaction, CoreV1, OneKey, Tx};
 use crate::service::log_service;
-use crate::service::search_service::SearchResultItem;
+use crate::service::search_service::{SearchRequest, SearchResult, SearchResultItem};
 
 #[derive(Debug, Deserialize, Clone)]
 pub struct Config {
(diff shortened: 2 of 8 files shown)


Parth (changes requested):

core/src/service/search_service.rs
```diff
@@ -49,4 +67,346 @@ impl RequestContext<'_, '_> {
...
+                                    should_continue,
+                                    input,
+                                ) {
+                                    if let Err(err) =
+                                        results_tx.send(SearchResult::Error(search_err))
+                                    {
+                                        warn!("Send failed: {:#?}", err);
+                                    }
+                                }
+                            });
```
Parth, line 157: What happens if the last search is still going on?

(I don't think this is an actual problem because of debounce, and how fast you say search results come in)

But I think it points to a larger data modeling issue.

Parth: I think you need to give some thought to:
1. how we can detect if a new search is stepping on an old search
2. what should the model of cancelling searches be? I get the sense there should be some vec of atomic bools or something like that. You want to be able to cancel all prior searches, there could be many.

You can test for all these things by adding sleeps.

core/Cargo.toml
```diff
@@ -45,6 +45,7 @@ hmdb = "0.1.12"
 lazy_static = "1.4.0"
 lockbook-models = { path = "libs/models" }
 lockbook-crypto = { path = "libs/crypto" }
+sublime_fuzzy = "0.7.0"
```
Parth, line 48: I can be sold that this is better, should we just use this for paths too?

Parth (changes requested):

core/src/lib.rs
```diff
@@ -437,6 +439,16 @@ impl Core {
         Ok(val?)
     }
 
+    #[instrument(level = "debug", skip(self, results_tx, search_rx), err(Debug))]
+    pub fn start_search(
+        &self, results_tx: Sender<SearchResult>, search_rx: Receiver<SearchRequest>,
+    ) -> Result<JoinHandle<()>, UnexpectedError> {
+        let val = self
```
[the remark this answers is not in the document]
steverusso: Agreed :+1:

core/src/service/search_service.rs
```diff
@@ -49,4 +67,346 @@ impl RequestContext<'_, '_> {
...
+                    None => match search_rx.recv() {
+                        Ok(search) => search,
+                        Err(_) => return Ok(()),
+                    },
+                },
+                Some(search) => {
+                    last_search = None;
+                    search
+                }
+            };
```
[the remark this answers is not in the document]
Parth: no thread.sleep is a nice side benefit

core/Cargo.toml
SmailBarkouch: Not sure, should I change it in this PR or leave it to another one?

core/src/service/search_service.rs
```diff
@@ -49,4 +67,345 @@ impl RequestContext<'_, '_> {
...
+            UnexpectedError("No matched indices.".to_string())
+        })?;
+
+        if *last_match < IDEAL_CONTENT_MATCH_LENGTH {
+            new_paragraph = new_paragraph
+                .chars()
+                .take(IDEAL_CONTENT_MATCH_LENGTH + CONTENT_MATCH_PADDING)
+                .collect();
+
+            new_paragraph.push_str("...");
```
steverusso, line 326: The following compiles, but can you make sure it still works as expected, and then commit it if you feel like it's better?

```suggestion
                .take(IDEAL_CONTENT_MATCH_LENGTH + CONTENT_MATCH_PADDING)
                .chain("...".chars())
                .collect();
```

steverusso (changes requested):

core/src/service/search_service.rs
```diff
@@ -1,10 +1,29 @@
+use crate::model::filename::DocumentType;
 use crate::model::repo::RepoSource;
-use crate::{CoreError, RequestContext};
+use crate::{CoreError, RequestContext, UnexpectedError};
+use crossbeam::channel;
+use crossbeam::channel::{Receiver, RecvTimeoutError, Sender};
```
steverusso, line 5:
```suggestion
use crossbeam::channel::{self, Receiver, RecvTimeoutError, Sender};
```

core/src/service/search_service.rs
```diff
@@ -1,10 +1,29 @@
...
 use fuzzy_matcher::skim::SkimMatcherV2;
 use fuzzy_matcher::FuzzyMatcher;
+use lockbook_models::file_metadata::DecryptedFiles;
+use lockbook_models::tree::FileMetadata;
+use serde::Serialize;
 use std::cmp::Ordering;
+use std::sync::atomic::AtomicBool;
+use std::sync::{atomic, Arc};
+use std::thread;
+use std::thread::JoinHandle;
```
steverusso, line 15:
```suggestion
use std::thread::{self, JoinHandle};
```

core/src/service/search_service.rs
```diff
@@ -1,10 +1,29 @@
...
+use crossbeam::channel;
+use crossbeam::channel::{Receiver, RecvTimeoutError, Sender};
 use fuzzy_matcher::skim::SkimMatcherV2;
 use fuzzy_matcher::FuzzyMatcher;
+use lockbook_models::file_metadata::DecryptedFiles;
+use lockbook_models::tree::FileMetadata;
+use serde::Serialize;
 use std::cmp::Ordering;
+use std::sync::atomic::AtomicBool;
+use std::sync::{atomic, Arc};
```
steverusso, line 13:
```suggestion
use std::sync::atomic::{self, AtomicBool};
use std::sync::Arc;
```

steverusso (changes requested):

core/src/service/search_service.rs
```diff
@@ -49,4 +68,350 @@ impl RequestContext<'_, '_> {
...
+        let mut files_info = Vec::new();
+
+        for file in files_not_deleted.values() {
+            if file.is_document() {
+                let content =
+                    match DocumentType::from_file_name_using_extension(&file.decrypted_name) {
+                        DocumentType::Text => self
+                            .read_document(self.config, RepoSource::Local, file.id)
+                            .map(|content| String::from(String::from_utf8_lossy(content.as_ref())))
+                            .map(Some)?,
```
steverusso, line 85: 1. Why `as_ref` instead of just borrowing?
2. `.to_string()` on the end is more concise (imo) than using `String::from`.
3. Why map the `Some` in a separate call instead of doing it all in one shot?
4. Are you cool with changing `content` to `bytes`?

The above would look like the following. Feel free to commit if you agree.

```suggestion
                            .map(|bytes| Some(String::from_utf8_lossy(&bytes).to_string()))?,
```

steverusso (changes requested):

core/src/service/search_service.rs
```diff
@@ -49,4 +66,344 @@ impl RequestContext<'_, '_> {
...
+                .best_match()
+            {
+                if fuzzy_match.score().is_positive() {
+                    if *no_matches {
+                        *no_matches = false;
+                    }
+
+                    if !should_continue.load(atomic::Ordering::Relaxed) {
+                        return Ok(());
+                    }
```
steverusso, line 210: Why do you do this check at the beginning of each iteration, and then here if the iteration is a match? Is at the beginning of the loop not enough?

SmailBarkouch: If `should_continue` becomes false after the start of the loop, I don't want it to send if it finds a match in that iteration. It will return an old query's result then.

steverusso: I see, ok.

Parth (approved):

PR #1238, opened 2022-07-26
server, cli: new admin cli

SmailBarkouch:


clients/cli/src/main.rs
```diff
@@ -214,6 +216,35 @@ enum Lockbook {
...
+
+    /// Subcommands of the new account feature flag
+    NewAccounts(FeaturesSwitchOptions),
+}
+
+#[derive(Debug, PartialEq, StructOpt)]
+pub enum FeaturesSwitchOptions {
+    ToggleOn,
+
+    ToggleOff,
```
Parth, line 247: friends don't let friends re-invent booleans

containers/local.env
```diff
@@ -15,6 +15,8 @@ GOOGLE_CLOUD_PUBSUB_NOTIFICATION_TOKEN="${GOOGLE_CLOUD_PUBSUB_NOTIFICATION_TOKEN
 MINUTES_BETWEEN_METRICS_REFRESH=5
 MILLIS_BETWEEN_REDIS_CALLS=100
 
+ADMINS="smailbarkouch"
```
Parth, line 18: locally this should probably be a user named admin

core/libs/models/src/api.rs
```diff
@@ -409,5 +410,66 @@ impl Request for GetSubscriptionInfoRequest {
...
+#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
+pub struct ToggleFeatureFlagRequest {
+    pub feature: FeatureFlag,
+    pub enable: bool,
+}
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
+pub enum ToggleFeatureFlagError {
+    Unauthorized,
+}
```
Parth, line 445: You can re-use errors if they're the same, just call this `AuthError` or something like that

core/src/model/errors.rs
```diff
@@ -853,6 +855,68 @@ impl From<CoreError> for Error<GetSubscriptionInfoError> {
...
+            CoreError::Unauthorized => UiError(GetFeatureFlagsStateError::Unauthorized),
+            CoreError::ServerUnreachable => UiError(GetFeatureFlagsStateError::CouldNotReachServer),
+            CoreError::ClientUpdateRequired => {
+                UiError(GetFeatureFlagsStateError::ClientUpdateRequired)
+            }
+            _ => unexpected!("{:#?}", e),
+        }
+    }
+}
+
```
Parth, line 919: dry

core/src/service/admin_service.rs
```diff
@@ -0,0 +1,55 @@
...
+        api_service::request(&account, ToggleFeatureFlagRequest { feature, enable }).map_err(
+            |err| match err {
+                ApiError::Endpoint(ToggleFeatureFlagError::Unauthorized) => CoreError::Unauthorized,
+                ApiError::SendFailed(_) => CoreError::ServerUnreachable,
+                ApiError::ClientUpdateRequired => CoreError::ClientUpdateRequired,
+                _ => core_err_unexpected(err),
+            },
+        )
+    }
+}
```
Parth, line 55: dry

server/server/src/account_service.rs
```diff
@@ -167,3 +177,22 @@ pub async fn delete_account(
...
+        })?;
+
+        if admin.0 == *public_key {
+            is_authorized = true;
+            break;
+        }
+    }
+
+    Ok(is_authorized)
+}
```
Parth, line 198: There should be no loop involved in this calculation:
1. lookup the public key's username
2. check if that username is an admin

server/server/src/account_service.rs
```diff
@@ -167,3 +177,22 @@ pub async fn delete_account(
 
     Ok(())
 }
+
+pub fn is_user_authorized<T: Debug>(
```
Parth, line 181: this should be `is_admin`

server/server/src/schema.rs
```diff
@@ -12,7 +13,8 @@ hmdb::schema! {
         metas: <Uuid, EncryptedFileMetadata>,
         sizes: <Uuid, u64>,
         google_play_ids: <String, Owner>,
-        stripe_ids: <String, Owner>
+        stripe_ids: <String, Owner>,
+        feature_flags: <FeatureFlag, bool>
```
Parth, line 17: I don't think this is the move, I think you should have `OneKey -> FeatureFlags` where `FeatureFlags` is a struct that contains the current settings for all the feature flags. Not every ff will be a bool.

Parth: if you do this a lot of `server/server/src/feature_flags.rs` goes away because you can do things like `unwrap_or_default` and you can specify the default values for the flags, and they only need to get set if you change them.

Parth (changes requested):

clients/cli/src/main.rs
SmailBarkouch: I didn't do the boolean stuff like we had in the old admin cli since it felt wrong. Unless you have something else in mind?

server/server/src/schema.rs
SmailBarkouch: I did it slightly different than what you are mentioning which I believe is cleaner. Check it out and let me know if you instead want me to do it exactly your way.

core/src/service/admin_service.rs
SmailBarkouch: I would need to `impl` for `ApiError`, which really isn't done for any other error. Wouldn't it be better to keep it as is for the sake of consistency?

clients/cli/src/error.rs
```diff
@@ -323,6 +331,11 @@ make_errcode_enum!(
     56 => DocumentReadError,
     57 => WarningsFound,
 
+    // Admin errors (58 - 59)
+
```
steverusso, line 335:
```suggestion
```

core/src/lib.rs
```diff
@@ -14,9 +14,12 @@ pub use uuid::Uuid;
 pub use lockbook_shared::account::Account;
 pub use lockbook_shared::api::{GooglePlayAccountState, StripeAccountTier, SubscriptionInfo};
 pub use lockbook_shared::api::{PaymentMethod, PaymentPlatform};
+use lockbook_shared::clock;
 pub use lockbook_shared::core_config::Config;
+use lockbook_shared::crypto::AESKey;
```
steverusso, line 19: Keep the non-`pub` use statements separate, like they were before.

steverusso (changes requested):

Parth (approved):

Issue #1199, opened 2022-07-06
make CI great again


Parth:
+ [ ] remove dependencies on docker
+ [ ] figure out if something [sccache](https://github.com/mozilla/sccache) is needed
+ [ ] run on an arch laptop

Parth:
+ Docker not a great tool from a cache perspective, Cargo is a great tool
+ un-needed parallelism -- cargo clippy at the top level, no cargo build for just a compilation check
+ want it to be in pure rust, no make, bash, etc glue
+ challenge 1, not building from scratch in each PR. Short term, just change `target` location to some global locaiton. long term: sscache for shared cache
+ cargo build && mv is not atomic, we could just name the output accordingly and then your program could move that binary over to that location renaming it to what your build script actually expects.

integration tests
+ cargo test --all, with API_URL specified for this PR
+ Generate a random number between valid port range, try to start the server, if it fails, retry the whole process. cargo build with a name and then running that binary is better than rebuilding
+ `cargo run utils/ci step1` 
+ `cargo run utils/ci step2`

PR #1249, opened 2022-08-05
infra: new ci

SmailBarkouch:
fixes: #1199


utils/dev-tool/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "lockbook-dev"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
steverusso, line 6: nit: drop this comment

utils/dev-tool/src/android.rs
```diff
@@ -0,0 +1,139 @@
+use crate::utils::HashInfo;
+use crate::{panic_if_unsuccessful, utils, ToolEnvironment};
```
steverusso, line 2:
```suggestion
use crate::utils::{self, HashInfo};
use crate::{panic_if_unsuccessful, ToolEnvironment};
```

utils/dev-tool/src/utils.rs
```diff
@@ -0,0 +1,168 @@
+use serde::{Deserialize, Serialize};
+use std::env::VarError;
+use std::fs::File;
+use std::io::Write;
+use std::path::{Path, PathBuf};
+use std::process::{Command, Stdio};
+use std::{env, fs};
```
steverusso, line 7:
```suggestion
use std::env::{self, VarError};
use std::fs::{self, File};
use std::io::Write;
use std::path::{Path, PathBuf};
use std::process::{Command, Stdio};
```

steverusso (changes requested):
(round 1)

utils/dev-tool/src/error.rs
```diff
@@ -0,0 +1,14 @@
+#[derive(Debug)]
```
steverusso, line 1: This entire file can go

utils/dev-tool/src/setup.rs
```diff
@@ -0,0 +1,34 @@
...
+        .stderr(Stdio::null())
+        .status()
+        .unwrap();
+
+    if !command_result.success() {
+        panic!("You do not have '{}'", command);
+    }
+}
+
+pub fn verify_env_var_exists(env_var: &str) {
```
steverusso, line 30: Neither of these functions need to be `pub`, and so imo they don't need to be functions if they're used in one spot in this small file.

utils/dev-tool/src/utils.rs
```diff
@@ -0,0 +1,168 @@
...
+        File::create(&self.hash_info_dir)
+            .unwrap()
+            .write_all(serde_json::to_string(self).unwrap().as_bytes())
+            .unwrap();
+    }
+}
+
+pub fn get_api_url(port: u16) -> String {
+    format!("http://localhost:{}", port)
+}
```
steverusso, line 168: nit: move this above with the other free floating functions

utils/dev-tool/src/utils.rs
```diff
@@ -0,0 +1,168 @@
+use serde::{Deserialize, Serialize};
+use std::env::VarError;
+use std::fs::File;
+use std::io::Write;
+use std::path::{Path, PathBuf};
+use std::process::{Command, Stdio};
+use std::{env, fs};
+
+#[macro_export]
+macro_rules! panic_if_unsuccessful {
```
steverusso, line 10: This is an interesting name for what is essentially an assert macro. I propose `assert_successful` or something along those lines instead.

Parth (changes requested):
intended order:
1. fmt
2. clippy
3. build & run server
4. run rust tests
5. android
6. build iOS (todo for parth later)
7. run kotlin tests
8. run swift 
9. cleanup

core/tests/upgrade_account_google_play_tests.rs
```diff
@@ -8,6 +8,7 @@ use test_utils::{
 };
 
 #[test]
+#[ignore]
```
Parth, line 11: close the issue

utils/dev-tool/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+
+[dependencies]
+structopt = "0.3.13"
+port_scanner = "0.1.5"
```
Parth, line 10: we probably don't want this, just pick a random number and later we'll have logic in there to have the server restart if it fails to start

Parth: between 1024-u16::max_value

utils/dev-tool/src/main.rs
```diff
@@ -0,0 +1,102 @@
...
+    /// Run swift integration tests
+    RunSwiftTests,
+
+    /// Kill server for commit hash
+    KillServer,
+
+    /// Kill all servers running
+    KillAllServers,
+}
+
```
Parth, line 59: order these in order of yaml file

utils/dev-tool/src/main.rs
```diff
@@ -0,0 +1,102 @@
...
+pub struct ToolEnvironment {
+    root_dir: PathBuf,
+    target_dir: PathBuf,
+    hash_info_dir: PathBuf,
+    commit_hash: String,
+}
+
+impl Default for ToolEnvironment {
+    fn default() -> Self {
+        let (root_dir, target_dir, hash_info_dir) = utils::get_dirs();
```
Parth, line 70: you should either move this impl into utils and inline the logic here, or you should have 3 fns that get you these three values

utils/dev-tool/src/utils.rs
```diff
@@ -0,0 +1,168 @@
...
+    ($success_code:expr) => {
+        if !$success_code.success() {
+            panic!()
+        }
+    };
```
Parth, line 15: I think there's something even more clever we can do let's discuss on call

Parth: do the trait thing everywhere

utils/dev-tool/src/utils.rs
```diff
@@ -0,0 +1,168 @@
...
+}
+
+pub fn tmp_dir() -> PathBuf {
```
Parth, line 18:
```suggestion
pub fn tmp() -> PathBuf {
```

utils/dev-tool/src/utils.rs
```diff
@@ -0,0 +1,168 @@
...
+pub fn local_env_path<P: AsRef<Path>>(root: P) -> PathBuf {
+    root.as_ref().join("containers/local.env")
+}
+
+pub fn test_env_path<P: AsRef<Path>>(root: P) -> PathBuf {
+    root.as_ref().join("containers/test.env")
+}
+
+pub fn get_commit_hash() -> String {
+    let commit_hash = Command::new("git")
```
Parth, line 55: github can hand you this info, not sure if it's better or not:
1. in your yml you can like export github specific yaml values as env vars, you could get things like PR number for instance this way

utils/dev-tool/src/server.rs
```diff
@@ -0,0 +1,86 @@
...
+use std::fs;
+use std::process::{Command, Stdio};
+
+pub fn build_server(tool_env: ToolEnvironment) {
+    dotenv::from_path(utils::local_env_path(&tool_env.root_dir)).unwrap();
+
+    let build_results = Command::new("cargo")
+        .args(["build", "-p", "lockbook-server"])
+        .status()
+        .unwrap();
```
Parth, line 13: you need to ask cargo to rename the server otherwise 2 prs will step on each other. Create then rename is not atomic

utils/dev-tool/src/server.rs
```diff
@@ -0,0 +1,86 @@
...
+    dotenv::from_path(utils::local_env_path(&tool_env.root_dir)).unwrap();
+
+    let mut hash_info = HashInfo::get_from_dir(&tool_env.hash_info_dir, &tool_env.commit_hash);
+
+    Command::new(&hash_info.server_binary_path)
+        .env("SERVER_PORT", port.to_string())
+        .stderr(Stdio::null())
+        .stdout(Stdio::null())
+        .spawn()
+        .unwrap();
```
Parth, line 39: in the future we want to wait 5s and see if the process is still running and then move on. Perhaps also we'd want to pipe logs somewhere.

core/tests/upgrade_account_google_play_tests.rs
SmailBarkouch: My PR doesn't fully complete the issue unfortunately.

utils/dev-tool/src/server.rs
SmailBarkouch: Renaming a binary on build is an unstable feature. I feel like it doesn't matter that much because any rebuild would take more time than rename wouldn't it?

utils/dev-tool/src/server.rs
SmailBarkouch: I ended up doing that since it was quick and easy.

utils/dev-tool/src/server.rs
Parth: then you should probably do `cargo run` instead

utils/dev-tool/src/server.rs
```diff
@@ -0,0 +1,93 @@
...
+        }
+    }
+}
+
+pub fn run_rust_tests(tool_env: ToolEnvironment) {
+    let hash_info = HashInfo::get_from_dir(&tool_env.hash_info_dir, &tool_env.commit_hash);
+    dotenv::from_path(utils::test_env_path(&tool_env.root_dir)).unwrap();
+
+    Command::new("cargo")
+        .args(["test", "--release", "--no-fail-fast", "--all", "--", "--nocapture"])
```
Parth, line 89: instead of `--all`, can you use `--workspace`? I think you don't need `--nocapture`, I think these days, failing tests will print their output. I think we also don't want `--no-fail-fast` so that we abort early.

Parth (changes requested):

utils/dev-tool/src/server.rs
Parth: this will run tests inside shared too

SmailBarkouch: okay

utils/dev-tool/src/setup.rs
```diff
@@ -0,0 +1,34 @@
+use std::env;
+use std::process::{Command, Stdio};
+
+const NEEDED_COMMANDS: [&str; 4] = ["cargo", "cbindgen", "swift", "cargo-ndk"];
+const NEEDED_ENV_VARS: [&str; 1] = ["ANDROID_HOME"];
```
steverusso, line 5: nit: "required" is more typical than "needed" imo

utils/dev-tool/src/setup.rs
```diff
@@ -0,0 +1,34 @@
...
+        .stderr(Stdio::null())
+        .status()
+        .unwrap();
+
+    if !command_result.success() {
+        panic!("You do not have '{}'", command);
+    }
+}
+
+fn verify_env_var_exists(env_var: &str) {
```
steverusso, line 30: This function exists for just an if statement, so I don't think it needs to be its own function. Same for the one above.

utils/dev-tool/src/main.rs
```diff
@@ -0,0 +1,110 @@
...
+use std::{env, fs};
+use structopt::StructOpt;
+
+pub mod android;
+pub mod apple;
+pub mod server;
+pub mod setup;
+pub mod utils;
+pub mod workspace;
+
```
steverusso, line 11: Place these first in the file.

utils/dev-tool/src/main.rs
```diff
@@ -0,0 +1,110 @@
...
+        let target_dir = utils::target_dir(&dev_dir, &root_dir);
+        let hash_info_dir = utils::hash_infos_dir(&dev_dir);
+
+        fs::create_dir_all(&dev_dir).unwrap();
+        fs::create_dir_all(&hash_info_dir).unwrap();
+        fs::create_dir_all(&target_dir).unwrap();
+
+        env::set_var("CARGO_TARGET_DIR", &target_dir.to_str().unwrap());
+
+        ToolEnvironment {
```
steverusso, line 81:
```suggestion
        Self {
```

utils/dev-tool/src/server.rs
```diff
@@ -0,0 +1,75 @@
+use crate::utils::{CommandRunner, HashInfo};
+use crate::{utils, ToolEnvironment};
```
steverusso, line 2:
```suggestion
use crate::utils::{self, CommandRunner, HashInfo};
use crate::ToolEnvironment;
```

utils/dev-tool/src/apple.rs
```diff
@@ -0,0 +1,67 @@
+use crate::utils::{CommandRunner, HashInfo};
+use crate::{utils, ToolEnvironment};
+use std::fs;
+use std::fs::File;
```
steverusso, line 4:
```suggestion
use crate::utils::{self, CommandRunner, HashInfo};
use crate::ToolEnvironment;
use std::fs::{self, File};
```

steverusso (changes requested):

utils/dev-tool/src/utils.rs
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    pub fn get_from_dir<P: AsRef<Path>>(hash_infos_dir: P, commit_hash: &str) -> Self {
+        let hash_info_dir = hash_infos_dir.as_ref().join(commit_hash);
+        match Self::maybe_get_at_path(hash_info_dir) {
+            None => {
+                panic!("No hash info file found. Server may not be running or even built!")
+            }
+            Some(hash_info) => hash_info,
+        }
```
steverusso, line 157: Isn't this essentially an `expect`?

steverusso (changes requested):

utils/dev-tool/src/main.rs
```diff
@@ -0,0 +1,110 @@
...
+    MakeKotlinTestLib,
+
+    /// Make swift jni libs for tests
+    MakeSwiftTestLib,
+
+    /// Kill all servers running
+    KillAllServers,
+}
+
+#[derive(Clone)]
```
steverusso, line 60: `Clone` is derived for use in two places where borrowing would just work. Curious the aversion to borrowing throughout?

steverusso (changes requested):

utils/dev-tool/src/main.rs
```diff
@@ -0,0 +1,104 @@
+pub mod android;
+pub mod apple;
+pub mod server;
+pub mod setup;
+pub mod utils;
+pub mod workspace;
```
steverusso, line 6: Why are these `pub`?

steverusso (changes requested):

utils/dev-tool/src/main.rs
```diff
@@ -0,0 +1,104 @@
...
+
+pub struct ToolEnvironment {
+    root_dir: PathBuf,
+    target_dir: PathBuf,
+    hash_info_dir: PathBuf,
+    commit_hash: String,
+}
+
+impl Default for ToolEnvironment {
+    fn default() -> Self {
```
steverusso, line 68: This function makes file system calls, so this isn't a good implementation of the `Default` trait. It's also unnecessary because you are just using it to initialize `ToolEnvironment` in the first line in main.

I propose just moving this code into main.

However, if for some reason you wanted to keep `Default` around, then just use it to set the default values and then place the file system / init logic somewhere else. But really I see no need for this.

Parth: I agree about the default part being weird, unsure whether the implementation detail needs to end up in main

.github/workflows/integration.yml
```diff
@@ -9,188 +9,27 @@ on:
...
+    - name: Run Server
+      run: cargo run -p lockbook-dev -- run-server
+    - name: Run Rust Tests
+      run: cargo run -p lockbook-dev -- run-rust-tests
+    - name: Run Android Fmt
+      run: cargo run -p lockbook-dev -- check-android-fmt
+    - name: Run Android Lint
+      run: cargo run -p lockbook-dev -- check-android-lint
+    - name: Run Swift Tests
+      run: cargo run -p lockbook-dev -- run-swift-tests
```
Parth, line 32: for now comment this out, I will fix these shortly, but that would make this a green build, and we can require it for pr's immediately

utils/dev-tool/src/main.rs
steverusso: I mean the body of the `default` function as it is now can basically just be moved into `main`

.github/workflows/integration.yml
Parth: never mind actually, we should just fix these asap

utils/dev-tool/src/main.rs
Parth: inlined into main :heavy_check_mark:

PR #1263, opened 2022-08-12
server: fix async log spans

### LINKED ISSUE #1250: tracing spans are setup incorrectly
we have to use something like `in_scope` when we're in async futures: https://github.com/lockbook/lockbook/blob/9d4a990183d5f7eeb77b85b7a6206595f4e580ca/server/server/src/router_service.rs#L92

but the whole thing is an async future: https://github.com/lockbook/lockbook/blob/9d4a990183d5f7eeb77b85b7a6206595f4e580ca/server/server/src/router_service.rs#L46

we probably want to, along with #1239, do the tracing stuff in a more warp-native way.

without this, this is what log output can look like:

image

Parth:
before (during heavy concurrent load): 
image
disaster: you can see that there are multiple overlapping pieces of metadata with each log entry.

after:
image

I misread the docs here: https://docs.rs/tracing/latest/tracing/struct.Span.html#method.enter

It says `in_scope` is for _synchronous_ code.

fixes #1250

### DIFF
--- a/server/server/src/router_service.rs
+++ b/server/server/src/router_service.rs
@@ -42,71 +42,74 @@ macro_rules! core_req {
             .and(warp::path(&<$Req>::ROUTE[1..]))
             .and(warp::any().map(move || cloned_state.clone()))
             .and(warp::body::bytes())
-            .then(|state: Arc<ServerState>, request: Bytes| async move {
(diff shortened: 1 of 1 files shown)


steverusso (approved):
Nice

PR #1284, opened 2022-08-16
the egui client

### LINKED ISSUE #229: Markdown editing

### LINKED ISSUE #957: windows: move is broken
after moving a file and refreshing the file view (e.g. by syncing), the move is undone

### LINKED ISSUE #1075: windows/core: windows produces invalid drawings; core doesn't validate
steps:
1. create a drawing on windows
2. sync to macos
3. open the drawing

expected:
- drawing opens

actual:
- invalid alpha value

note: alpha is checked to be in the range [0, 1] by `export_drawing` but not by `save_drawing`. windows exports a value in the range [0, 255].

### LINKED ISSUE #1076: windows: crash on invalid drawing
steps:
- make a new drawing with content `{"currentView":{"transformation":{"scale":0.2816802,"translation":{"x":1304.0437,"y":2291.7407}}},"events":[]}`
- open the drawing on windows -> crash

I am not sure if this is a valid drawing or how it got to be this way but I am pretty sure infinite loading and crashing is not the target behavior here.

At one point I had the following experience on macos but now I can't seem to reproduce: when opening the drawing there was a progress spinner where the file editor should be, it spun forever and I could not select another document, so the app was stuck until I restarted

### LINKED ISSUE #1156: windows light mode support

steverusso:
This is an alpha version of the `egui` client! At a minimum, this will be our web client and (for the foreseeable future) our windows client. However, `egui` is cross platform and works well on Linux and will also work on MacOS.

---

### Immediate Mode (Description and Links)

Currently, every graphical Lockbook client is built using a [_retained mode_](https://en.wikipedia.org/wiki/Retained_mode) UI toolkit. This means we hand those libraries our application state and the [library takes care of rendering](https://twitter.com/ianh_/status/1136753753720147969).

[Egui](https://github.com/emilk/egui#why-immediate-mode) is an [_immediate mode_](https://eliasnaur.com/blog/immediate-mode-gui-programming) library. This means we control our state and are responsible for when and how things are drawn.

---

### Code Organization

I tried to organize the code as intuitively as possible in line with the highest level stages of our app state:
1. Loading core / initialization. (`src/splash.rs`)
2. Onboard screen. (`src/onboard`)
3. Account screen. (`src/account`)

Outside of that:
* `src/model.rs`, `src/util.rs` - self explanatory
* `src/theme` - icons, drawing palette colors, syncing theme with system theme, etc.

---

### Functionality Checklist

- [x] Create account
- [x] Import account
- [x] Export account to clipboard (not QR code yet)
- [x] Sync
- [x] Space utilized
- [x] Create file
- [x] Rename file
- [x] Move files
- [x] Delete files (via right-click -> Delete) 
- [x] Markdown
- [x] Drawings
- [x] Images (untested)
- [ ] Auto sync
- [ ] Auto save
- [ ] Export files to host
- [ ] Import files from host
- [ ] Billing
- [ ] Sharing

---

### Notable TODOs

Besides the missing functionality listed above, there are still many, many things left to do with this client, and they'll each get an issue once this is merged. However, to mention a few big ones right now:

* Prompt the user to backup their account after they make one.
* Use the enhanced search in core instead of just the file path search.
* Nicer tabs for opened files with close buttons and middle-click close ability.
* Nicer settings tabs: icons above the labels.
* Markdown rendering is probably only 10% of what it can and should be.
* Syncing doesn't update file tree state.

---

### Closes

Assuming this PR deletes the current windows client...
* Closes #229.
* Closes #957.
* Closes #1075.
* Closes #1076.
* Closes #1156.


clients/egui/src/account/modals/help.rs
```diff
@@ -0,0 +1,57 @@
...
+                    ui.label(egui::RichText::new("Shortcut").strong().underline());
+                });
+                header.col(|ui| {
+                    ui.label(egui::RichText::new("Action").strong().underline());
+                });
+            })
+            .body(|mut body| {
+                for (k, v) in [
+                    ("Ctrl-N", "Open the New File prompt"),
+                    ("Ctrl-Space, Ctrl-L", "Open the search prompt"),
```
SmailBarkouch, line 25: The first column seems to be too small to show both shortcuts.

clients/egui/src/account/modals/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+pub use error::ErrorModal;
+pub use help::HelpModal;
+pub use new_file::{NewFileModal, NewFileParams};
+pub use search::SearchModal;
+pub use settings::SettingsModal;
+
+use eframe::egui;
+
+#[derive(Default)]
+pub struct Modals {
```
SmailBarkouch, line 18: I think "model" might be the appropriate spelling for this and the other instances using this word.

clients/egui/src/account/tabs/markdown/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+
+                    if out.response.changed() {
+                        self.events = parse(&self.content);
+                    }
+                });
+            });
+    }
+
+    fn draw_rendered(&mut self, ui: &mut egui::Ui) {
+        //const SPACE_AFTER_PARA: f32 = 17.0;
```
SmailBarkouch, line 70:
```suggestion
```

clients/egui/src/account/tabs/markdown/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+                ui.allocate_ui_with_layout(initial_size, layout, |ui| {
+                    ui.spacing_mut().item_spacing.x = 0.0;
+                    ui.spacing_mut().item_spacing.y = 0.0;
+                    ui.set_row_height(17.0);
+
+                    let mut s = Styling::default();
+                    let mut in_list = false;
+
+                    for el in &self.events {
+                        //println!("{:?}", el);
```
SmailBarkouch, line 90:
```suggestion
```

clients/egui/src/account/tree/node.rs
```diff
@@ -0,0 +1,384 @@
...
+        self.children = all_metas
+            .find_children(self.file.id)
+            .values()
+            .map(|f| {
+                let mut node = TreeNode::from((f.clone(), self.depth + 1));
+                node.populate_from(all_metas);
+                node
+            })
+            .collect();
+            */
```
SmailBarkouch, line 67:
```suggestion

```

clients/egui/src/account/modals/mod.rs
SmailBarkouch: You also used the spelling "model" for a file.

clients/egui/src/onboard.rs
```diff
@@ -0,0 +1,335 @@
...
+                },
+            }
+        }
+
+        egui::CentralPanel::default().show(ctx, |ui| {
+            ui.vertical_centered(|ui| {
+                ui.set_max_width(400.0);
+
+                ui.add_space(40.0);
+                self.logo.show_scaled(ui, 1.0);
```
SmailBarkouch, line 129: I feel like the logo shouldn't be translucent and should follow the other lockbook clients by having a solid one.

clients/egui/src/account/modals/mod.rs
steverusso: Nope, it's [modal](https://www.quora.com/Whats-the-difference-between-a-modal-and-dialog) haha. If it's confusing enough though I can switch back to dialog, but modal is more appropriate imo. In the GTK client, all of these dialogs are `modal` (a GTK property) for the very reason that they present content that must be addressed on its own while disallowing interaction with the main layer of the application.

I used the word "model" for miscellaneous data modelling, like we do in core and other places.

clients/egui/src/onboard.rs
steverusso: Have we chosen one specific logo? Or just any non-transparent one will do?

clients/egui/src/account/modals/mod.rs
SmailBarkouch: Oh, I didn't realize you were using it in that sense. Fair enough.

clients/egui/src/onboard.rs
steverusso: It looked much worse than this to me, but I still kept it around in the client directory to easily switch back.

SmailBarkouch: I think every client uses a different variation of the logo. So, in my opinion, any is fine, but maybe @Parth wants to chime in.

clients/egui/src/account/modals/help.rs
steverusso: Good catch, https://github.com/lockbook/lockbook/pull/1284/commits/97a4f76cd15d77b92d4e64a03f944fdd54eefef2

clients/egui/src/onboard.rs
Parth: yeah we should unify, but not in this pr

SmailBarkouch (changes requested):
Getting this behavior:

image

steverusso:
@SmailBarkouch https://github.com/lockbook/lockbook/pull/1284/commits/c9ef2653a5fdcb77a3dcc76307129f8830fdfa6d

Parth (approved):
This is good to go as soon as we have a windows installer POC

SmailBarkouch (approved):

PR #1289, opened 2022-08-17
infra: create releaser

Parth:
Push a button to:
+ [x] Build server
+ [x] Release it to our server

In a different PR this will also release the various clients we have to the various distribution channels we use.

I didn't make it part of the existing devtool because the only thing I'd want from there is probably the `assert_success` bits. But even building apple for CI (linux) is different from building it for release. So I figured an isolated binary is probably going to be a better time for me.

### DIFF
--- a/utils/releaser/Cargo.toml
+++ b/utils/releaser/Cargo.toml
@@ -6,3 +6,4 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+structopt = "0.3.13"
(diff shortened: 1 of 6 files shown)


utils/releaser/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "releaser"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
steverusso, line 6:
```suggestion
```

steverusso (changes requested):
Nice!

steverusso (approved):

PR #1322, opened 2022-08-24
core: resolve links when listing files

Parth:


core/tests/sharing_tests.rs
```diff
@@ -1024,3 +1025,67 @@ fn delete_write_shared_folder() {
...
+    let accounts = cores
+        .iter()
+        .map(|core| core.get_account().unwrap())
+        .collect::<Vec<_>>();
+    let roots = cores
+        .iter()
+        .map(|core| core.get_root().unwrap())
+        .collect::<Vec<_>>();
+
+    let folder = cores[0]
```
steverusso, line 1041: This is created only for the `id` it seems? If so, could just grab id as you do with `link_id` below.

Parth: [c830e43](https://github.com/lockbook/lockbook/pull/1322/commits/c830e43226c06391e2d4c7b03b555ec0630620b5)

core/src/service/file_service.rs
```diff
@@ -162,9 +191,14 @@ impl RequestContext<'_, '_> {
             .get(&OneKey {})
             .ok_or(CoreError::AccountNonexistent)?;
 
-        let mut files = vec![tree.finalize(id, account)?];
+        let id = match tree.find(id)?.file_type() {
+            FileType::Document | FileType::Folder => *id,
+            FileType::Link { target } => target,
+        };
```
tvanderstad, line 197: This will probably fail in some way when called on the ancestor of a link

PR #1325, opened 2022-08-24
cli, server: admin can now disappear files

Parth:


clients/cli/src/admin.rs
```diff
@@ -1,19 +1,48 @@
-use crate::CliError;
+use crate::{CliError, Uuid};
```
steverusso, line 1: Import `Uuid` from lockbook_core

steverusso (changes requested):
Rest lgtm

PR #1327, opened 2022-08-24
infra: removed hash info and added server logs step to ci

SmailBarkouch:


.github/workflows/integration.yml
```diff
@@ -14,6 +14,7 @@ jobs:
     env:
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       LOCKBOOK_CI: 1
```
Parth, line 16: [github](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables) has an env var for CI already we should just use that and avoid thinking and dealing with this.

utils/dev-tool/src/main.rs
```diff
@@ -52,40 +55,26 @@ enum Commands {
...
-    let hash_info_dir = utils::hash_infos_dir(&dev_dir);
+    let server_db_dir = utils::server_dbs_dir(&dev_dir);
 
-    fs::create_dir_all(&dev_dir).unwrap();
-    fs::create_dir_all(&hash_info_dir).unwrap();
     fs::create_dir_all(&target_dir).unwrap();
-    fs::create_dir_all(&server_dbs_dir).unwrap();
+    fs::create_dir_all(&server_db_dir).unwrap();
 
     env::set_var("CARGO_TARGET_DIR", &target_dir.to_str().unwrap());
```
Parth, line 75: I think you can just set this in `integration.yml` once and then not have this program think about this, ever, right?

utils/dev-tool/src/utils.rs
```diff
@@ -1,10 +1,9 @@
-use serde::{Deserialize, Serialize};
 use std::env::{self, VarError};
-use std::fs::{self, File};
-use std::io::Write;
 use std::path::{Path, PathBuf};
 use std::process::{Command, Output, Stdio};
 
+pub const SERVER_PORT: u16 = 8501;
```
Parth, line 5: Why are you overriding this at all? Why not use the default `test.env` values?

utils/dev-tool/src/server.rs
```diff
@@ -1,80 +1,54 @@
...
-            .env("INDEX_DB_LOCATION", &server_db_dir)
-            .current_dir(&tool_env.root_dir)
-            .stderr(Stdio::null())
-            .stdout(Stdio::null())
-            .spawn()
-            .unwrap();
+    let mut run_result = Command::new(tool_env.target_dir.join("release/lockbook-server"))
+        .env("SERVER_PORT", SERVER_PORT.to_string())
+        .env("INDEX_DB_LOCATION", &tool_env.server_db_dir)
+        .env("LOG_PATH", &tool_env.server_db_dir)
```
Parth, line 18: Why are you overriding this at all? Why not use the default `test.env` values?

utils/dev-tool/src/server.rs
```diff
@@ -1,80 +1,54 @@
...
 
-        thread::sleep(Duration::from_millis(5000));
+    thread::sleep(Duration::from_millis(5000));
 
-        if run_result.try_wait().unwrap().is_none() {
-            break;
-        }
+    if run_result.try_wait().unwrap().is_some() {
+        panic!("Server failed to start on port: '{}'", SERVER_PORT)
     }
```
Parth, line 29: I thought about this more, and I'm pretty certain we don't want this tool to build then run. I think we want to cargo run, and check if the server is up by polling `http://localhost:8000/get-build-info`.

Either the server will be up, then you're good to go, or the command will terminate early (fail to compile or start).

clients/egui/build.rs
```diff
@@ -1,6 +1,8 @@
 #[cfg(windows)]
 fn main() -> std::io::Result<()> {
-    winres::WindowsResource::new().set_icon("lockbook.ico").compile()
+    winres::WindowsResource::new()
+        .set_icon("lockbook.ico")
+        .compile()
```
Parth, line 5: If you rebase this will disappear from your diff

Parth (changes requested):

utils/dev-tool/src/server.rs
SmailBarkouch: Fair point, defaults can be used.

Parth (approved):

PR #1334, opened 2022-08-25
infra: android release automation

SmailBarkouch:


utils/releaser/src/apple/core.rs
```diff
@@ -11,7 +11,7 @@ static MAC_LIB_DIR: &str = "clients/apple/CLockbookCore/Sources/CLockbookCore/li
 static LIB: &str = "liblockbook_core.a";
 static HEAD: &str = "lockbook_core.h";
 
-pub fn build() {
+pub fn build_apple() {
```
Parth, line 14: why? The full path of this function is `apple::core::build`, so you don't need to specify `apple`.

utils/releaser/src/main.rs
```diff
@@ -10,11 +11,14 @@ use structopt::StructOpt;
...
+    ReleaseAndroid,
 }
 
-fn main() {
+#[tokio::main]
+async fn main() {
     match Releaser::from_args() {
         Releaser::DeployServer => server::deploy_server(),
         Releaser::ReleaseApple => apple::release_apple(&Github::env(), &AppStore::env()),
+        Releaser::ReleaseAndroid => android::release_android(&Github::env()).await,
```
Parth, line 22: You should make an `Android::env()` in secrets, so that, I know rust doesn't use it and only gradle will, but this will allow us to fail much earlier if someone's machine is not setup properly. In the `env()` function you create you can also do some basic validations that assert that whatever file they're providing via env vars actually exists.

Parth (changes requested):

Parth (approved):

PR #1345, opened 2022-08-28
core: assorted sharing fixes

tvanderstad:


core/libs/shared/src/core_ops.rs
```diff
@@ -43,7 +44,7 @@ where
             file_type,
             last_modified,
             last_modified_by,
-            shares: Vec::new(),
+            shares: Vec::new(), // todo
```
tvanderstad, line 47: shares use the sharer/sharee's usernames, so this will require us to add a public key -> username cache

core/libs/shared/src/core_ops.rs
```diff
@@ -670,11 +696,32 @@ where
...
+                                local_change,
+                                user_access_keys,
+                            )
                         };
 
                         let key = result.decrypt_key(&id, account)?;
-                        let parent_key = result.decrypt_key(remote_change.parent(), account)?;
+                        let name = {
+                            // we may not have the parent of a direct share
+                            // in that case changes are unauthorized anyway
```
tvanderstad, line 716: before this change, any concurrent edit on a directly shared file would make sync persistently fail

core/libs/shared/src/lazy.rs
```diff
@@ -32,31 +32,34 @@ impl<T: Stagable> LazyTree<T> {
 }
 
 impl<T: Stagable> LazyTree<T> {
-    // todo: use greatest of all access modes instead of access mode of nearest keyed ancestor
```
tvanderstad, line 35: done

core/libs/shared/src/lazy.rs
```diff
@@ -181,10 +184,10 @@ impl<T: Stagable> LazyTree<T> {
         Ok(name)
     }
 
-    pub fn link(&self, id: &Uuid) -> SharedResult<Option<Uuid>> {
+    pub fn link(&mut self, id: &Uuid) -> SharedResult<Option<Uuid>> {
         for link_id in self.owned_ids() {
             if let FileType::Link { target } = self.find(&link_id)?.file_type() {
-                if id == &target {
+                if id == &target && !self.calculate_deleted(&link_id)? {
```
tvanderstad, line 190: this call to calculate deleted necessitates a mut reference which necessitates slight re-arrangement of `in_pending_share` above

PR #1346, opened 2022-08-29
android, core: fix search crashes

### LINKED ISSUE #1293: searching crashes core
Searching from android for some folder names will sometimes cause core to crash.

SmailBarkouch:
Search was crashing on android. The problem ended up being with the content window being miss-computed, and thus the matched indices were false. Sometimes, the matched indices would be larger than the content window, thus giving an out of bounds error.

The crash happened in android, but it was caused by core.

fixes #1293

### DIFF
--- a/core/src/service/search_service.rs
+++ b/core/src/service/search_service.rs
@@ -372,7 +372,7 @@ impl RequestContext<'_, '_> {
                 let take_chars_len = if at_least_take > IDEAL_CONTENT_MATCH_LENGTH {
                     at_least_take
                 } else {
-                    IDEAL_CONTENT_MATCH_LENGTH - (last_match - index_offset)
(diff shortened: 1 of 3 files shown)


core/src/service/search_service.rs
```diff
@@ -372,7 +372,7 @@ impl RequestContext<'_, '_> {
                 let take_chars_len = if at_least_take > IDEAL_CONTENT_MATCH_LENGTH {
                     at_least_take
                 } else {
-                    IDEAL_CONTENT_MATCH_LENGTH - (last_match - index_offset)
+                    IDEAL_CONTENT_MATCH_LENGTH
```
Parth, line 375: you found a flaw in your logic, what you really found is a missing test case

Parth (changes requested):

Parth (approved):

PR #1360, opened 2022-09-02
egui: implement auto-sync

steverusso:
Implement very basic auto syncing functionality: every second, check if there's work and sync if so.


clients/egui/src/account/mod.rs
```diff
@@ -418,6 +437,31 @@ impl From<SyncUpdate> for AccountUpdate {
...
+    update_tx: &mpsc::Sender<AccountUpdate>, ctx: &egui::Context,
+) {
+    let settings = settings.clone();
+    let core = core.clone();
+    let update_tx = update_tx.clone();
+    let ctx = ctx.clone();
+
+    thread::spawn(move || loop {
+        if settings.read().unwrap().auto_sync {
+            match core.calculate_work() {
```
SmailBarkouch, line 451: I think calling calculate work every instant may not be good since it does a network call each time. I'm not too sure what's the best duration between calls, I know on android its 30 seconds, although it should maybe be something like 1 or 2 minutes.

SmailBarkouch (changes requested):

clients/egui/src/account/mod.rs
steverusso: True, I'll rfc this

SmailBarkouch (approved):

clients/egui/src/account/mod.rs
```diff
@@ -96,6 +100,9 @@ impl AccountScreen {
     fn process_updates(&mut self, ctx: &egui::Context, frame: &mut eframe::Frame) {
         while let Ok(update) = self.update_rx.try_recv() {
             match update {
+                AccountUpdate::BgUpdate(update) => match update {
+                    BgUpdate::AutoSync => self.perform_sync(ctx),
+                },
```
tvanderstad, line 105: This doesn't block the UI thread? This is beautiful :sob:

Parth (changes requested):
When we first discussed this PR I had 3 concerns, at least one of those has been addressed, but probably best to flesh them out:

1. load on the server by syncing and checking for work more often. Server is now like 200x faster than when this PR was first created, so probably fine now. Additionally you've convinced me the UX is just worthwhile.
2. there's a lot of overlap between what calculate work & sync do. It's generally worth it to just sync if you're calling the network variant of calculate work. There's a non network variant that on mobile we call more often for battery life reasons
3. I'm concerned that if you sync often and are receiving updates you're going to be overwriting them. Apple explicitly handles this case by doing the following: 
+ it always saves before syncing (easy enough)
+ it sees after a sync if the (single) open file has changed and resets the editor if it has (someone else modified a file or you modified it on another device

Maybe you could tell me about the state of 2 & 3 and we can plan from there

PR #1362, opened 2022-09-07
infra: release website via releaser

Parth:
we accumulated a lot of unreleased changes to our site, our existing script was from a time when we were still using `hugo` to generate our site. This replaces the script by adding the release logic to releaser, and the site is now up to date.


utils/releaser/src/main.rs
```diff
@@ -14,13 +17,24 @@ enum Releaser {
...
     match Releaser::from_args() {
         Releaser::DeployServer => server::deploy_server(),
         Releaser::ReleaseApple => apple::release_apple(&Github::env(), &AppStore::env()),
         Releaser::ReleaseAndroid => android::release_android(&Github::env(), &PlayStore::env()),
         Releaser::ReleaseWindows => windows::release(&Github::env()),
+        Releaser::ReleasePublicSite => public_site::release(),
     }
 }
+
+pub fn root() -> PathBuf {
```
steverusso, line 34: Kind of feels like a util function, but either way, I'd throw a comment next to the vague `root();` line in main.

Parth: yep you're right

Parth: I will fix & merge

PR #1368, opened 2022-09-09
admin: list premium users

### LINKED ISSUE #1367: list premium users using admin cli

SmailBarkouch:
fixes #1367


core/libs/shared/src/api.rs
```diff
@@ -442,5 +442,31 @@ impl Request for AdminDisappearFileRequest {
     const ROUTE: &'static str = "/admin-disappear-file";
 }
 
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone)]
+pub enum ShallowPaymentPlatform {
+    GooglePlay,
+    Stripe,
+}
```
Parth, line 449: very un-worthwhile imo, just return the `PaymentPlatform`

Parth (changes requested):

Parth (changes requested):
checkout #1372, once it's merged, rebase and migrate these changes into that

steverusso (changes requested):

Parth (changes requested):
>  also I think you're going to want to setup structure that allows us to list all users, and have arbitrary filters

Parth:
I'm still thinking about this, but maybe this is what we need:

A way to list users, and a way to get all the info we have on a particular user.

Listing users maybe we want to filter, this is the first filter we've come up with (premium) but likely we'll add more filters in the future, like root created after a specified date or whatever.

When we get the info for a user we should be able to specify a username or a public key, and it should print out all the billing info (and we can add other data points it prints out later) Data like which files they own, who they share files with, etc.

Parth:
In the future we can add filters as we need and data on a user's "profile" as we need. Maybe we make a lil egui app for this not sure.

Parth (approved):

clients/admin/src/main.rs
```diff
@@ -73,6 +99,58 @@ fn delete_account(core: &Core, username: String) -> Res<()> {
...
+fn account_info(core: &Core, username: Option<String>, public_key: Option<String>) -> Res<()> {
+    let identifier = if let Some(username) = username {
+        AccountIdentifier::Username(username)
+    } else if let Some(public_key) = public_key {
+        AccountIdentifier::PublicKey(PublicKey::parse_compressed(<&[u8; 33]>::try_from(
+            base64::decode(public_key)?.as_slice(),
+        )?)?)
+    } else {
+        println!("Please specify a username or public key.");
+        return Ok(());
```
steverusso, line 144: nit: this feels like an error to run account info without providing a username or public key

core/src/lib.rs
```diff
@@ -30,6 +32,8 @@ pub use crate::model::errors::*;
 pub use crate::service::import_export_service::{ImportExportFileInfo, ImportStatus};
 pub use crate::service::sync_service::{SyncProgress, WorkCalculated};
 pub use crate::service::usage_service::{bytes_to_human, UsageItemMetric, UsageMetrics};
+pub use base64;
+pub use libsecp256k1::PublicKey;
```
steverusso, line 36: Please toss these with the other third party pub uses (uuid above) and not with the crates.

core/src/lib.rs
```diff
@@ -12,7 +12,9 @@ mod external_interface;
 pub use uuid::Uuid;
 
 pub use lockbook_shared::account::Account;
-pub use lockbook_shared::api::{GooglePlayAccountState, StripeAccountTier, SubscriptionInfo};
+pub use lockbook_shared::api::{
+    AccountFilter, AccountIdentifier, GooglePlayAccountState, StripeAccountTier, SubscriptionInfo,
+};
 pub use lockbook_shared::api::{PaymentMethod, PaymentPlatform};
```
steverusso, line 18: Might as well combine these now

PR #1372, opened 2022-09-14
admin: isolate admin and add many commands

Parth:


clients/admin/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "admin"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+structopt = "0.3.13"
+lockbook-core = { path = "../../core" }
+dialoguer = { version = "0.10.1", features = ["fuzzy-select"] }
+backtrace = "0.3"
```
Parth, line 10:
```suggestion
backtrace = "0.3"

```

containers/local.env
```diff
@@ -1,5 +1,5 @@
 INDEX_DB_LOCATION=/tmp/lbdev
-FILES_PATH=/tmp/lbdev/files
+FILES_PATH=/tmp/lbdev/docs
```
Parth, line 2: becomes consistent with `prod.env` so that a tar'd backup of prod can be tested locally

clients/admin/src/error.rs
```diff
@@ -0,0 +1,12 @@
+use backtrace::Backtrace;
+use std::fmt::Debug;
+
+pub struct Error;
```
Parth, line 4: I think playing around with backtraces inside errors is going to be useful for the future

core/libs/shared/src/file_metadata.rs
```diff
@@ -1,4 +1,5 @@
 use std::fmt;
+use std::fmt::{Debug, Formatter};
```
steverusso, line 2:
```suggestion
use std::fmt::{self, Debug, Formatter};
```

steverusso (changes requested):
```
if maybe_confirm.unwrap_or(false) {
```

```
if let Some(confirm) = maybe_confirm {
        if confirm {
```

Consistently process confirmation result option.

steverusso (approved):

PR #1395, opened 2022-09-22
android: fixed android crashes from egui drawings

### LINKED ISSUE #1359: bug: drawings created in egui are causing problems on android
1. Simply creating a drawing in egui client and opening on android causes a blank gray screen (no black canvas as far as I can tell).
2. If you _swipe_ in any way on this gray screen and then press the back button, the android client crashes.

SmailBarkouch:
fixes #1359

### DIFF
--- a/core/libs/shared/src/drawing.rs
+++ b/core/libs/shared/src/drawing.rs
@@ -2,13 +2,19 @@ use std::collections::HashMap;
 
 use serde::{Deserialize, Serialize};
 
-#[derive(Serialize, Deserialize, Debug, Default)]
+#[derive(Serialize, Deserialize, Debug)]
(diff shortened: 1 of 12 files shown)


core/libs/shared/src/drawing.rs
```diff
@@ -11,6 +11,12 @@ pub struct Drawing {
     pub theme: Option<HashMap<ColorAlias, ColorRGB>>,
 }
 
+impl Default for Drawing {
```
Parth, line 14: @steverusso why did we do `new` for `Stroke`?

Parth (approved):
We want to be consistent between `default` / `new` in that file, but other than that lgtm

PR #1422, opened 2022-10-03
server: reduce build times by ~20s

tvanderstad:
reduce build times with _this one weird trick_! https://fasterthanli.me/articles/why-is-my-rust-build-so-slow

### DIFF
--- a/server/server/src/router_service.rs
+++ b/server/server/src/router_service.rs
@@ -111,6 +111,7 @@ macro_rules! core_req {
                 }
                 .instrument(span1)
             })
+            .boxed()
     }};
 }
 


Parth (changes requested):
this macro is already pretty funky, can you share some numbers to convince me this indirection is worthwhile? Maybe from macOS & Linux and debug & release builds.

I'm interested in enabling compiler features at release times that would give us blanket gains at the cost of compile time in our release builds, I feel like this is working in the other direction

tvanderstad:
```
| Platform        | Task                                      | Before | After |
|-----------------|-------------------------------------------|--------|-------|
| Macbook Pro x64 | `cargo check`                             | 11s    | 11s   |
| Macbook Pro x64 | `cargo build` (one line change)           | 52s    | 31s   |
| Macbook Pro x64 | `cargo build --release` (one line change) | 1m22s  | 1m3s  |
```

PR #1423, opened 2022-10-03
admin, server: improved `get_account_info`

SmailBarkouch:


clients/admin/src/account.rs
```diff
@@ -48,7 +51,21 @@ pub fn info(core: &Core, username: Option<String>, public_key: Option<String>) -
...
+        (SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap()
+            .as_millis() as u64
+            - account_info.last_active) as i64,
+    )
+    .format_human()
+    .to_string();
+    println!("last_active: {}", last_active);
+    println!("usage: {}B", account_info.usage);
```
tvanderstad, line 68: Thoughts about using our human readable bytes function here? The number would be less precise but maybe we'd want "10.1MB" instead of "1014349B"

clients/admin/src/account.rs
```diff
@@ -48,7 +51,21 @@ pub fn info(core: &Core, username: Option<String>, public_key: Option<String>) -
...
+    println!("root: {}", account_info.root);
+    println!("payment_platform: {:#?}", account_info.payment_platform);
+
+    let last_active = Duration::milliseconds(
+        (SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap()
+            .as_millis() as u64
+            - account_info.last_active) as i64,
+    )
```
Parth, line 64: + I think we just want to `{:#?}` this
+ I don't think we want to say how long ago they were active, I think we want to print the date / time they were last active. I think if this object contained an `Instant` or something the debug print would handle that too.
+ if you need to give anything special attention, consider impl'ing `Debug` the way we do for `Trial`.

server/server/src/account_service.rs
```diff
@@ -279,8 +280,37 @@ pub async fn admin_get_account_info(
...
+    let mut last_active = 0;
+
+    if let Some(ids) = db.owned_files.get(&owner)? {
+        for id in ids {
+            let metadata = db.metas.get(&id)?.ok_or_else(|| {
+                internal!("Metadata for id should exist, id: {:?}, owner: {:?}", id, owner)
+            })?;
+
+            if metadata.version > last_active {
+                last_active = metadata.version
```
Parth, line 301: This won't capture modifications to shared files, I think we actually can't capture this field without storing data on the server, so I'm down to remove `last_active` from this entirely for now.

Parth (changes requested):

Parth (approved):

PR #1426, opened 2022-10-04
fuzzer: add multiple users and sharing ops

tvanderstad:
I ran 100k trials locally without errors. ~TPS drops by about half for the initial trials, which are dominated by `SyncAndCheck`, which is now doing double the syncing and checking.~ I modified the order in which trials are run so that the initial TPS more closely matches the steady state TPS.

fixes https://github.com/lockbook/lockbook/issues/1398

### DIFF
--- a/core/libs/shared/src/core_ops.rs
+++ b/core/libs/shared/src/core_ops.rs
@@ -5,12 +5,12 @@ use libsecp256k1::PublicKey;
 use tracing::debug;
 use uuid::Uuid;
 
-use crate::access_info::UserAccessInfo;
+use crate::access_info::{UserAccessInfo, UserAccessMode};
 use crate::account::Account;
 use crate::core_config::Config;
 use crate::crypto::{DecryptedDocument, EncryptedDocument};
 use crate::document_repo::RepoSource;
-use crate::file::File;
+use crate::file::{File, Share, ShareMode};
 use crate::file_like::FileLike;
 use crate::file_metadata::{FileMetadata, FileType, Owner};
 use crate::filename::{DocumentType, NameComponents};
@@ -30,13 +30,31 @@ where
     Local: Stagable<F = Base::F>,
 {
     pub fn finalize(&mut self, id: &Uuid, account: &Account) -> SharedResult<File> {
-        let meta = self.find(id)?;
+        let meta = self.find(id)?.clone();
         let file_type = meta.file_type();
         let parent = *meta.parent();
         let last_modified = meta.timestamped_value.timestamp as u64;
         let name = self.name(id, account)?;
         let id = *id;
         let last_modified_by = account.username.clone();
+
+        // todo: this is a hack to get fuzzer running on sharing
+        let mut shares = Vec::new();
+        for user_access_key in meta.user_access_keys() {
+            if user_access_key.encrypted_by == user_access_key.encrypted_for {
+                continue
+            }
+            let mode = match user_access_key.mode {
+                UserAccessMode::Read => ShareMode::Read,
+                UserAccessMode::Write => ShareMode::Write,
+                UserAccessMode::Owner => { continue },
+            };
+            shares.push(Share{
+                mode,
+                shared_by: if user_access_key.encrypted_by == account.public_key() { account.username.clone() } else { String::from("<unknown>") },
+                shared_with: if user_access_key.encrypted_for == account.public_key() { account.username.clone() } else { String::from("<unknown>") }
+            });
+        }
 
         Ok(File {
             id,
@@ -45,7 +63,7 @@ where
             file_type,
             last_modified,
             last_modified_by,
-            shares: Vec::new(), // todo
+            shares,
         })
     }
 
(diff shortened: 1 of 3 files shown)


core/tests/exhaustive_sync/trial.rs
```diff
@@ -110,117 +117,145 @@ impl Trial {
...
-                    let doc = find_by_name(db, &doc_name).id;
+                MoveFile { user_index, device_index, doc_name: non_folder_name, destination_name } => {
+                    let db = &self.devices_by_user[user_index][device_index];
+                    let non_folder = find_by_name(db, &non_folder_name).id;
                     let dest = find_by_name(db, &destination_name).id;
 
-                    if let Err(err) = db.move_file(doc, dest) {
+                    let move_file_result = db.move_file(non_folder, dest);
+                    match move_file_result {
+                        Ok(()) | Err(UiError(MoveFileError::LinkInSharedFolder)) | Err(UiError(MoveFileError::FolderMovedIntoItself)) => {}
```
tvanderstad, line 167: Because core resolves links, clients never know if what they're moving is a link, and therefore fuzzer won't know whether `LinkInSharedFolder` is acceptable. I just went ahead and combined `MoveDocument` and `MoveFolder` into this op assuming `FolderMovedIntoSelf` won't come up when moving a document (and `LinkInSharedFolder` won't come up when moving a non-link)

core/libs/shared/src/core_ops.rs
```diff
@@ -30,22 +30,40 @@ where
...
+                UserAccessMode::Read => ShareMode::Read,
+                UserAccessMode::Write => ShareMode::Write,
+                UserAccessMode::Owner => { continue },
+            };
+            shares.push(Share{
+                mode,
+                shared_by: if user_access_key.encrypted_by == account.public_key() { account.username.clone() } else { String::from("<unknown>") },
+                shared_with: if user_access_key.encrypted_for == account.public_key() { account.username.clone() } else { String::from("<unknown>") }
+            });
+        }
```
tvanderstad, line 57: pretty reasonable implementation for now

core/tests/exhaustive_sync/trial.rs
```diff
@@ -251,84 +287,126 @@ impl Trial {
...
+                    for file in all_files.clone() {
+                        if file.id != file.parent {
+                            mutants.push(self.create_mutation(MoveFile {
+                                user_index,
+                                device_index,
+                                doc_name: file.name.clone(),
+                                destination_name: parent_name.clone(),
+                            }))
+                        }
+                    }
```
tvanderstad, line 378: This replaces the separate branches for moving documents or folders into this folder. The previous folder logic felt a little backwards to me (attempting to move this folder into another folder, whereas with documents it was written as attempting to move a document into this folder) but I'm _pretty sure_ I got this right

core/tests/exhaustive_sync/trial.rs
```diff
@@ -413,8 +494,8 @@ impl Default for Trial {
     fn default() -> Self {
         Self {
             id: Uuid::new_v4(),
-            clients: vec![],
-            target_clients: 2,
+            devices_by_user: vec![],
+            target_devices_by_user: vec![2, 2],
```
tvanderstad, line 498: user 0 has devices 0 and 1; user 1 has devices 0 and 1

core/tests/exhaustive_sync/experiment.rs
```diff
@@ -88,7 +88,8 @@ impl Experiment {
                 .spawn(move || loop {
                     match Self::grab_ready_trial_for_thread(thread_id, thread_state.clone()) {
                         (Some(mut work), _) => {
-                            let mutants = work.execute(thread_id);
+                            let mut mutants = work.execute(thread_id);
+                            mutants.reverse();
```
tvanderstad, line 92: Because mutants are pushed and popped, the last-pushed mutant was being run first. That means the first trials start with many SyncAndChecks (last mutation added by get_mutants), which are the most expensive and least useful operation for rapid feedback on fuzzer changes

core/tests/exhaustive_sync/trial.rs
tvanderstad: for some real fun we could set this to `vec![1, 2, 3]`

core/libs/shared/src/core_ops.rs
```diff
@@ -30,23 +30,41 @@ where
...
-            parent,
-            name,
-            file_type,
-            last_modified,
-            last_modified_by,
-            shares: Vec::new(), // todo
-        })
+        // todo: this is a hack to get fuzzer running on sharing
+        let mut shares = Vec::new();
+        for user_access_key in meta.user_access_keys() {
```
Parth, line 43: Can't you just clone here?

Parth: The hack being right now we're not caching keys, so most shares are going to show up as unknown?

tvanderstad: I think one of the above statements moves some portion of `meta`, but I forget. Pretty sure impact of cloning here vs there is going to be negligible tho

tvanderstad: Yeah that is the hack

Parth (approved):

Issue #1405, opened 2022-09-25
shares not populated


Parth:
(during finalization of files)

tvanderstad:
work required to resolve this:
* add a public key -> username map and a username -> public key map as hmdb tables in core
* when you initiate a share, there is code to fetch the public key for that user; this code should be modified to add that usrename/key pair to both maps and to first check if the username -> public key map contains the username before looking it up
* server needs a GetUsername endpoint which does the reverse lookup that GetPublicKey does. server will not need a new hmdb table for this; it can use the accounts table to look up the account and the username is stored in the account
* at the end of a sync, core should go through all the pending shares, and if there's a sharer public key that isn't in the local maps yet, it should look up the username using the new server endpoint and store the result in both maps
* when finalizing files, look up the public keys of shares in the file user access keys and put the usernames in the share field

tvanderstad:
on a related note, the 'last modified by' field is hardcoded as the current user during finalization of files, and should check the public key of the file signature and look it up in the maps added as part of this effort for the username of the last user to modify the file

PR #1433, opened 2022-10-08
core/server: finalize shares using public key cache

tvanderstad:
~Note: there isn't a migration required because the cache is populated on the next sync and shares come up as `<unknown>` in the meantime. If we required a migration, it would require network access, which would be force a user either to not use or to downgrade the app if they first open the new version while offline, or leave the cache unpopulated until the next sync.~ There *is* a migration, just because any hmdb schema requires a migration. The migration just copies the old tables over.

closes: #1405


core/src/repo/schema.rs
```diff
@@ -15,6 +15,7 @@ hmdb::schema! {
         root: <OneKey, Uuid>,
         local_metadata: <Uuid, SignedFile>,
         base_metadata: <Uuid, SignedFile>,
-        usernames: <String, Owner>
```
tvanderstad, line 18: this was unused

Parth (approved):

PR #1455, opened 2022-12-05
apple, core, server: app store billing

SmailBarkouch:
Apple users can now purchase premium via App Store billing. This PR is split into two parts:
- Front-end UI work on the apple clients. The majority of this UI is found in the `clients/apple/Shared` folder.
- Back-end UI work on core and server. This includes validations when first upgrading to premium, and processing app store server notifications.

### New Server Endpoints
- `/upgrade-account-app-store`: Accepts the original transaction id (first transaction id), app account token (used to link a lockbook account with an App Store account), and an encoded receipt. The encoded receipt is sent to apple to for validation, and the returned details will be checked up against the app account token and original transaction id.
- `/app_store_notification_webhook`: Processes app store server notifications.

### Code Organization
- `server/.../billing/app_store_client.rs`: Builds the requests that are made to apple's servers. This includes creating a JWS for authentication.
- `server/.../billing/app_store_service.rs`: Helper functions for `upgrade_account_app_store` and `app_store_notification_webhook`.
- `server/.../billing/app_store_model.rs`: The data types that are needed for parsing the responses and request bodys' of apple requests.

### Progress
- [x] Create iOS UI
- [ ] Create MacOS UI
- [x] Implement `upgrade_account_app_store` endpoint
- [x] Implement `app_store_notifications_webhook` endpoint
- [x] Test billing for iOS
- [ ] Test billing for MacOS
- [ ] Test edge case billing operations on app store server notifications
- [ ] screenshots in docs


clients/apple/Shared/Stateful Logic/BillingService.swift
```diff
@@ -0,0 +1,197 @@
...
+    var makingPurchaseAttempt = false
+    
+    
+    init(_ core: LockbookApi) {
+        self.core = core
+        updateListenerTask = listenForTransactions()
+
+        Task {
+            await requestProducts()
+        }
```
Parth, line 42: I wonder if we don't want to do this every time the app opens

clients/apple/Shared/Stateful Logic/BillingService.swift
```diff
@@ -0,0 +1,197 @@
...
+            if storeProducts.count == 1 {
+                maybeMonthlySubscription = storeProducts[0]
+            } else {
+                print("No products!")
+            }
+            
+        } catch {
+            print("Failed product request from the App Store server: \(error)")
+        }
+    }
```
Parth, line 172: I wonder if this fails on startup, will something later on act unexpectedly

clients/apple/Shared/Stateful Logic/BillingService.swift
```diff
@@ -0,0 +1,197 @@
...
+                }
+            }
+        }
+    }
+
+    
+    func purchasePremium() async throws -> PurchaseResult? {
+        let accountToken = UUID()
+        let purchaseOpt: Set<Product.PurchaseOption> = [Product.PurchaseOption.appAccountToken(accountToken)]
+
```
Parth, line 89: could await requestProducts right here right?

Parth: instead of `init`

clients/apple/Shared/Stateful Logic/BillingService.swift
```diff
@@ -0,0 +1,197 @@
...
+            guard let monthlySubscription = maybeMonthlySubscription else {
+                throw StoreError.noProduct
+            }
+            let result = try await monthlySubscription.purchase(options: purchaseOpt)
+            SKReceiptRefreshRequest().start()
+            
+            switch result {
+            case .success(let verification):
+                while true {
+                    if let receipt = getReceipt() {
```
Parth, line 100: Can there be no receipt? Wouldn't this inf loop if no receipt?

clients/apple/Shared/Stateful Logic/BillingService.swift
```diff
@@ -0,0 +1,197 @@
...
+                        makingPurchaseAttempt = true
+                        let transaction = try checkVerified(verification)
```
Parth, line 102: what happens if checkVerified throws?

clients/apple/Shared/Stateful Logic/BillingService.swift
```diff
@@ -0,0 +1,197 @@
...
+            print("Couldn't read receipt data with error: " + error.localizedDescription)
+        }
+
+        return .failure
+    }
+    
+    func checkVerified<T>(_ result: VerificationResult<T>) throws -> T {
+        switch result {
+        case .unverified:
+            throw StoreError.failedVerification
```
Parth, line 138: Why does this need to throw?

clients/apple/Shared/Stateful Logic/BillingService.swift
```diff
@@ -0,0 +1,197 @@
...
+                while true {
+                    if let receipt = getReceipt() {
+                        makingPurchaseAttempt = true
+                        let transaction = try checkVerified(verification)
+                        
+                        let result = self.core.upgradeAccountAppStore(originalTransactionId: String(transaction.originalID), appAccountToken: accountToken.uuidString.lowercased(), encodedReceipt: receipt)
+                        
+                        switch result {
+                        case .success(_):
+                            await transaction.finish()
```
Parth, line 108: pretty nested at this point, I think there's probably a way to write this all to be more readable

clients/apple/Shared/Views/ManageSubscription.swift
```diff
@@ -0,0 +1,118 @@
...
+            }
+            .padding(.vertical)
+                
+            switch settings.tier {
+            case .Trial: trial
+            case .Premium: trial
+            case .Unknown: trial
+            }
+            
+            Text("Expand your storage to **30GB** for just **\(billing.maybeMonthlySubscription!.displayPrice)** a month.")
```
Parth, line 27:
```suggestion
            Text("Expand your storage to **30GB** for **\(billing.maybeMonthlySubscription!.displayPrice)** a month.")
```

What language are we using elsewhere?

clients/apple/Shared/Views/ManageSubscription.swift
```diff
@@ -0,0 +1,118 @@
+import SwiftUI
```
Parth, line 1: Will re-review this after screenshots

clients/apple/Shared/Views/ManageSubscription.swift
```diff
@@ -0,0 +1,118 @@
...
+import AlertToast
+
+struct ManageSubscription: View {    
+    @EnvironmentObject var billing: BillingService
+    @EnvironmentObject var settings: SettingsService
+    
+    @Environment(\.presentationMode) var presentationMode
+        
+    @State var isLoading = false
```
Parth, line 10: prob can be more specific here

clients/apple/Shared/Views/ManageSubscription.swift
```diff
@@ -0,0 +1,118 @@
...
+    
+    @ViewBuilder
+    var trial: some View {
+        VStack(alignment: .leading) {
+            Text("If you upgraded, your usage would be:")
+            ColorProgressBar(value: settings.premiumProgress)
+        }
+    }
+}
+
```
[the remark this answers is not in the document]
Parth: Loading, errors, etc

containers/local.env
```diff
@@ -19,6 +19,12 @@ MILLIS_BETWEEN_METRICS=100
 
 ADMINS="admin1"
 
+APPLE_SUBSCRIPTION_PRODUCT_ID="basic.premium"
+APPLE_IAP_KEY=${APPLE_IAP_KEY:=1234}
+APPLE_IAP_KEY_ID=${APPLE_IAP_KEY_ID:=1234}
+APPLE_ISSUER_ID=${APPLE_ISSUER_ID:=1234}
+APP_STORE_CONNECT_SHARED_SECRET=${APP_STORE_CONNECT_SHARED_SECRET:=sharedsecret}
```
Parth, line 26: make this start with apple too? 

These can prob be shorter

core/libs/shared/src/api.rs
```diff
@@ -389,6 +389,32 @@ impl Request for UpgradeAccountGooglePlayRequest {
     const ROUTE: &'static str = "/upgrade-account-google-play";
 }
 
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone)]
+pub struct UpgradeAccountAppStoreRequest {
+    pub original_transaction_id: String,
```
Parth, line 394: what is this "original" business?

server/server/src/billing/app_store_client.rs
```diff
@@ -0,0 +1,92 @@
...
+use crate::config::AppleConfig;
+use jsonwebtoken::{encode, Algorithm, EncodingKey, Header};
+use lockbook_shared::clock::get_time;
+use reqwest::RequestBuilder;
+use serde::{Deserialize, Serialize};
+
+const VERIFY_PROD: &str = "https://buy.itunes.apple.com/verifyReceipt";
+const VERIFY_SANDBOX: &str = "https://sandbox.itunes.apple.com/verifyReceipt";
+
+const ALG: &str = "ES256";
```
Parth, line 11: can this change?

server/server/src/billing/app_store_client.rs
```diff
@@ -0,0 +1,92 @@
...
+const TYP: &str = "JWT";
+const AUDIENCE: &str = "appstoreconnect-v1";
+const BUNDLE_ID: &str = "app.lockbook";
+
+#[derive(Debug)]
+pub enum AppStoreError {
+    Other(String),
+}
```
Parth, line 19: seems weird

server/server/src/billing/app_store_client.rs
```diff
@@ -0,0 +1,92 @@
...
+            let resp_body: VerifyReceiptResponse = resp.json().await?;
+
+            if resp_body.status == 0 {
+                Ok(resp_body)
+            } else {
+                Err(AppStoreError::Other(format!("Unexpected response: {:?}", resp_body)))
+            }
+        }
+        _ => Err(AppStoreError::Other(format!("Unexpected response: {:?}", resp_body))),
+    }
```
Parth, line 91: should never happen in prod right?

server/server/src/schema.rs
```diff
@@ -45,7 +45,8 @@ pub mod v2 {
             file_children: <Uuid, HashSet<Uuid>>,
             sizes: <Uuid, u64>,
             google_play_ids: <String, Owner>,
-            stripe_ids: <String, Owner>
+            stripe_ids: <String, Owner>,
+            app_store_ids: <String, Owner>
```
Parth, line 49: I think we'll need a migration

server/server/src/router_service.rs
```diff
@@ -268,6 +269,47 @@ pub fn google_play_notification_webhooks(
...
+    server_state: &Arc<ServerState>,
+) -> impl Filter<Extract = impl warp::Reply, Error = warp::Rejection> + Clone {
+    let cloned_state = server_state.clone();
+
+    warp::post()
+        .and(warp::path(APP_STORE_WEBHOOK_ROUTE))
+        .and(warp::any().map(move || cloned_state.clone()))
+        .and(warp::body::bytes())
+        .then(|state: Arc<ServerState>, body: Bytes| async move {
+            let span = span!(
```
Parth, line 283: Can you make sure all the webhooks are getting timed? Just once over and compare to the main `core_req` and make sure we're doing everything the same way, logging & timing. Make sure we're creating spans the same way and that we're timing the same way.

server/server/src/config.rs
```diff
@@ -213,6 +214,36 @@ impl BillingConfig {
             ),
             google: GoogleConfig::from_env_vars(),
             stripe: StripeConfig::from_env_vars(),
+            apple: AppleConfig::from_env_vars(),
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct AppleConfig {
```
Parth, line 223: Feels like we're out growing env vars, maybe we want to start using toml for config in the future.

server/server/src/config.rs
```diff
@@ -213,6 +214,36 @@ impl BillingConfig {
...
+        let apple_root_cert_dest = env_or_empty("APPLE_ROOT_CERT_PATH");
+
+        Self {
+            iap_key: env_or_panic("APPLE_IAP_KEY"),
+            iap_key_id: env_or_panic("APPLE_IAP_KEY_ID"),
+            asc_public_key: env_or_panic("APPLE_APP_STORE_CONNECT_PUBLIC_KEY"),
+            issuer_id: env_or_panic("APPLE_ISSUER_ID"),
+            subscription_product_id: env_or_panic("APPLE_SUBSCRIPTION_PRODUCT_ID"),
+            asc_shared_secret: env_or_panic("APP_STORE_CONNECT_SHARED_SECRET"),
+            apple_root_cert: apple_root_cert_dest
```
Parth, line 244: how large is this? Can't just stuff it in an env var right?

Parth: Oh it's non-utf-8 or something right?

Parth: We could base64 encode it or something, want to avoid random files we have to keep track of.

server/server/src/billing/billing_service.rs
```diff
@@ -73,6 +79,60 @@ fn release_subscription_profile<T: Debug>(
     })?)
 }
 
+pub async fn upgrade_account_app_store(
```
Parth, line 82: ideal top level function, very readable

server/server/src/billing/app_store_service.rs
```diff
@@ -0,0 +1,131 @@
...
+                        .subject_pki,
+                ))
+                .map_err(|_| ClientError(AppStoreNotificationError::InvalidJWS))?;
+        }
+    }
+
+    let pem = format!(
+        "-----BEGIN PUBLIC KEY-----\n{}\n-----END PUBLIC KEY-----",
+        base64::encode(certs[0].public_key().raw)
+    );
```
Parth, line 118: Are you sure this is required?

server/server/src/billing/app_store_service.rs
```diff
@@ -0,0 +1,131 @@
...
+    let key = DecodingKey::from_ec_pem(pem.as_bytes())?;
```
Parth, line 119: Since this wasn't the main issue, couldn't we use the non-pem one and not turn our `der` into `pem` ourselves?

Parth (changes requested):

clients/apple/Shared/Stateful Logic/BillingService.swift
SmailBarkouch: It can be moved, but in the storekit example apple provides, it is called similarly. It shouldn't be a problem to move it though.

clients/apple/Shared/Stateful Logic/BillingService.swift
SmailBarkouch: It won't act unexpectedly later. The `maybeMonthlySubscription` is a nullable `Product`. If it is null at the `purchaseProduct` method, then I will simply say "Failed to complete purchase."

clients/apple/Shared/Stateful Logic/BillingService.swift
SmailBarkouch: The receipt should [be generated after a purchase](https://developer.apple.com/documentation/appstorereceipts/validating_receipts_with_the_app_store). It's just a matter of waiting.

core/libs/shared/src/api.rs
SmailBarkouch: Original means the first transaction. That is a way you can identify subscriptions.

server/server/src/billing/app_store_client.rs
SmailBarkouch: It cannot, it is a part of [apple's standards](https://developer.apple.com/documentation/appstoreserverapi/generating_tokens_for_api_requests).

clients/apple/Shared/Stateful Logic/BillingService.swift
SmailBarkouch: It will be caught by the `catch` on the outside of the loop.

clients/apple/Shared/Stateful Logic/BillingService.swift
SmailBarkouch: Doesn't have to but if I did return `nil` I would have to add more lines of code.

clients/apple/SwiftLockbookCore/.swiftpm/xcode/xcshareddata/xcschemes/SwiftLockbookCore.xcscheme
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Scheme
-   LastUpgradeVersion = "1240"
+   LastUpgradeVersion = "1400"
```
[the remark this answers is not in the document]
SmailBarkouch: Seems to be a value automatically modified by xcode.

clients/apple/SwiftLockbookCore/Sources/SwiftLockbookCore/Models/FileUsage.swift
```diff
@@ -4,16 +4,32 @@ public struct UsageMetrics: Codable {
...
 }
 
 public struct FileUsage: Codable {
     public var fileId: UUID
     public var sizeBytes: UInt64
+    
+    public init(fileId: UUID, sizeBytes: UInt64) {
+        self.fileId = fileId
+        self.sizeBytes = sizeBytes
+    }
```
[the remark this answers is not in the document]
SmailBarkouch: Seems `Codable` does not let us use the free constructor all `structs` have, so one is manually made. I did not make this addition by the way, it is a relic of the original PR.

server/server/src/billing/app_store_service.rs
SmailBarkouch: I tried to use der and it did not work as expected, likely a little thing I am missing.

clients/apple/Shared/Stateful Logic/BillingService.swift
Parth:
```
    func getReceipt() -> String? {
        if let appStoreReceiptURL = Bundle.main.appStoreReceiptURL,
            FileManager.default.fileExists(atPath: appStoreReceiptURL.path) {

            do {
                return try Data(contentsOf: appStoreReceiptURL, options: .alwaysMapped).base64EncodedString()
            }
            catch {
                return nil
            }
        }
        
        return nil
    }
    ```
    
I think waiting for a file to exist in a loop is fine, you should add a bit of a sleep regardless if you're going to do this. Can you link to a sample that does this? Feels strange regardless. But you should have a timeout, if some assumption of yours is wrong this will infinite loop.

1. handle the errors in the catch better
2. sleep and timeout

File: clients/apple/Shared/Stateful Logic/BillingService.swift [hunk 26]
```diff
@@ -0,0 +1,198 @@
...
+                                }
+                            }
+                        } else {
+                            await transaction.finish()
+                        }
+                    case .failure(let error):
+                        DI.errors.handleError(error)
+                    }
+                } catch {
+                    print("Transaction failed verification")
```
Parth, line 79: This is never going to make it to the user, we have a way to handle unexpected errors

PR #1457, opened 2022-12-08
core: sync returns work calculated

tvanderstad:


core/tests/sync_service_unsynced_client_tests.rs
```diff
@@ -78,7 +78,7 @@ fn delete() {
     core.delete_file(doc.id).unwrap();
     assert::all_paths(&core, &["/"]);
     assert::all_document_contents(&core, &[]);
-    assert::local_work_paths(&core, &["/document"]);
+    assert::local_work_paths(&core, &[]);
```
tvanderstad, line 81: new deleted documents, which already weren't synced, are now not considered work

core/libs/test_utils/src/assert.rs
```diff
@@ -206,7 +208,22 @@ pub fn server_work_paths(db: &Core, expected_paths: &[&'static str]) {
         .transaction(|tx| {
             let context = db.context(tx).unwrap();
             let account = context.tx.account.get(&OneKey {}).unwrap();
-            let remote_changes = context.get_updates().unwrap().file_metadata;
+            let remote_changes = context
```
tvanderstad, line 211: `get_updates` is now implemented on a different struct

core/libs/shared/src/staged.rs
```diff
@@ -125,6 +125,10 @@ where
             self.base.maybe_find(&id).cloned()
         }
     }
+
+    fn clear(&mut self) {
```
tvanderstad, line 129: can now stage the change of clearing a tree

core/libs/shared/src/core_ops.rs
```diff
@@ -95,7 +93,23 @@ where
             if self.link(&id)?.is_some() {
                 continue;
             }
+            user_visible_ids.push(id);
+        }
+        self.resolve_and_finalize_all(account, user_visible_ids.into_iter(), public_key_cache)
+    }
+
+    pub fn resolve_and_finalize_all<I, PublicKeyCache>(
```
tvanderstad, line 101: when resolving files for the purposes of reporting file names of calculated work, we don't want to skip e.g. deleted files

core/tests/sharing_tests.rs
```diff
@@ -299,9 +299,9 @@ fn write_document_in_rejected_shared_folder_in_share_folder() {
 
     assert_eq!(cores[1].read_document(document.id).unwrap(), b"document content by sharee");
     cores[1].sync(None).unwrap();
-    cores[1].get_file_by_id(document.id).unwrap_err();
+    cores[1].get_file_by_id(document.id).unwrap();
     cores[0].sync(None).unwrap();
-    assert_eq!(cores[0].read_document(document.id).unwrap(), b"document content by sharer");
+    assert_eq!(cores[0].read_document(document.id).unwrap(), b"document content by sharee");
```
tvanderstad, line 304: this test just had the wrong behavior before; the sharee should still have access to the file because it's in two shared folders and the pending share was only deleted for one of them

core/src/service/file_service.rs
```diff
@@ -154,6 +155,9 @@ impl<Client: Requester> RequestContext<'_, '_, Client> {
         if tree.calculate_deleted(id)? {
             return Err(CoreError::FileNonexistent);
         }
+        if tree.access_mode(Owner(account.public_key()), id)? < Some(UserAccessMode::Read) {
+            return Err(CoreError::FileNonexistent);
+        }
```
tvanderstad, line 160: don't finalize files you don't have access to e.g. files in deleted pending shares

core/src/service/sync_service.rs
```diff
@@ -24,122 +28,228 @@ pub struct WorkCalculated {
...
+        sync_context.local.to_lazy().promote();
+
+        Ok(WorkCalculated { work_units, most_recent_update_from_server: update_as_of as u64 })
+    }
+}
+
+struct SyncContext<'a, 'b, Base, Local, UsernameByPublicKey, PublicKeyByUsername, Client>
+where
+    Base: TreeLike<F = SignedFile>,
+    Local: TreeLike<F = SignedFile>,
```
tvanderstad, line 184: note: sync does not actually modify the trees, but stages changes so that they can be promoted

core/src/service/sync_service.rs
```diff
@@ -24,122 +28,228 @@ pub struct WorkCalculated {
...
+        }
+        SyncOperation::PullMetadataStart
+        | SyncOperation::PushMetadataEnd
+        | SyncOperation::PullDocumentStart(_)
+        | SyncOperation::PullDocumentEnd
+        | SyncOperation::PushDocumentStart(_)
+        | SyncOperation::PushDocumentEnd => {}
+    }
+    work_units
 }
```
tvanderstad, line 70: sync reports its SyncOperations; some of these become work units (for work calculated) and some become events for reporting sync progress

core/src/service/sync_service.rs
```diff
@@ -24,122 +28,251 @@ pub struct WorkCalculated {
...
+                SyncOperation::PullMetadataEnd(_)
+                | SyncOperation::PushMetadataEnd
+                | SyncOperation::PullDocumentEnd
+                | SyncOperation::PushDocumentEnd => {
+                    sync_progress.total += 1;
                 }
-            } else if local_change.document_hmac().is_some() {
-                num_doc_changes += 1;
-            }
+            })?;
```
tvanderstad, line 124: we calculate total sync progress using a dry run and handling operations by incrementing the total; this is _not_ the same as the length of work units in work calculated as it e.g. considers all pushed metadata as one unit

core/src/service/sync_service.rs
```diff
@@ -24,122 +28,251 @@ pub struct WorkCalculated {
...
+    fn sync<F>(&mut self, report_sync_operation: &mut F) -> CoreResult<i64>
+    where
+        F: FnMut(SyncOperation),
+    {
+        self.prune()?;
+        let update_as_of = self.pull(report_sync_operation)?;
+        self.last_synced = Some(update_as_of as u64);
+        self.push_metadata(report_sync_operation)?;
+        self.push_documents(report_sync_operation)?;
+        Ok(update_as_of)
```
tvanderstad, line 229: a couple of extraneous calls to prune and pull were removed now that server does not apply implicit file deletions or anything like that

core/src/service/sync_service.rs
```diff
@@ -24,122 +28,251 @@ pub struct WorkCalculated {
...
+            self.populate_public_key_cache(&remote_changes)?;
+
+            let mut remote = self.base.stage(remote_changes).to_lazy();
+            report_sync_operation(SyncOperation::PullMetadataEnd(
+                remote.resolve_and_finalize_all(
+                    &self.account,
+                    remote.tree.staged.owned_ids().into_iter(),
+                    self.username_by_public_key,
+                )?,
+            ));
```
tvanderstad, line 257: note: metadata is pulled dry run or not, and metadata is pulled twice during a sync (once for computing total work and once for actually doing the work)

core/src/service/sync_service.rs
```diff
@@ -24,122 +28,251 @@ pub struct WorkCalculated {
...
+            let (_, remote_changes) = remote.unstage();
+            (remote_changes, update_as_of)
+        };
 
         // initialize root if this is the first pull on this device
-        if self.tx.root.get(&OneKey {}).is_none() {
+        if self.root.is_none() {
```
tvanderstad, line 263: use local fields of new struct instead of actual tables so that changes can be read by later parts of sync and optionally applied

core/src/service/sync_service.rs
```diff
@@ -156,14 +289,26 @@ impl<Client: Requester> RequestContext<'_, '_, Client> {
...
+                            .client
+                            .request(account, GetDocRequest { id, hmac: remote_hmac })?
+                            .content;
+                        document_repo::insert(
+                            self.config,
+                            &id,
+                            Some(&remote_hmac),
+                            &remote_document,
+                        )?;
+                    }
```
tvanderstad: document reads and writes are avoided during dry run

core/src/service/sync_service.rs
```diff
@@ -24,122 +28,251 @@ pub struct WorkCalculated {
...
+
+struct SyncContext<'a, 'b, Base, Local, UsernameByPublicKey, PublicKeyByUsername, Client>
+where
+    Base: TreeLike<F = SignedFile>,
+    Local: TreeLike<F = SignedFile>,
+    UsernameByPublicKey: SchemaEvent<Owner, String>,
+    PublicKeyByUsername: SchemaEvent<String, Owner>,
+    Client: Requester,
+{
+    // when dry_run is true, sync skips document downloads/uploads and disk reads/writes
```
Parth, line 189:
```suggestion
    /// when dry_run is true, sync skips document downloads/uploads and disk reads/writes
```

Issue #1296, opened 2022-08-20
Metrics problems


Parth:
+ user activity currently disregards sharing, if I do a bunch of work in a shared folder that "activity" will contribute to the owner's folder's active state.
+ we should probably measure how many people use sharing

Parth:
Our metrics are defined and calculated in https://github.com/lockbook/lockbook/blob/master/server/server/src/metrics.rs. This is an example of logic that we used to try to write in sql and we're much happier to have written it in Rust.

Checkout how the various fields are calculated, explore how trees are created and what questions we can ask a tree. 

One of our most important metrics is "active users", and we want to be more strict about how we're counting and calculating active users. Currently there's 2 known problems with active users:
+ all members of a shared folder are being marked as active when anyone modifies something in a shared folder
+  new users are being marked as active even if they didn't author a file.

We want to be strict about this metric because I feel like it's the single number that captures how well we're doing, I'm sure we can convince a large number of people to try the app, but we want to see how many people are returning to the app and making it a part of their daily routine. Our second most important metric is probably premium users.

So we could fix the way active users are being calculated by doing the following:
+ Determining if _they_ modified the file in that file tree within the last 2 days. Which means modifying [this](https://github.com/lockbook/lockbook/blob/master/server/server/src/metrics.rs#L164-L166) to filter out changes by other people.
+ We could also filter out people if the only file that was changed in the last 2 days was their root, (see `.is_root()`).

This would be a 90% solution, try to take a moment to see if you can figure out under what circumstances this would not work well.

Answer:

If user A edits a file, they're active. If user B edits that same file 1 hour later, they "overwrite" the fact that user A was active.

So to answer this question properly we don't have all the data readily available, we need to store more information in our database.

I believe we're going to want to store an additional table that tracks which public_keys were _seen last_.

The schema for our server's database is defined [here](https://github.com/lockbook/lockbook/blob/master/server/server/src/schema.rs#L48). Take a moment to observe what's going on here. There's a bit of messiness introduced by hmdb that I (maybe you) will fix soon, but it's a good look into how things happen. So there's 2 schemas here, an older v1, and a v2. We added 2 "tables" to our database in v2. Similarly you're probably going to add a `last_seen: <Owner, u64>`. But you're not going to add it to `V2` you're going to define a `V3`. And then you're going to migrate data from `V2` -> `V3` there's an example of this in that file.

For your reference the messiness introduced by hmdb is the following:
+ `v2` needs to be in it's own mod
[...]

PR #1463, opened 2022-12-23
apple, server, core: apple billing

### LINKED ISSUE #984: apple pay server support

### LINKED ISSUE #985: use apple pay for payments

SmailBarkouch:
Apple users can now purchase premium via App Store billing. This PR is split into two parts:
- Front-end UI work on the apple clients. The majority of this UI is found in the `clients/apple/Shared` folder.
- Back-end UI work on core and server. This includes validations when first upgrading to premium, and processing app store server notifications.

### New Server Endpoints
- `/upgrade-account-app-store`: Accepts the original transaction id (first transaction id), app account token (used to link a lockbook account with an App Store account), and an encoded receipt. The encoded receipt is sent to apple to for validation, and the returned details will be checked up against the app account token and original transaction id.
- `/app_store_notification_webhook`: Processes app store server notifications.

### Code Organization
- `server/.../billing/app_store_client.rs`: Builds the requests that are made to apple's servers. This includes creating a JWS for authentication.
- `server/.../billing/app_store_service.rs`: Helper functions for `upgrade_account_app_store` and `app_store_notification_webhook`.
- `server/.../billing/app_store_model.rs`: The data types that are needed for parsing the responses and request bodys' of apple requests.

### Progress
- [x] Create iOS UI
- [x] Create MacOS UI
- [x] Implement `upgrade_account_app_store` endpoint
- [x] Implement `app_store_notifications_webhook` endpoint
- [x] Test billing for iOS
- [x] Test billing for MacOS
- [x] Test edge case billing operations on app store server notifications
- [x] screenshots in docs

fixes #984
fixes #985


clients/apple/Shared/Stateful Logic/BillingService.swift
```diff
@@ -0,0 +1,203 @@
...
+    
+    init(_ core: LockbookApi) {
+        self.core = core
+        
+        if case .success(_) = core.getAccount() {
+            Task {
+                await launchBillingBackgroundTasks()
+            }
+        }
+    }
```
Parth, line 44: What I'm thinking here is: I don't want the app to do this every time I open it. Ideally it only does this when I'm doing the billing flow. Costs some data, some battery, and maybe adds a hair of latency to startup (i know it's async). Can't we fire this off at some later point in the flow than every app startup?

clients/egui/src/widgets/subscription.rs
```diff
@@ -32,6 +32,7 @@ fn subscription_info(
         Some(info) => match &info.payment_platform {
             Stripe { card_last_4_digits } => draw_stripe(ui, card_last_4_digits),
             GooglePlay { account_state } => draw_google_play(ui, account_state),
+            AppStore { .. } => todo!(),
```
Parth, line 35: I think not showing this for now is fine, but `panic`'ing here is prob not the move, let's create a tracking issue for this and not panic.

server/server/Cargo.toml
```diff
@@ -38,6 +38,9 @@ tracing = "0.1.5"
 tracing-subscriber = "0.3.9"
 tracing-appender = "0.2"
 itertools = "0.10.1"
+reqwest = { version = "0.11", features = ["json"] }
+jsonwebtoken = "*"
```
Parth, line 42: why *?

server/server/src/billing/app_store_client.rs
```diff
@@ -0,0 +1,124 @@
...
+                            |err| {
+                                internal!(
+                                    "Cannot decode apple jwt payload: {:?}, err: {:?}",
+                                    part,
+                                    err
+                                )
+                            },
+                        )?)?;
+                    return Ok((last_trans.clone(), trans_info));
+                }
```
Parth, line 106: safe to ignore the rest of this data in the loop?

server/server/src/billing/app_store_client.rs
```diff
@@ -0,0 +1,124 @@
...
+            }
+
+            Err(internal!("No usable data returned from apple subscriptions statuses endpoint despite assumed match. resp_body: {:?}, monthly_sub_group: {}", sub_status, config.monthly_sub_group_id))
+        }
+        400 | 404 => {
+            let resp_body = resp.text().await.ok();
+            info!(
```
Parth, line 113: perhaps this is `error!` worthy?

server/server/src/billing/billing_service.rs
```diff
@@ -73,6 +79,67 @@ fn release_subscription_profile<T: Debug>(
...
+    debug!("Upgrading the account of a user through app store billing");
+
+    if let Some(owner) = server_state
+        .index_db
+        .app_store_ids
+        .get(&request.app_account_token)?
+    {
+        if let Some(other_account) = server_state.index_db.accounts.get(&owner)? {
+            if other_account.billing_info.billing_platform.is_some() {
+                return Err(ClientError(UpgradeAccountAppStoreError::AppStoreAccountAlreadyLinked));
```
Parth, line 102: wouldn't we return this error if they had `Strip`? Prob need to rename the error

server/server/src/billing/billing_service.rs
```diff
@@ -73,6 +79,67 @@ fn release_subscription_profile<T: Debug>(
...
+            }
+        }
+    }
+
+    let expires = verify_details(
```
Parth, line 107: I would refer to this as `app_store_service::verify_details` but nbd

Parth, line 107: Do we need to check this `expires` for validity?

server/server/src/schema.rs
```diff
@@ -45,7 +45,8 @@ pub mod v2 {
             file_children: <Uuid, HashSet<Uuid>>,
             sizes: <Uuid, u64>,
             google_play_ids: <String, Owner>,
-            stripe_ids: <String, Owner>
+            stripe_ids: <String, Owner>,
+            app_store_ids: <String, Owner>
```
Parth, line 49: This needs to end up in a v3 with a migration, see the rest of this file. Also see how this is gonna interact with the proposed schema changes [here](https://github.com/lockbook/lockbook/issues/1296#issuecomment-1368230780)

cc: @ad-tra 

I'm down to not do it in this PR and let @ad-tra just do it in his PR

server/server/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+extern crate core;
```
Parth, line 1: prob an accidental auto import

Parth (changes requested):

server/server/src/billing/app_store_client.rs
SmailBarkouch: Yes it is, we are only looking for a particular subscription. There should not be other things in this loop regardless though.

server/server/src/billing/billing_service.rs
SmailBarkouch: I ended up `if let`ing an app store subscription, so there is no need to rename. Also, it is what should be done since we only care if it is an app store subscription.

server/server/src/billing/billing_service.rs
SmailBarkouch: We do not since in `verify_details`, the subscription status endpoint we make a request to tells us if the subscription has expired or not.

server/server/src/billing/billing_service.rs
Parth: What if they have another platform like stripe?

SmailBarkouch: It doesn't matter if they have another platform so long as the platform is not apple. But on second thought, it is possible that the platform is apple but not the same apple account. I will add a check for the origin transaction id.

If the account happened to be stripe it would not matter because all this check is looking for is if an apple account is not being used by someone else. An apple account can only pay for one subscription.

Parth (approved):
Congrats dawg

PR #1464, opened 2022-12-23
android, core, server: android sharing

SmailBarkouch:
Add sharing capabilities to android using core.

## Files of Interest
- `clients/android/.../screen/CreateLinkFragment.kt`
- `clients/android/.../screen/ShareFileFragment.kt`
- `clients/android/.../screen/PendingSharesFragment.kt`

## Example Video

<details>

https://user-images.githubusercontent.com/20663038/190881424-f2127593-1ca6-47bd-8363-3d227b99ed80.mp4
</details>

**Note**: The pending shares screen does not list the owner of the shared files but lists who last edited the file. This is because there is no `owner` field on the `File` type yet.


clients/admin/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "admin"
-version = "0.5.6"
```
Parth, line 3: there's a handful of these unintentional downgrades of code, once over this whole PR

Parth (changes requested):
can we add the most primitive hard coded feature flag so only engineering team can create new shares, steve, travis, adam, you etc.

planning on lifting that requirement in feb or so

Issue #1239, opened 2022-07-26
core version should be stored in the header of the request


Parth:
We can't store the version inside requests as these can change depending on the nature of a breaking change. We want to seek a more stable place to store this information, such as http headers.

Parth:
Currently we send version information in the field of a struct:

This is what every request type is parameterized using. 

```rust
#[derive(Serialize, Deserialize, Debug, Clone)]
pub struct RequestWrapper<T: Request> {
    pub signed_request: ECSigned<T>,
    pub client_version: String,
}
```

So if you take:

`NewAccountRequest`:

```rust
#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
pub struct NewAccountRequest {
    pub username: Username,
    pub public_key: PublicKey,
    pub root_folder: SignedFile,
}
```
and plug substitute it in for `T` what you'd get is:

something like:

```rust
#[derive(Serialize, Deserialize, Debug, Clone)]
pub struct RequestWrapper<NewAccountRequest: Request> {
    pub signed_request: ECSigned<NewAccountRequest>,
    pub client_version: String,
}
```

wip

Parth:
Core and Server parse serialize and deserialize these into and from json.

The idea of the `client_version` field is to be able to evolve the protocol and force clients to upgrade if we need them to.

The point of this is often defeated however because if `NewAccountRequest` is changed significantly then older `RequestWrapper<NewAccountRequest>`s fail to parse as the newer `RequestWrapper<NewAccountRequest>` (the server just rejects these requests). This is fine internally we've come to recognize these error messages and know that they mean upgrade. But we should handle this situation better so external customers understand what's going on. (`UnexpectedError` vs. `ClientUpdateRequired `).

The two ways to handle this that come to mind are:

## approach 1:

Even if our ultimate goal is to parse a `Request<NewAccount>` I think we could parse `Request<Value>` first [see more](https://stackoverflow.com/questions/56279350/how-to-get-at-one-particular-item-in-json-file-using-serde-json-without-deriving). And then check the value of `client_version`, and then we can proceed with the rest of the request.

My concern here is roughly around performance, I'd want to make sure we're not incurring expensive parsing operations repeatedly. So how does the performance of:

```rust
let a: Req<NewAccount> = serde_json::parse_json(s);
check_version(a.client_version)?;
```
compare to
```rust
let a: Req<Value> = serde_json::parse_json(s);
check_version(a.client_version)?;
[...]

PR #1467, opened 2022-12-25
core, server: core version should be stored in the header of the request 

ad-tra:
## Approach 
Changes were made on two fronts:
1. In the routes handler located in `router_service.rs` >  `core_req!`, a warp header filter was added for an `Accept` header which stores the core version. 
2. the `Requester` in `api_service.rs` was changed to inject the core version as an `Accept` header. As far as I understand, most core requests sent to the server pass through  `fn request` in `impl Requester for Network`

### Tests
no tests were written. I'd love to have a discussion about the granularity of tests that we maintain. @Parth suggested a test for requests missing an `Accept` header, but I don't know if or why that's necessary. isn't the purpose of warp filters to validate the request? 

fixes: #1239

### DIFF
--- a/server/server/src/lib.rs
+++ b/server/server/src/lib.rs
@@ -56,7 +56,7 @@ macro_rules! internal {
     }};
 }
 
-pub fn verify_client_version<Req: Request>(
+pub fn handle_version<Req: Request>(
     version: &Option<String>,
 ) -> Result<(), ErrorWrapper<Req::Error>> {
     let versions = vec!["0.5.5", "0.5.6"];
--- a/server/server/src/lib.rs
+++ b/server/server/src/lib.rs
@@ -59,10 +59,14 @@ macro_rules! internal {
 pub fn verify_client_version<Req: Request>(
     version: &Option<String>,
 ) -> Result<(), ErrorWrapper<Req::Error>> {
+    let versions = vec!["0.5.5", "0.5.6"];
     match version.as_deref() {
-        Some("0.5.5") => Ok(()),
-        Some("0.5.6") => Ok(()),
-        None => Ok(()),
+        Some(x) if versions.contains(&x) => {
+            router_service::CORE_VERSION_COUNTER
+                .with_label_values(&[x])
+                .inc();
+            Ok(())
+        }
         _ => Err(ErrorWrapper::<Req::Error>::ClientUpdateRequired),
     }
 }
--- a/server/server/src/lib.rs
+++ b/server/server/src/lib.rs
@@ -56,6 +56,16 @@ macro_rules! internal {
     }};
 }
 
+pub fn verify_client_version_header<Req: Request>(
+    version: &String
+) -> Result<(), ErrorWrapper<Req::Error>> {
+    match version as &str {
+        "0.5.5" => Ok(()),
+        "0.5.6" => Ok(()),
+        _ => Err(ErrorWrapper::<Req::Error>::ClientUpdateRequired),
+    }
+}
+
 pub fn verify_client_version<Req: Request>(
     request: &RequestWrapper<Req>,
 ) -> Result<(), ErrorWrapper<Req::Error>> {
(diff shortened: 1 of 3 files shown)


Parth (changes requested):
This is good, but I don't think [accept](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept) makes sense. I didn't look too hard but it looks like there's a [specific schema](https://dotnettutorials.net/lesson/web-api-versioning-using-accept-header/) around `Accept`. I'm happy to just use a header called `core_version` so we don't have to fuss around with this.

I'm going back on my word a bit, but I thought about it a bit and for now, this header needs to be optional. There's a few ways to do this, including [getting all the headers](https://users.rust-lang.org/t/how-to-get-all-the-headers-in-warp/75142/2) and then checking if a particular one exists. Or you can try to: [map the header](https://github.com/seanmonstar/warp/issues/177) to some type and take in that type (prob `Option<String>`).

And finally, work with @SmailBarkouch and get some metrics in here

ad-tra:
> I don't think [accept](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept) makes sense.

@Parth Do you want to use `Accept-Version`, it's not standard but still [commonly used](https://cloud.google.com/blog/products/api-management/api-design-which-version-of-versioning-is-right-for-you)

> There's no standard request header that's ideal for the client to say what format version it wants. The standard "Accept" header specifies which media types the client can accept (e.g., json, yaml, xml, html, plain text), and the standard "Accept-Language" header denotes which natural languages the client can accept (e.g., French, English, Spanish). Some API designers (e.g., the authors of the Restify framework) use a non-standard header called "Accept-Version". If you're doing format versioning, I recommend this header.

If not, we cause use `Core-Version` uppercase, and kebab case is usually the convention for HTTP headers

ad-tra:
What's the line of reasoning behind making the version header optional?

How to handle the missing version header? Do we 
1. polyfill using the version specified in the request body to perform the version check. 
2. or ignore the version check altogether? 

I'm heading towards approach (2)

ad-tra:
@SmailBarkouch is a `HistogramVec` appropriate in this context. Ideally, it would be a pie chart.

Parth:
I'm on board with accept-version.

All I'm looking to do with making the header optional for now, so that people with older versions of core are given a message to upgrade that actually makes sense to them -- otherwise, they'll just get `Unexpected(BadRequest)` or something like that. So what you have now is fine, all you have to change is: in the `None` case, return `ClientUpdateRequired`. After we see, via metrics, that everyone has upgraded, we can elect to make this a required field if desired. 

So all you have to change is:

delete this line:

```rust
        None => Ok(()),
```

(and add some metrics)

SmailBarkouch:
> @SmailBarkouch is a `HistogramVec` appropriate in this context. Ideally, it would be a pie chart.

My knowledge of prometheus's histograms is somewhat limited, but you should be able to achieve a pie chart using a `CounterVec` (what you have right now) or a `GaugeVec`. These series of values can be expressed in Grafana to produce a pie chart pretty effortlessly.

server/server/src/lib.rs
```diff
@@ -57,11 +57,16 @@ macro_rules! internal {
 }
 
 pub fn verify_client_version<Req: Request>(
```
Parth, line 59:
```suggestion
pub fn handle_version<Req: Request>(
```

to suggest that this fn has a side effect

server/server/src/router_service.rs
```diff
@@ -25,6 +27,12 @@ lazy_static! {
         &["request"]
     )
     .unwrap();
+    pub static ref CORE_VERSION_COUNTER: CounterVec = register_counter_vec!(
+        "lockbook_server_core_version",
+        "Lockbook server's supplied core version",
```
Parth, line 32:
```suggestion
        "Core version request attempts",
```

Parth (changes requested):

Parth (changes requested):
a test that proves that current cores will be able to be asked to update (and will not get a `BadRequest`).

Parth (approved):

Issue #1262, opened 2022-08-12
CI should fail a PR if Cargo.lock changes


Parth:

Parth:
Our CI is defined [here](https://github.com/lockbook/lockbook/blob/master/utils/dev-tool/src/main.rs).

If you change any `Cargo.toml` there's a corresponding change to the single `Cargo.lock`. We want those changes to come from the same PR. Often someone will push a change to `Cargo.toml` and commit the corresponding change to `Cargo.lock`. This change then just shows up in an unrelated PR. Slightly more annoyingly some other tools don't handle this well either.

At the end of our CI run, if there is a change to the `Cargo.lock` file that was not a part of the original PR, we want to fail the build. We want this to be it's own step at the end: https://github.com/lockbook/lockbook/blob/master/.github/workflows/integration.yml#L42

PR #1468, opened 2022-12-26
CI should fail a PR if Cargo.lock changes 

ad-tra:
Fixes: #1262

I haven't tested this yet, but it should work in theory.


ad-tra (commented):
It was unintentional 

I didn't change any `Cargo.toml`, but I ran `cargo generate-lockfile` for testing purposes which according to the docs will auto-upgrade if the lock file already exists: 
> This command will create the Cargo.lock lockfile for the current package or workspace. If the lockfile already exists, it **will be rebuilt with the latest available version of every package.**

PR #1470, opened 2022-12-30
core,server,admin: remove useless extern crate statements

steverusso:
I don't believe these `extern crate` statements serve any purpose whatsoever.


Parth (approved):

Parth:
I think some much older version of rust required it not sure tho, thanks regardless

PR #1476, opened 2023-01-01
server: fix metrics problems

ad-tra:
context: core
- [x] fix active users
- [x] capture metrics on sharing
- [ ] update grafana to show sharing metrics alongside existing metrics

fixes: #1296


server/server/src/schema.rs
```diff
@@ -49,9 +47,34 @@ pub mod v2 {
...
+            sizes: <Uuid, u64>,
+            google_play_ids: <String, Owner>,
+            stripe_ids: <String, Owner>,
+            app_store_ids: <String, Owner>,
+            last_seen: <Owner, u64>
+        }
+    }
 
     pub fn migrate(
-        source: &mut super::transaction::ServerV1, destination: &mut transaction::Server,
```
Parth, line 54: can you order the operations in the order that they appear in v3? makes it easier to make sure you didn't miss one, something tests or compiler will not check for

server/server/src/file_service.rs
```diff
@@ -216,8 +216,9 @@ pub async fn change_doc(
             return Err(ClientError(DocumentDeleted));
         }
 
+        tx.last_seen.insert(owner, get_time().0 as u64);
         // Here is where you would check if the person is out of space as a result of the new file.
-        // You could make this a transaction and check whether or not this is an increase in size or
+        // You could make this a transaction and che™ck whether or not this is an increase in size or
```
Parth, line 221: ?

server/server/src/file_service.rs
```diff
@@ -216,8 +216,9 @@ pub async fn change_doc(
             return Err(ClientError(DocumentDeleted));
         }
 
+        tx.last_seen.insert(owner, get_time().0 as u64);
```
Parth, line 219: I think you additionally want to do this in `upsert` if they're re-organizing their stuff, they're active

Parth: You don't want to flag them as active here, you want to flag them as active in the second transaction. 

Observe the details of this function, it's modeling what's known as the 2 phased commit.

there's files on disk that we're trying to keep in sync with the contents of our database.

And this function can be interrupted at any point and still be fine:

1. we do some prelim checks to make sure they can update this file
2. we write the new file
3. we update our database to point to a new file
4. we cleanup the old file

if we were naive about it we would just do this:

1. overwrite the old file
2. update the pointer in our db

or 

1. update the pointer
2. overwrite the file

if either of those operations fail, we point to a file that doesn't exist 

in our 2 phased commit strategy, we can recover from a failure at any stage and the worst outcome is that we have a file that we need to cleanup.

server/server/src/metrics.rs
```diff
@@ -152,7 +161,7 @@ pub async fn get_user_billing_platform(
 
 pub async fn get_user_info(
     state: &ServerState, owner: Owner,
-) -> Result<(i64, i64, bool), ServerError<MetricsError>> {
+) -> Result<(i64, i64, bool, bool), ServerError<MetricsError>> {
```
Parth, line 164: this crosses the threshold for me with regards to number of acceptable tuple variants, I would make this a struct at this point. IMO as soon as there were 2 different fields with the same type we should have done it then.

Parth (changes requested):
this should be `server: fix metrics problems` as the title.

also you can sanity check this locally by spinning up a server, and looking at the contents of `localhost/metrics` or whatever. Create an account, shouldn't cause any active users, write a file, should cause an active user, etc.

You don't need to write a test for this scenario, but just commenting the local testing steps you took would go a long way.

server/server/src/file_service.rs
ad-tra: Lockbook™ 
lol, but that was a typo, apologies. As a windows user, I have the reflex of hitting ctrl+ 2 to switch tabs in vscode. In mac, I sometimes hit option+2 due to that reflex. option + 2 = ™

ad-tra:
> also you can sanity check this locally by spinning up a server, and looking at the contents of localhost/metrics or whatever. 

@Parth how to run a client that communicates with the local server instead of lockbook.net?
I attempted a local check, by deleting my `.lockbook` and trying to create a new account with the server running(`./run-server.sh`). I used the egui client. but the client was still communicating with lockbook.net instead of the local server instance. Is there a flag I need to pass when spinning up the egui app?

Parth:
You can override the environment variable `API_URL` and use the CLI. The way I do it is:
(forgot the exact port but you can look at the logs)
`export API_URL=http://localhost:8000`

ad-tra:
@parth just gave this a sanity check; everything works smoothly! Both the number of users who use share and active users are calculated. 

The two small tests I did: 
- running an egui client. Creating a new user. No active users were added, only when I create a new file, insert content, and sync that it does increase the active user count.  
- ran the `shares_finalized` test manually and it increase the share feature users by two.

Issue #1449, opened 2022-11-22
local usage history in core


Parth:

Parth:
similar to #1296, define, in core, add 2 tables that keep track of the files the user reads / writes the most often, migrate over the old data into a new schema.

I'm thinking 2 tables with id -> usize (representing a count of read or write).

@SmailBarkouch how does android currently do recent files? it does it inside android based on metadata right? 

@ad-tra you should expose lib.rs endpoints and then work with smail to get android to show this information properly.

ad-tra:
Instead of adding two tables that keep track of writes and reads per file, why not just add two attributes to the `FileMetadata` struct and store the info there?

also, what do you think about modeling read and write history as a time series instead of a counter? Collecting timestamps of the operations might allow us to perform smart personalization. I don't know if this is possible, given our commitment to data privacy, but an ideal state would have a small prediction engine that recommends files a user might want to read/edit based on past usage. 

Example: I have a doc `journal. md` that I append to each day before going to sleep at 11:00. It would be ideal for core to recommend `jounral.md` at night while suggesting different files in the morning.

Parth:
Here's a whole bunch of surrounding context in addition to the answer to your q: 

> Instead of adding two tables that keep track of writes and reads per file, why not just add two attributes to the FileMetadata struct and store the info there?

`FileMetadata` is used everywhere and is perhaps one of the hardest structs to modify. It would take small not-really-breaking change and turn it unto a platform wide paradigm shift. It's possible that this data, or other data may make it's way into the struct at some point, but we want to be really sure about the things that do. Things that make it into `FileMetadata` impact all clients, all integrations, any servers without consideration for their use case. For something like `name` this is fine. For something like "metrics associated with reads and writes" there's a few more considerations:

The first is does this need to be synced? Does this data need to be backed up, replicated across devices, etc. Deciding yes may be tempting but is a literal cost to us, and in a lot of contexts it's nicer for devices to store info local-only. I'm interested in my work laptop, my phone, and my tablet having different histories and recommendations.

Having data be synced to the server also makes features online-only, we want to let people do mostly everything offline. So if this data needs to make it to the server then we need to decide at what point that happens, and what to do if the server's offline, what to do if someone changed the data after we read it, but before we wrote it. Generally an app would just make these interactions block for network access, but we want to be more thoughtful about this.

[...]

PR #1485, opened 2023-01-08
core: local usage history / active documents 

ad-tra:
context: core
fixes: #1449


libs/core/src/lib.rs
```diff
@@ -372,6 +372,14 @@ impl<Client: Requester> CoreLib<Client> {
         })
     }
 
+    #[instrument(level = "debug", skip(self), err(Debug))]
+    pub fn get_most_active_documents(&self) -> Result<Vec<(Uuid, usize)>, Error<GetUsageError>> {
```
ad-tra, line 376: Hey, @SmailBarkouch, I exposed an endpoint in core `get_most_active_documents ` that you can use in android to improve the "Recent Documents" view. This endpoint will recommend more relevant files for the user.

The current approach for recommendation is to build a tree of all the files the user has access to, then sort them based on the newest. 
This new approach measures the count of read/write operations per document, then returns the most active documents (the ones with the most reads/writes).

This alternative approach will thus give more engaging file suggestions. How does this sound to you? Any editions or modifications you recommend?

libs/core/src/repo/schema_v3.rs
```diff
@@ -0,0 +1,123 @@
...
+
+hmdb::schema! {
+    CoreV3 {
+        account: <OneKey, Account>,
+        last_synced: <OneKey, i64>,
+        root: <OneKey, Uuid>,
+        local_metadata: <Uuid, SignedFile>,
+        base_metadata: <Uuid, SignedFile>,
+        read_activity: <Uuid, usize>,
+        write_activity: <Uuid, usize>,
```
Parth, line 27: thinking about whether this should be 1 or 2 tables, also maybe pop these guys at the bottom

Parth: I think 2 tables is fine

libs/core/src/repo/schema_v3.rs
```diff
@@ -0,0 +1,123 @@
...
+                }
+                for (k, v) in source.username_by_public_key.get_all()? {
+                    tx.username_by_public_key.insert(k, v);
+                }
+                Ok::<_, hmdb::errors::Error>(())
+            })
+            .expect("failed to migrate local database from v2 to v3")?;
+        }
+
+        // migrate documents from id+source structure to id+hmac structure
```
Parth, line 68: still a wip I assume because you didn't actually solicit a review, but v2 -> v3 will likely not need any of this stuff

Parth:
overall looking good so far

libs/core/src/service/document_service.rs
```diff
@@ -20,6 +20,9 @@ impl<Client: Requester> RequestContext<'_, '_, Client> {
 
         let doc = tree.read_document(self.config, &id, account)?;
 
+        let read_count = *self.tx.write_activity.get(&id).unwrap_or(&1);
```
SmailBarkouch, line 23: Did you mean to access `tx.read_activity` here?

libs/core/src/service/activity_service.rs
```diff
@@ -0,0 +1,25 @@
...
+use crate::{CoreResult, RequestContext};
+use itertools::Itertools;
+use uuid::Uuid;
+
+impl<Client: Requester> RequestContext<'_, '_, Client> {
+    pub fn most_active_documents(&mut self) -> CoreResult<Vec<(Uuid, usize)>> {
+        let mut result: HashMap<Uuid, usize> = HashMap::new();
+
+        for (k, v) in self.tx.read_activity.get_all() {
+            result.insert(*k, if result.contains_key(k) { *v } else { *v + result[k] });
```
SmailBarkouch, line 13: If `result` does contain a key already, wouldn't you add what it has? (in other words, I think you should take the inverse of your `if` statement)

libs/core/src/service/activity_service.rs
```diff
@@ -0,0 +1,25 @@
...
+        }
+        for (k, v) in self.tx.write_activity.get_all() {
+            result.insert(*k, if result.contains_key(k) { *v } else { *v + result[k] });
```
SmailBarkouch, line 16: Similar concern to the comment above.

libs/core/src/lib.rs
SmailBarkouch: I think this is a good approach, but I would somewhat argue there should be a balance between ranking what is accessed most, and what was most recently used. I feel this approach kinda makes "Recent Documents" into "Most Accessed Documents", as the most recently used file may not be accessed nearly as much (like a new file).

How this balance should be achieved is arguable. One approach I am thinking of is first ordering documents into "buckets" of time ranges (within 5 hour, 5 hour to 1 day, 1 day to 3 days, and so on), and within those ranges you do these read and write comparisons. To be honest this does complicate it a lot so definitely some other approach may be preferred.

libs/core/src/service/activity_service.rs
SmailBarkouch: To get rid of the if statement, you can use this: `result.get(k).unwrap_or_else(0) + *v`.

ad-tra: ty for the oneliner!

Parth (changes requested):

libs/core/src/repo/schema_v3.rs
```diff
@@ -0,0 +1,68 @@
...
+        read_activity: <Uuid, usize>,
+        write_activity: <Uuid, usize>
+    }
+}
+
+impl CoreV3 {
+    pub fn init_with_migration(writeable_path: &str) -> Result<CoreV3, CoreError> {
+        let db = CoreV3::init(writeable_path)?;
+
+        // migrate metadata from v1
```
Parth, line 35:
```suggestion
        // migrate metadata from v2
```

Parth (changes requested):
Additionally as we talked about, let's use a richer data model right off the bat, once you have that you can answer the 4 questions we talked about:

{most frequent, most recently} x {read, wrote}

and we can play around with some suggestive algorithm like @SmailBarkouch was thinking about.

libs/core/src/service/activity_service.rs
```diff
@@ -0,0 +1,75 @@
...
+
+        Ok(scores.into_iter().map(|f| f.0).collect_vec())
+    }
+
+    fn normalize(scores: &mut [(Uuid, DocActivityMetrics)]) {
+        let mut docs_avg_read_timestamps = scores
+            .iter_mut()
+            .map(|f| f.1.avg_read_timestamp)
+            .collect_vec();
+        docs_avg_read_timestamps.sort_by(|a, b| a.raw.cmp(&b.raw));
```
Parth, line 38: I wouldn't sort these to just get the min & max. That's `nlogn` vs `n`. I expect these to be reasonably large collections.

libs/core/src/service/document_service.rs
```diff
@@ -20,6 +23,23 @@ impl<Client: Requester> RequestContext<'_, '_, Client> {
 
         let doc = tree.read_document(self.config, &id, account)?;
 
+        let mut doc_events = self.tx.docs_events.get(&id).unwrap_or(&Vec::new()).clone();
+        doc_events.sort();
+        let latest_event = doc_events
+            .iter()
+            .filter(|e| matches!(e, DocEvents::Read(_)))
+            .last();
```
Parth, line 31: I would use this [`last()`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.last) instead, O(n) vs O(1)

Parth: You don't have to filter for the last read, you can just check if the last is a read, same goes for write.

Parth (changes requested):

libs/core/src/lib.rs
```diff
@@ -293,6 +293,10 @@ impl<Client: Requester> CoreLib<Client> {
         })
     }
 
+    #[instrument(level = "debug", skip(self), err(Debug))]
+    pub fn get_most_active_documents(&self) -> Result<Vec<Uuid>, Error<GetUsageError>> {
+        Ok(self.in_tx(|s| s.suggested_docs())?)
```
Parth, line 298: should call these the same thing

libs/core/src/repo/mod.rs
```diff
@@ -24,6 +25,7 @@ pub struct CoreV3 {
     pub local_metadata: LookupTable<Uuid, SignedFile>,
     pub base_metadata: LookupTable<Uuid, SignedFile>,
     pub pub_key_lookup: LookupTable<Owner, String>,
+    pub doc_events: LookupTable<Uuid, Vec<DocEvents>>,
```
Parth, line 28: You should add a `LookupList` to `db_rs`

libs/core/src/repo/schema_v3.rs
```diff
@@ -0,0 +1,69 @@
+use crate::CoreError;
```
Parth, line 1: this whole file can / should go away

libs/core/src/service/activity_service.rs
```diff
@@ -0,0 +1,75 @@
...
+        Ok(scores.into_iter().map(|f| f.0).collect_vec())
+    }
+
+    fn normalize(scores: &mut [(Uuid, DocActivityMetrics)]) {
+        let docs_avg_read_timestamps = StatisticValueRange {
+            max: scores
+                .iter_mut()
+                .map(|f| f.1.avg_read_timestamp)
+                .max()
+                .unwrap(),
```
Parth, line 36: no unwraps, I believe if a doc has no activity this will crash

Parth: you can `unwrap_or` or map it to an error

libs/core/src/service/activity_service.rs
```diff
@@ -0,0 +1,75 @@
...
+            min: scores
+                .iter_mut()
+                .map(|f| f.1.avg_read_timestamp)
+                .min()
+                .unwrap(),
+        };
+
+        let docs_avg_write_timestamps = StatisticValueRange {
+            max: scores
+                .iter_mut()
```
Parth, line 46: why are these `iter_mut`s?

libs/core/src/service/document_service.rs
```diff
@@ -41,6 +65,27 @@ impl<Client: Requester> CoreState<Client> {
         let hmac = tree.find(&id)?.document_hmac();
         document_repo::insert(&self.config, &id, hmac, &encrypted_document)?;
 
+        let mut doc_events = self
+            .db
+            .doc_events
+            .data()
+            .get(&id)
+            .unwrap_or(&Vec::new())
+            .clone();
```
Parth, line 74: could be an expensive operation

libs/core/src/service/document_service.rs
```diff
@@ -41,6 +65,27 @@ impl<Client: Requester> CoreState<Client> {
...
+
+        doc_events.sort_by(|a, b| b.cmp(a)); //sort in descending order
```
Parth, line 76: could be an expensive operation

Parth: can't you just keep the lists sorted by time?

libs/core/src/service/document_service.rs
```diff
@@ -41,6 +65,27 @@ impl<Client: Requester> CoreState<Client> {
...
+        let latest_event = doc_events.iter().find(|e| matches!(e, DocEvents::Read(_)));
```
Parth, line 77: i think you just need to peak at the last element and determine whether enough time has elapsed before writing

libs/core/src/service/document_service.rs
```diff
@@ -41,6 +65,27 @@ impl<Client: Requester> CoreState<Client> {
...
+
+        let is_capped = match latest_event {
+            Some(event) => Utc::now().timestamp() - event.timestamp() > RATE_LIMIT_MILLIS,
+            None => true,
+        };
+
+        if !is_capped {
+            doc_events.push(document_repo::DocEvents::Read(Utc::now().timestamp()));
+            self.db.doc_events.insert(id, doc_events)?;
```
Parth, line 86: This little interaction should entirely inform the db-rs `LookupList`

libs/core/src/service/document_service.rs
```diff
@@ -19,6 +21,28 @@ impl<Client: Requester> CoreState<Client> {
             .ok_or(CoreError::AccountNonexistent)?;
 
         let doc = tree.read_document(&self.config, &id, account)?;
+        self.db.doc_events.data();
```
Parth, line 24: same comments as write

libs/core/src/service/document_service.rs
Parth: there is also a copy paste related mistake here, avoid copy pasting things

Parth (changes requested):

ad-tra (commented):
- [x] [data model change](https://discord.com/channels/@me/1029027220179669004/1075120834961756180) use a list instead of linked list 
- [x] uniform use of iter 
- [ ] `///`
- [x] last access instead of averages

libs/core/libs/shared/src/document_repo.rs
```diff
@@ -1,15 +1,155 @@
...
+pub struct DocActivityMetrics {
+    pub id: Uuid,
+    pub latest_read_timestamp: StatisticValue,
+    pub latest_write_timestamp: StatisticValue,
+    pub read_count: StatisticValue,
+    pub write_count: StatisticValue,
+}
+impl DocActivityMetrics {
+    pub fn rank(&self) -> i64 {
+        (self.latest_read_timestamp.normalized + self.latest_write_timestamp.normalized) as i64 * 70
```
ad-tra, line 85: @Parth, What do you think about exposing the weights (70%, 30%) in the core endpoint as parameters to give flexibility to the caller? That way, this endpoint can serve as both suggested documents and most recent (if you pass the timestamp weight to 0), most accessed(if you pass the write/read count as 0 )

server/server/src/file_service.rs
ad-tra, line 1: The change in this file is unrelated to the pr main goal. Clippy was complaining, so I just applied its suggestions.

server/server/src/utils.rs
ad-tra, line 1: same as `file_service.rs`

libs/core/libs/shared/src/document_repo.rs
Parth: I think we should expose these weights when clients want to customize them / expose them to users, for now it's fine

libs/core/libs/shared/src/document_repo.rs
```diff
@@ -1,15 +1,156 @@
...
+    pub min: StatisticValue,
+}
+impl StatisticValue {
+    pub fn normalize(&mut self, range: StatisticValueRange) {
+        self.normalized = (self.raw - range.min.raw) as f64 / (range.max.raw - range.min.raw) as f64
+    }
+}
+/// DocActivityMetrics stores key document activity features, which are used to recommend relevant documents to the user.
+/// Here's a walkthrough of the recommendation procedure: collect 1k most recent document events (write/read), use that activity to construct a DocActivtyMetrics struct for each document. Min-max normalizes the activity features, then rank the documents.
+/// latest_read_timestamp: the latest epoch timestamp that the user read a document
```
Parth, line 72: you could just add these as field level comments, but not a huge deal

libs/core/libs/shared/src/document_repo.rs
```diff
@@ -1,15 +1,156 @@
...
+}
+
+#[derive(Default, Copy, Clone, PartialEq)]
+pub struct StatisticValue {
+    pub raw: i64,
+    pub normalized: f64,
+}
+impl StatisticValue {
+    fn from_raw(raw: i64) -> StatisticValue {
+        StatisticValue { raw, normalized: f64::default() }
```
Parth, line 43: I would personally either
1. not have this exist until you can calculate the normalized value
2. make `normalized: Option<f64>` to represent it's un-calculated state

usually I try to do 1

libs/core/libs/shared/src/document_repo.rs
```diff
@@ -1,15 +1,156 @@
...
+    pub id: Uuid,
+    pub latest_read_timestamp: StatisticValue,
+    pub latest_write_timestamp: StatisticValue,
+    pub read_count: StatisticValue,
+    pub write_count: StatisticValue,
+}
+impl DocActivityMetrics {
+    pub fn rank(&self) -> i64 {
+        (self.latest_read_timestamp.normalized + self.latest_write_timestamp.normalized) as i64 * 70
+            + (self.read_count.normalized + self.write_count.normalized) as i64 * 30
```
Parth, line 87: I would however extract these as constants

libs/core/libs/shared/src/document_repo.rs
```diff
@@ -1,15 +1,156 @@
...
+#[derive(Default, Copy, Clone)]
+pub struct DocActivityMetrics {
+    pub id: Uuid,
+    pub latest_read_timestamp: StatisticValue,
+    pub latest_write_timestamp: StatisticValue,
+    pub read_count: StatisticValue,
+    pub write_count: StatisticValue,
+}
+impl DocActivityMetrics {
+    pub fn rank(&self) -> i64 {
```
Parth, line 85: rank usually indicates a position in a list (think valedictorian). score is what I think you're going for here (think gpa).

libs/core/libs/shared/src/document_repo.rs
```diff
@@ -1,15 +1,156 @@
...
+                None => {
+                    set.insert(event.id(), vec![event]);
+                }
+                Some(events) => {
+                    events.push(event);
+                }
+            }
+        }
+
+        for pair in set {
```
Parth, line 112:
```suggestion
        for (key, value) in set {
```

Parth: then you don't have to do the `.1` `.0` business

Parth: can probably do one step better than `key`, `value`

libs/core/libs/shared/src/document_repo.rs
```diff
@@ -1,15 +1,156 @@
...
+            let read_events: Vec<&&DocEvent> = pair
+                .1
+                .iter()
+                .filter(|e| matches!(e, DocEvent::Read(_, _)))
+                .collect();
+            let latest_read = read_events
+                .iter()
+                .max_by(|x, y| x.timestamp().cmp(&y.timestamp()));
```
Parth, line 120: can't you just say it's the _last_ read?

libs/core/libs/shared/src/document_repo.rs
```diff
@@ -1,15 +1,156 @@
...
+                    set.insert(event.id(), vec![event]);
+                }
+                Some(events) => {
+                    events.push(event);
+                }
+            }
+        }
+
+        for pair in set {
+            let read_events: Vec<&&DocEvent> = pair
```
Parth, line 113: I don't think you actually need random access to the read events, you just need either the `.count()` or the `.last()` element. `collect()`'ing into a `vec` is an expensive allocation that it doesn't look like you benefit from here.

libs/core/src/service/activity_service.rs
```diff
@@ -0,0 +1,71 @@
...
+            max: self.map(|f| f.latest_write_timestamp).max().unwrap(),
+            min: self.map(|f| f.latest_write_timestamp).min().unwrap(),
+        };
+        self.for_each(|f| {
+            f.read_count.normalize(read_count_range);
+            f.write_count.normalize(write_count_range);
+            f.latest_read_timestamp.normalize(latest_read_range);
+            f.latest_write_timestamp.normalize(latest_write_range);
+        });
+        vec![]
```
Parth, line 69: looks like this is a prototyping mistake?

libs/core/src/service/activity_service.rs
```diff
@@ -0,0 +1,71 @@
...
+    fn normalize(&mut self) -> Vec<DocActivityMetrics>;
+}
+impl<'a, T> Normalizer for T
+where
+    T: Iterator<Item = &'a mut DocActivityMetrics>,
+{
+    fn normalize(&mut self) -> Vec<DocActivityMetrics> {
+        let read_count_range = StatisticValueRange {
+            max: self.map(|f| f.read_count).max().unwrap(),
+            min: self.map(|f| f.read_count).min().unwrap(),
```
Parth, line 48: no unwraps!

libs/core/src/service/activity_service.rs
```diff
@@ -0,0 +1,71 @@
...
+use db_rs::DbError;
+use lockbook_shared::document_repo::DocActivityMetrics;
+use lockbook_shared::document_repo::DocEvent;
+use lockbook_shared::document_repo::StatisticValueRange;
+use lockbook_shared::document_repo::Stats;
+use uuid::Uuid;
+
+impl<Client: Requester> CoreState<Client> {
+    pub(crate) fn suggested_docs(&mut self) -> LbResult<Vec<Uuid>> {
+        let mut scores = self
```
Parth, line 13: with the current implemtnation, I'm pretty sure `score == vec![]` no matter what the current state of things is. And in some situations you're going to panic. You need some tests in these parts to catch these mistakes.

libs/core/src/service/activity_service.rs
```diff
@@ -0,0 +1,71 @@
...
+            .db
+            .doc_events
+            .data()
+            .iter()
+            .get_activity_metrics()
+            .iter_mut()
+            .normalize();
+
+        scores.sort_by(|a, b| DocActivityMetrics::rank(a).cmp(&DocActivityMetrics::rank(b)));
```
Parth, line 22: I think it makes sense to sort_by this by default

Parth (changes requested):

libs/core/libs/shared/src/document_repo.rs
ad-tra: it's either collecting into a vec or cloning the iter.

Parth (approved):

Issue #1228, opened 2022-07-16
limit user defined fields


Parth:
malicious people can probably store ungodly amounts of text in strange places like usernames, or filenames. Set some limits in core and server for the overall size of these fields.

Parth:
Identify the places that people get to determine fields that are not documents (usernames and filenames I think are the only things @tvanderstad anything else come to mind?). 

Fields are encrypted so this isn't an exact science, once you've identified a max size hunt down the 2 places (new account, upsert) where these fields are used, introduce a new error for them being too large, and make the code base compile by addressing the errors throughout the stack.

With regards to upsert, upsert is called during a sync, and we don't want a user to successfully create a file locally and then never be able to sync it. So also validate the name size during file creation. 

You're gonna wanna identify a max size unencrypted and a max encrypted size. You can be rough here, but it should never be possible to successfully create a file and then not be able to sync it.

Parth:
(these fields are stored in memory, hence the importance of protecting against abuse)

tvanderstad:
A rogue client could store arbitrary data in anything encrypted, like the user/folder access keys

Parth:
Yeah so you'd probably want to limit those per key to a reasonable amount as well

PR #1493, opened 2023-01-10
core, server: limit the size of user fields 

ad-tra:
context: core, server

server: 
- [x] usernames 
- [x] filenames 
- [ ] access keys

core: 
- [x] usernames 
- [x] filenames 
- [ ] access keys

fixes: #1228


libs/core/libs/shared/src/api.rs
```diff
@@ -94,6 +97,7 @@ pub enum ChangeDocError {
     DocumentDeleted,
     NotPermissioned,
     OldVersionIncorrect,
+    InvalidFileName,
```
Parth, line 100: prob doesn't need to be here

libs/core/src/model/errors.rs
```diff
@@ -145,6 +145,7 @@ pub enum CoreError {
     InvalidCardNumber,
     InvalidPurchaseToken,
     InvalidAuthDetails,
+    InvalidFileName,
```
Parth, line 148: if you're adding maybe just call it `FileNameTooLong`

libs/core/src/service/account_service.rs
```diff
@@ -14,6 +14,9 @@ impl<Client: Requester> RequestContext<'_, '_, Client> {
     ) -> CoreResult<Account> {
         let username = String::from(username).to_lowercase();
 
+        if username.len() > 254 {
```
Parth, line 17: this can prob be a smaller value lol, I think like 20

PR #1497, opened 2023-01-11
server: support sandbox testing requests

SmailBarkouch:
The app store verifies our Storekit implementation by testing it out themselves. They do this using "sandbox testing," which utilizes our production server.

Formerly, this was not supported since I believed our production server would not need to have testing capabilities, but since then I have learned apple's app review team may need it.

### DIFF
--- a/server/server/src/billing/app_store_client.rs
+++ b/server/server/src/billing/app_store_client.rs
@@ -1,4 +1,6 @@
-use crate::billing::app_store_model::{LastTransactionItem, SubsStatusesResponse, TransactionInfo};
+use crate::billing::app_store_model::{
+    ErrorBody, LastTransactionItem, SubGroupIdentifierItem, SubsStatusesResponse, TransactionInfo,
+};
 use crate::config::AppleConfig;
 use crate::{ClientError, ServerError};
 use itertools::Itertools;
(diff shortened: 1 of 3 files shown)


server/server/src/billing/app_store_client.rs
```diff
@@ -57,69 +59,107 @@ pub async fn get_sub_status(
...
             }
 
-            Err(internal!("No usable data returned from apple subscriptions statuses endpoint despite assumed match. resp_body: {:?}, monthly_sub_group: {}", sub_status, config.monthly_sub_group_id))
+            Err(internal!("No usable data returned from apple's production subscriptions statuses endpoint despite assumed match. resp_body: {:?}, monthly_sub_group: {}", sub_status, config.monthly_sub_group_id))
         }
         400 | 404 => {
-            let resp_body = resp.text().await.ok();
+            let error: ErrorBody = resp.json().await?;
+
+            if error.error_code == 4040005 {
```
Parth, line 85: thought you said it was: 21007

Parth (changes requested):
how did you test?

server/server/src/billing/app_store_client.rs
SmailBarkouch: That was for the "verify receipt" endpoint, but no error code like that exists for the "get all subscription statuses" endpoint which we now use. This error code signifies that the original transaction id was not found, so when that happens I check sandbox.

The way this implementation now runs is that I will first check production for a subscription, then sandbox.

I tested this myself using a sandbox account, verifying it successfully makes the jump to check sandbox.

Parth (approved):

PR #1515, opened 2023-01-13
core, server: set size limits on userfields

ad-tra:
context: core, server
server: 
- [x] usernames 
- [x] filenames 

core: 
- [x] usernames 
- [x] filenames 

fixes: #1228


libs/core/libs/shared/src/api.rs
```diff
@@ -93,6 +96,7 @@ pub enum ChangeDocError {
     DocumentNotFound,
     DocumentDeleted,
     NotPermissioned,
+    FileNameTooLong,
```
ad-tra, line 99: > @parth: prob doesn't need to be here
the file name length should be checked both in upsert and when the user renames. So it should be included.

libs/core/src/service/account_service.rs
```diff
@@ -14,6 +14,10 @@ impl<Client: Requester> RequestContext<'_, '_, Client> {
     ) -> CoreResult<Account> {
         let username = String::from(username).to_lowercase();
 
+        if username.len() > 254 {
```
ad-tra, line 17: > @Parth this can prob be a smaller value lol, I think like 20

Parth (changes requested):
Your approach for usernames is fine. 

Your approach for filenames is not quite what we want:

You don't want to hunt down all the places that names could change. 

Checkout `validate.rs`, specifically the `validate()` function. You can add a size check (on the encrypted copy). And it will apply everywhere (renames, paths, creates etc).

Parth (changes requested):
This looks better, but a few more things required:
+ [x] `get_encryption_overhead` shouldn't be called for usernames, they're always cleartext.
+ [x] that function is only ever called with 1 value (right?) you can just have a const for the `MAX_ENCRYPTED_FILENAME_LENGTH` with the math hardcoded
 + [x] you want to go through `lib.rs` and see which endpoints could cause a filename to become too long (create, rename, etc) and you don't want this new error (`FileNameTooLong`) to be mapped to `Unexpected`. You want it to be handled by our various clients.
 + [ ] when you do this, a handful of clients will break, you want to chase down those compiler errors and CI failures and fix them.
 + [ ] Currently if you use CLI to create a file with a name too long you'll just get an `Unexpected`, the target state would provide a better error message, and our tooling should help you do that. @SmailBarkouch can do iOS and android.
+ [ ] is a UUID 50 characters long? I thought it was shorter than that by a good margin.

libs/core/libs/shared/src/validate.rs
```diff
@@ -80,6 +81,11 @@ where
     // note: deleted access keys permissible
     pub fn assert_all_files_decryptable(&mut self, owner: Owner) -> SharedResult<()> {
         for file in self.ids().into_iter().filter_map(|id| self.maybe_find(id)) {
+            if file.secret_name().encrypted_value.value.len() > MAX_ENCRYPTED_FILENAME_LENGTH {
```
Parth, line 84: deserves it's own fn, idk if it makes sense to tack this check onto this fn

libs/core/tests/file_ops.rs
```diff
@@ -55,6 +56,13 @@ fn name_empty() {
     assert_matches!(core.rename_file(id, ""), Err(UiError(RenameFileError::NewNameEmpty)));
 }
 
+#[test]
+fn name_invalid() {
+    let core = test_core_with_account();
+    let result = core.create_at_path(&"x".repeat(MAX_FILENAME_LENGTH + 1));
```
Parth, line 62: is this a copy paste error?

Parth (changes requested):
get this rebased and we'll get the errors piped through this week

ad-tra:
Make sure not to return unexpected

Parth:
Before we merge this, we need to run the new validations with the current server data to make sure there aren't any files currently that are too long

Parth (approved):

libs/core/libs/shared/src/validate.rs
```diff
@@ -76,7 +78,6 @@ where
 
         Ok(())
     }
-
```
steverusso, line 79: Leave this in, and then...

libs/core/libs/shared/src/validate.rs
```diff
@@ -92,6 +93,16 @@ where
...
+    pub fn assert_all_filenames_size_limit(&self) -> SharedResult<()> {
+        for file in self.all_files()? {
+            if file.secret_name().encrypted_value.value.len() > MAX_ENCRYPTED_FILENAME_LENGTH {
+                return Err(SharedErrorKind::ValidationFailure(
+                    ValidationFailure::FileNameTooLong(*file.id()),
+                ))?;
+            }
+        }
+        Ok(())
+    }
```
steverusso, line 105: ... add one to give functions some breathing room.

libs/core/src/lib.rs
steverusso, line 1: @Parth should `create_at_path` also have `FileNameTooLong` as an expected error?

server/server/src/utils.rs
```diff
@@ -1,11 +1,15 @@
 use crate::ServerError;
-use lockbook_shared::api::{GetBuildInfoError, GetBuildInfoResponse};
+use lockbook_shared::{
+    account::MAX_USERNAME_LENGTH,
+    api::{GetBuildInfoError, GetBuildInfoResponse},
+};
```
steverusso, line 5:
```suggestion
use lockbook_shared::account::MAX_USERNAME_LENGTH;
use lockbook_shared::api::{GetBuildInfoError, GetBuildInfoResponse};
```

server/server/src/file_service.rs
```diff
@@ -88,6 +88,7 @@ pub async fn upsert_file_metadata(
 
     for update in request.updates {
         let new = update.new;
+
```
steverusso, line 91:
```suggestion
```
Try to remove all standalone inconsequential changes like this, esp before final review.

steverusso (changes requested):

libs/core/src/lib.rs
steverusso: @ad-tra @Parth will `create_at_path` ever return a `FileNameTooLong`? Don't we have to check the path components?

Parth: @ad-tra `create_link_at_path` too. Can you throw some sanity-check-tests in there too?

Parth (changes requested):

steverusso (approved):

PR #1525, opened 2023-01-19
admin/server: add support for more indexes to rebuild; fix existing one

tvanderstad:
code based on server schema v2 migration deleted in this commit: https://github.com/lockbook/lockbook/commit/ea5ecb9c7e36af5b3571d1c2ce0e49329ca26852#

### DIFF
--- a/server/server/src/file_service.rs
+++ b/server/server/src/file_service.rs
@@ -804,15 +804,59 @@ pub async fn admin_rebuild_index(
         .index_db
         .transaction(|tx| match context.request.index {
             ServerIndex::OwnedFiles => {
-                tx.owned_files.clear();
+                let mut owned_files = HashMap::new();
+                for owner in tx.accounts.keys() {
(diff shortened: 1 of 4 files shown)


server/server/src/file_service.rs
```diff
@@ -804,15 +804,59 @@ pub async fn admin_rebuild_index(
...
+                    shared_files.insert(*owner, HashSet::new());
+                }
                 for (id, file) in tx.metas.get_all() {
-                    let mut set = tx
-                        .owned_files
-                        .get(&file.owner())
-                        .cloned()
-                        .unwrap_or_default();
-                    set.insert(*id);
-                    tx.owned_files.insert(file.owner(), set);
```
tvanderstad, line 815: iiuc this doesn't create empty sets for users with no files, which is prob fine bc each user has a root, but I did it more explicitly for clarity and consistency with how other indexes are rebuilt

Parth (approved):

PR #1550, opened 2023-01-24
editor: undo/redo

tvanderstad:


libs/editor/egui_editor/Cargo.toml
```diff
@@ -21,6 +21,7 @@ egui_wgpu_backend = "0.19"
 libc = "0.2"
 pollster = "0.2"
 pulldown-cmark = { version = "0.9.2", default-features = false }
+queues = "1.0.2"
```
Parth, line 24: This is prob the only thing I have a problem with here

Parth (approved):

PR #1556, opened 2023-01-29
admin, server: track stripe billing state, fix webhooks, and add upgrade to premium command

SmailBarkouch:
Recently, we noticed a bug with stripe billing. If a user experienced a payment failure in one of their cards, which afterwards, was immediately resolved, bugs would ensue. The way stripe billing currently works, we clear a user's billing information the moment they fail a payment. This makes it impossible to recover from a billing failure via webhooks. In addition to this, the subscription field in the webhooks do not always contain the data we expect. Sometimes it holds just an ID, and other times, the subscription object associated with the ID.

## Fixes

This PR contains three changes:
1. If a subscription object is not present in the webhooks, we will send a request to stripe to receive it. The subscription object is used to populate our stripe metadata to help the user understand when their subscription will expire.
2. We do not clear a user's stripe billing information when they cancel or a failed payment. Now we set state within our stripe metadata to signify their billing situation.
3. A new admin command is added to manually populate a user's billing information. It contains three subcommands, one for each of the billing platforms (stripe, google play, and apple).


libs/core/libs/shared/src/api.rs
```diff
@@ -719,3 +752,67 @@ impl Request for AdminRebuildIndexRequest {
...
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone)]
+pub enum AppStoreAccountState {
+    Ok,
+    GracePeriod,
+    FailedToRenew,
+    Expired,
+}
+
+impl FromStr for StripeAccountState {
+    type Err = ();
```
Parth, line 780: I wonder if this is gonna be a nightmare to debug, I think even `String` could be fine with an error message, maybe you can include `s` in the error.

server/server/src/billing/stripe_client.rs
```diff
@@ -294,6 +294,21 @@ pub async fn cancel_subscription(
     Ok(())
 }
 
+pub async fn get_subscription(
+    stripe_client: &stripe::Client, subscription_id: &stripe::SubscriptionId,
+) -> Result<stripe::Subscription, SimplifiedStripeError> {
+    {
+        let subscription_id = subscription_id.as_str();
+        info!(?subscription_id, "Retrieving stripe subscription");
```
Parth, line 302: maybe `debug!`

Parth (approved):

PR #1560, opened 2023-01-31
infra: auto bump version

ad-tra:
context: infra


utils/releaser/src/main.rs
```diff
@@ -11,10 +11,16 @@ use crate::secrets::*;
 use crate::utils::root;
 
 use structopt::StructOpt;
+use utils::bump_versions;
 
 #[derive(PartialEq, StructOpt)]
+#[structopt(name = "basic")]
 enum Releaser {
-    All,
+    All {
```
Parth, line 19: This needs to be it's own command, `All` gets executed on a handful of machines at the same time

utils/releaser/src/utils.rs
```diff
@@ -69,3 +70,72 @@ pub fn root() -> PathBuf {
...
     project_root
 }
+
+pub fn determine_new_version(bump_type: Option<String>) -> Option<String> {
+    let mut current_version: Vec<i32> = core_version()
+        .split('.')
+        .map(|f| f.parse().unwrap())
+        .collect();
+
+    let bump_type = bump_type.unwrap_or_else(|| "patch".to_string());
```
Parth, line 80: reasonable and good

utils/releaser/src/utils.rs
```diff
@@ -69,3 +70,72 @@ pub fn root() -> PathBuf {
...
+    match bump_type.as_ref() {
+        "major" => current_version[0] += 1,
+        "minor" => current_version[1] += 1,
```
Parth, line 83: if a minor version is bumped, the patch version should be set to 0, similar deal w major

utils/releaser/src/utils.rs
```diff
@@ -69,3 +70,72 @@ pub fn root() -> PathBuf {
...
+pub fn bump_versions(bump_type: Option<String>) {
+    let new_version = determine_new_version(bump_type).unwrap_or_else(core_version);
+    let new_version = new_version.as_str();
+
+    let cargos_to_update = vec![
+        "/clients/admin/",
+        "/clients/cli/",
+        "/clients/egui/",
+        "/server/server/",
+        "/libs/core/",
```
Parth, line 120: there are a handful more hissing here that now we can just keep in sync. If we own the cargo toml it should prob stay in sync.

utils/releaser/src/utils.rs
```diff
@@ -69,3 +70,72 @@ pub fn root() -> PathBuf {
...
+        "$PLIST",
+    ]);
+
+    Command::new("/usr/libexec/Plistbuddy").args([
+        "-c",
+        "Set CFBundleShortVersionString $BUNDLE_SHORT_VERSION",
+        "$PLIST",
+    ]);
+    //android
+    Command::new("./gradlew").args(["wrapper", "--gradle-version", &core_version()]);
```
Parth, line 140: pretty sure this is wrong, i'm pretty sure this changes the wrapper gradle version, not our project's version.

You may have to explore: https://stackoverflow.com/questions/21329849/set-android-app-version-using-gradle

utils/releaser/src/utils.rs
```diff
@@ -69,3 +70,72 @@ pub fn root() -> PathBuf {
...
+        "/server/server/",
+        "/libs/core/",
+    ];
+    for cargo_path in cargos_to_update {
+        edit_cargo_version(cargo_path, new_version);
+    }
+
+    //apple
+    //TODO figure out what sbuild number and appropriate plist
+    Command::new("/usr/libexec/Plistbuddy").args([
```
Parth, line 128: image

this are the info files you want to modify

`CFBundleShortVersionString` will get the core version

`CFBundleVersion` actually will never be updated I'm pretty sure

Parth:
When I do the release, bumping version is step 0, step 1 after that is usually going over to github releases, creating a new release with the next version as the name, tagging the latest commit on master (generally the version bump commit), and hitting the generate changelog button.

See if this stuff can be automated as well. @SmailBarkouch authored the github releases crate we use. Perhaps you could add this sort of functionality there. Check it out.

Parth:
So, in addition to what you have here, this guy should ideally push to master, create a tag, create a release with that tag and generate the changelog

(generally my order of certainty that can be trivially automated)

utils/releaser/src/utils.rs
ad-tra: These are the remaining rust packages, all of which have a version of `0.1.0`. Should they all be in sync with the rest?
- `lockbook-shared`
- `test_utils`
- `egui_editor`
- `lockbook-dev`
- `releaser`
- `winstaller`

utils/releaser/src/utils.rs
ad-tra: thanks for pointing that out. There's no gradle command that you call through cli to bump versions. I used regex instead

utils/releaser/src/utils.rs
Parth: yeah

utils/releaser/src/utils.rs
```diff
@@ -69,3 +71,106 @@ pub fn root() -> PathBuf {
...
+        .map(|f| f.to_string())
+        .collect::<Vec<String>>()
+        .join(".");
+
+    Some(new_version)
+}
+
+pub fn edit_cargo_version(cargo_path: &str, version: &str) {
+    let cargo_path = &[cargo_path, "/Cargo.toml"].join("");
+    let mut server = fs::read_to_string(cargo_path)
```
Parth, line 110: you don't want to call this `server` right?

utils/releaser/src/utils.rs
```diff
@@ -69,3 +71,106 @@ pub fn root() -> PathBuf {
...
+        edit_cargo_version(cargo_path, new_version);
+    }
+
+    //apple
+    let plists = ["clients/apple/iOS/info.plist", "clients/apple/macOS/info.plist"];
+    for plist in plists {
+        Command::new("/usr/libexec/Plistbuddy")
+            .args(["-c", &format!("Set CFBundleVersion {}", new_version), plist])
+            .spawn()
+            .unwrap();
```
Parth, line 167: we don't need to update this 
```suggestion

```

Parth (changes requested):

PR #1579, opened 2023-02-12
editor/egui client: update egui

tvanderstad:
This bumps egui version from 0.19 to 0.20, which has an image painting UI that supports resizing images (most recent is 0.21 but has additional breaking changes that I didn't want to deal with).

~Todo: for some reason, text is larger and colors aren't right 🤔~


steverusso (changes requested):
Might as well bump egui deps for our winstaller as well

steverusso (approved):

PR #1581, opened 2023-02-13
server: version check updates

### LINKED ISSUE #1563: version checks refinement
+ [x] remove the dual support version check business due to avoiding a breaking change
+ [x] version checks should contain a list of versions that are _not_ supported rather than ones that are. I don't want to update this list every release, I just want to add to unsupported versions whenever they become unsupported.

ad-tra:
context: server
fixes #1563


ad-tra:
Should the version header still be optional? The current impl makes it so that if you don't provide a version header, you will get `ClientUpdateRequired`, but why not catch the error at an earlier stage by leveraging `warp`

Parth (approved):

PR #1583, opened 2023-02-15
core: solve link problems

ad-tra:
context: core
fixes: #1640


libs/core/tests/sharing_tests.rs
```diff
@@ -299,7 +299,7 @@ fn write_document_in_rejected_shared_folder_in_share_folder() {
 
     assert_eq!(cores[1].read_document(document.id).unwrap(), b"document content by sharee");
     cores[1].sync(None).unwrap();
-    cores[1].get_file_by_id(document.id).unwrap();
+    //cores[1].get_file_by_id(document.id).unwrap();
```
tvanderstad, line 302: I think the test is correct and the code is somehow broken

the line is probably there to raise an issue like this even though the return value isn't used
the expected behavior is that if I share a folder with you, and I share a folder in that folder with you, if you reject one of those shares, I should still have that file because it is still in the other (not rejected) share

maybe resolve_and_finalize can't handle pending shares

indeed, it skips them on purpose https://github.com/lockbook/lockbook/blob/share-finalization/libs/core/libs/shared/src/core_ops.rs#L78-L99

Perhaps try resolve_and_finalize_all? Maybe the naming here could be better 😅 https://github.com/lockbook/lockbook/blob/share-finalization/libs/core/libs/shared/src/core_ops.rs#L101-L102

tvanderstad: and when you have the test passing I would see if you can make finalize not pub, remove its usages from anywhere that breaks, and replace those usages just like you did with `get_file_by_id`

tvanderstad: (copied from discord dms)

libs/core/src/service/file_service.rs
```diff
@@ -23,7 +23,11 @@ impl<Client: Requester> CoreState<Client> {
         let id =
             tree.create(Uuid::new_v4(), symkey::generate_key(), parent, name, file_type, account)?;
 
-        let ui_file = tree.finalize(&id, account, &mut self.db.pub_key_lookup)?;
+        let ui_file = tree
+            .resolve_and_finalize_all(account, [id].into_iter(), &mut self.db.pub_key_lookup)?
+            .get(0)
+            .ok_or(CoreError::InvalidFinalization)?
```
Parth, line 29: Should probably be `Unexpected` here

Parth: not a new-error-worthy

Parth (changes requested):
a lil unfortunate that we're allocating a bunch of 1-long vecs. I think there's probably a way to structure this code so you're returning a single `File` per invocation, and taking a single id.

furthermore the names of these functions need to be addressed, still not clear which someone should call `resolve_and_finalize` or `resolve_and_finalize_all`.

probably there should be one that takes 1 id returns one file, and another that takes `&[id]` and returns `Vec<File>`. I think you should probably have `is_visible_id(id) -> bool` and a `get_visible_ids() -> Vec<Uuid>`.

most of the `//`'s above these `fn`'s should be replaced by `///` (@tvanderstad)

libs/core/libs/shared/src/core_ops.rs
```diff
@@ -137,6 +150,14 @@ where
         Ok(files)
     }
 
+    fn is_invisible_id(&mut self, id: Uuid) -> SharedResult<bool> {
+        if self.calculate_deleted(&id)? || self.in_pending_share(&id)? || self.link(&id)?.is_some()
+        {
+            return Ok(true);
+        }
+        Ok(false)
```
tvanderstad, line 158:
```suggestion
        Ok(self.calculate_deleted(&id)? || self.in_pending_share(&id)? || self.link(&id)?.is_some())
```

libs/core/libs/shared/src/core_ops.rs
```diff
@@ -75,32 +75,41 @@ where
...
-    }
 
-    // this variant used when we must include all files e.g. work calculated
+        if let Some(new_parent) = parent_substitutions.get(&file.parent) {
+            file.parent = *new_parent;
+        }
+
+        Ok(file)
+    }
+    //finalizes multiple ids with the option to include all files or skip over invisible ones (eg useful when listing paths)
```
tvanderstad, line 109:
```suggestion
    /// finalizes multiple ids, optionally skipping invisible files (e.g. when listing paths)
```

tvanderstad: the triple slash makes it a doc comment which your IDE shows when you hover over a usage of the function and I think it needs a space (I think comments are just prettier with a space after the slashes anyway)

libs/core/libs/shared/src/core_ops.rs
tvanderstad: 😉

libs/core/libs/shared/src/core_ops.rs
```diff
@@ -75,32 +75,41 @@ where
...
-            }
-            if self.in_pending_share(&id)? {
-                continue;
-            }
-            if self.link(&id)?.is_some() {
-                continue;
+    //finalizes one single id.
+    pub fn resolve_and_finalize(
+        &mut self, account: &Account, id: Uuid, public_key_cache: &mut LookupTable<Owner, String>,
+    ) -> SharedResult<File> {
```
tvanderstad, line 81: can the body of this function just be
```
    self.resolve_and_finalize_all(account, [id], public_key_cache, false)
```

tvanderstad: @Parth is this presumably stack-allocated length-one slice a performance concern? I lean away from code duplication in this case

libs/core/libs/shared/src/core_ops.rs
```diff
@@ -28,7 +28,7 @@ where
     T: TreeLike<F = SignedFile>,
 {
     // todo: revisit logic for what files can be finalized and how e.g. link substitutions, deleted files, files in pending shares, linked files
-    pub fn finalize(
+    fn finalize(
```
tvanderstad, line 31: while we're here, I'm down to rename this to anything else, rename `resolve_and_finalize` to `finalize`, and rename `resolve_and_finalize_all` to `finalize_all`

libs/core/libs/shared/src/core_ops.rs
Parth: @tvanderstad

libs/core/libs/shared/src/core_ops.rs
```diff
@@ -28,17 +28,21 @@ where
     T: TreeLike<F = SignedFile>,
 {
     // todo: revisit logic for what files can be finalized and how e.g. link substitutions, deleted files, files in pending shares, linked files
```
Parth, line 30: can't this `todo` go now, @tvanderstad @ad-tra?

libs/core/libs/shared/src/core_ops.rs
```diff
@@ -28,17 +28,21 @@ where
...
         let meta = self.find(id)?.clone();
         let file_type = meta.file_type();
-        let parent = *meta.parent();
         let last_modified = meta.timestamped_value.timestamp as u64;
         let name = self.name_using_links(id, account)?;
-        let id = *id;
+        let parent = *meta.parent();
         let last_modified_by = account.username.clone();
 
+        let id = match file_type {
```
Parth, line 41: if we are directly getting the info of a link by id, it's id shouldn't change, right?

libs/core/libs/shared/src/core_ops.rs
```diff
@@ -75,50 +79,50 @@ where
...
-        for id in ids {
-            if self.calculate_deleted(&id)? {
-                continue;
-            }
-            if self.in_pending_share(&id)? {
-                continue;
-            }
-            if self.link(&id)?.is_some() {
-                continue;
+    //finalizes one single id.
```
Parth, line 82:
```suggestion
    /// finalizes one single id.
```

imo this comment provides little value, but an explanation of this function is required and important

libs/core/libs/shared/src/core_ops.rs
```diff
@@ -75,50 +79,50 @@ where
...
+        let file;
+
+        let finalized = self.decrypt(&id, account, public_key_cache)?;
+
+        match finalized.file_type {
+            FileType::Document | FileType::Folder => file = finalized,
+            FileType::Link { target } => {
+                let mut target_file = self.decrypt(&target, account, public_key_cache)?;
+
+                target_file.id = finalized.id;
```
Parth, line 95: you've done this already in `decrypt`. 

imo a refinement on the link discussion we had:

you're never going to change the id of a file. You're just going to change the parent sometimes, and you're going to change the parent if any link points to this file.

libs/core/libs/shared/src/core_ops.rs
```diff
@@ -128,15 +132,12 @@ where
...
-            if let Some(new_parent) = parent_substitutions.get(&item.parent) {
-                item.parent = *new_parent;
-            }
-        }
-
         Ok(files)
     }
 
+    fn is_invisible_id(&mut self, id: Uuid) -> SharedResult<bool> {
+        Ok(self.calculate_deleted(&id)? || self.in_pending_share(&id)? || self.link(&id)?.is_some())
```
Parth, line 139: if you agree with what I've said above, then the criteria for invisibility is:

invisible == true if:
+ deleted
+ in pending share
+ it's a link

Parth (changes requested):
slight tweak in approach makes things a little more straightforward, still largely what we talked about on the call regarding what id we're returning and who has to unwind what links. Mainly being more critical of which id the programmer asks to finalize, the link's id or the target's id when returning a shared folder.

I think we don't want to be in the business of changing any file's id, that sounds like a debugging disaster. Swapping a parent seems more fine.

Parth:
I'll review the rest once you do that ^

libs/core/libs/shared/src/core_ops.rs
```diff
@@ -75,50 +79,50 @@ where
...
-                continue;
-            }
-            if self.link(&id)?.is_some() {
-                continue;
+    //finalizes one single id.
+    pub fn finalize(
+        &mut self, account: &Account, id: Uuid, public_key_cache: &mut LookupTable<Owner, String>,
+    ) -> SharedResult<File> {
+        let file;
+
```
tvanderstad, line 87: I would remove this...
```suggestion
```

libs/core/libs/shared/src/core_ops.rs
```diff
@@ -75,50 +79,50 @@ where
...
+        let finalized = self.decrypt(&id, account, public_key_cache)?;
+
+        match finalized.file_type {
```
tvanderstad, line 90: ...declare/assign `file` here...
```suggestion
        let file = match finalized.file_type {
```

libs/core/libs/shared/src/core_ops.rs
```diff
@@ -75,50 +79,50 @@ where
...
+            FileType::Document | FileType::Folder => file = finalized,
```
tvanderstad, line 91: ...just use the expression here...
```suggestion
            FileType::Document | FileType::Folder => finalized,
```

libs/core/libs/shared/src/core_ops.rs
```diff
@@ -75,50 +79,50 @@ where
...
+            FileType::Link { target } => {
+                let mut target_file = self.decrypt(&target, account, public_key_cache)?;
+
+                target_file.id = finalized.id;
+                target_file.parent = finalized.parent;
+                target_file.name = finalized.name;
+
+                file = target_file;
```
tvanderstad, line 99: ...as well as here
```suggestion
                target_file
```

libs/core/libs/shared/src/core_ops.rs
tvanderstad: I agree

libs/core/libs/shared/src/core_ops.rs
tvanderstad: I agree

libs/core/libs/shared/src/core_ops.rs
tvanderstad: right - if there's a link to this file, substitute its name and parent for the link's name and parent. I don't think the type of this file matters

libs/core/libs/shared/src/core_ops.rs
```diff
@@ -75,50 +79,50 @@ where
...
-            user_visible_ids.push(id);
         }
-        self.resolve_and_finalize_all(account, user_visible_ids.into_iter(), public_key_cache)
-    }
 
-    // this variant used when we must include all files e.g. work calculated
-    pub fn resolve_and_finalize_all<I>(
+        Ok(file)
+    }
+    /// finalizes multiple ids, optionally skipping invisible files (e.g. when listing paths)
```
tvanderstad, line 105:
```suggestion
    }

    /// finalizes multiple ids, optionally skipping invisible files (e.g. when listing paths)
```

libs/core/libs/shared/src/core_ops.rs
```diff
@@ -128,15 +132,12 @@ where
...
-            }
-        }
-
         Ok(files)
     }
 
+    fn is_invisible_id(&mut self, id: Uuid) -> SharedResult<bool> {
+        Ok(self.calculate_deleted(&id)? || self.in_pending_share(&id)? || self.link(&id)?.is_some())
+    }
     pub fn create_op(
```
tvanderstad, line 141:
```suggestion
    }

    pub fn create_op(
```

libs/core/libs/shared/src/core_ops.rs
tvanderstad: same goes for `finalize_all` below

libs/core/libs/test_utils/src/assert.rs
```diff
@@ -185,7 +185,7 @@ pub fn local_work_paths(db: &Core, expected_paths: &[&'static str]) {
             let mut local = s.db.base_metadata.stage(&mut s.db.local_metadata).to_lazy();
             Ok(dirty
                 .iter()
-                .filter(|id| !matches!(local.find(id).unwrap().file_type(), FileType::Link { .. }))
+                .filter(|id| matches!(local.link(id).unwrap(), None))
```
tvanderstad, line 188: 👌

libs/core/src/service/file_service.rs
```diff
@@ -23,7 +23,10 @@ impl<Client: Requester> CoreState<Client> {
         let id =
             tree.create(Uuid::new_v4(), symkey::generate_key(), parent, name, file_type, account)?;
 
-        let ui_file = tree.finalize(&id, account, &mut self.db.pub_key_lookup)?;
+        let mut ui_file = tree.finalize(account, id, &mut self.db.pub_key_lookup)?;
+        if matches!(file_type, FileType::Link { .. }) {
+            ui_file.id = id;
+        }
```
tvanderstad, line 29: I think after responding to other feedback, this (and other instances below) won't be necessary

libs/core/tests/link_tests.rs
```diff
@@ -237,21 +245,94 @@ fn list_metadatas_nested_linked_folders() {
...
     assert::all_paths(&cores[1], &["/", "/link1/", "/link2/", "/link2/document"]);
-    assert::all_children_ids(&cores[1], roots[1].id, &[link1.id, link2.id]);
-    assert::all_children_ids(&cores[1], link2.id, &[document.id]);
+    assert::all_children_ids(&cores[1], roots[1].id, &[folder1.id, folder2.id]);
+    assert::all_children_ids(&cores[1], folder2.id, &[document.id]);
     assert::all_recursive_children_ids(
         &cores[1],
         roots[1].id,
-        &[roots[1].id, link1.id, link2.id, document.id],
+        &[roots[1].id, folder1.id, folder2.id, folder2.id, document.id],
```
tvanderstad, line 260: `folder2.id` showing up twice here 🤔

libs/core/libs/shared/src/core_ops.rs
ad-tra: yeah isn't that the conditions that the function counts:
- deleted `self.calculate_deleted(&id)? `
- in pending share `self.in_pending_share(&id)?`
- it's a link `self.link(&id)?.is_some()`

libs/core/src/service/file_service.rs
ad-tra: i think that's still needed because when creating a link, the link's id, instead of the target id should be returned. This mirrors the behavior that happens at [create_link_at_path](https://github.com/lockbook/lockbook/pull/1583/files#diff-2b163d75063f996c29e322b77ef8ccea13fc40d82e5220bc27c5fe8ad470a014R24)

Parth (changes requested):
further specs added

https://github.com/lockbook/lockbook/issues/1640#issuecomment-1506019182

libs/core/src/service/file_service.rs
Parth: This doesn't do anything tho? You decrypt a particular id, and assign it's id to the same id.

Parth (changes requested):
- testing operations ( write, read, move, rename, delete) behavior on a link by passing in 
   - the link target id. (this shouldn't be authorized)
   - the link id 
  
- moving a link withen a link? 
- what happens when target is modified (deleted, moved, renamed) by the sharer. 
- what if you share a link. a link to a link? is that possible if we should not allow that
- can you create two links to the same target? we shouldn't allow that

Parth:
{operations} x {link, target} x {doc folder} x {owner x sharee}

libs/core/tests/sharing_tests.rs
```diff
@@ -329,6 +329,213 @@ fn write_document_in_rejected_shared_folder_in_rejected_share_folder() {
...
+        .unwrap();
+    cores[1]
+        .write_document(link.id, b"document content by sharee")
+        .unwrap();
+
+    cores[1].sync(None).unwrap();
+    cores[0].sync(None).unwrap();
+
+    assert_eq!(cores[1].read_document(link.id).unwrap(), b"document content by sharee");
+    assert_eq!(cores[0].read_document(document1.id).unwrap(), b"document content by sharee");
```
Parth, line 360: Getting the link back is the most reasonable behavior for `create_link_at_path`, but I believe you're not going to want to `read_document` and accept a `link`. I think trying to `read_document` and passing a link is probably an error. Ideally you would tweak this case to make sure that you can read the document by both people, and then I wonder if this test exists already.

libs/core/tests/sharing_tests.rs
```diff
@@ -329,6 +329,213 @@ fn write_document_in_rejected_shared_folder_in_rejected_share_folder() {
...
+        .unwrap();
+
+    cores[0].sync(None).unwrap();
+    cores[1].sync(None).unwrap();
+
+    cores[1].create_link_at_path("/link1", folder1.id).unwrap();
+    cores[1]
+        .create_file("document1", folder1.id, FileType::Document)
+        .unwrap();
+
```
Parth, line 467: for your reference this is what @tvanderstad is currently unable to do in our lockbook shared folder

libs/core/tests/sharing_tests.rs
```diff
@@ -329,6 +329,213 @@ fn write_document_in_rejected_shared_folder_in_rejected_share_folder() {
...
+    cores[1].sync(None).unwrap();
+
+    let link = cores[1].create_link_at_path("/link1", folder1.id).unwrap();
+    cores[1].sync(None).unwrap();
+    cores[0].sync(None).unwrap();
+    let result = cores[0]
+        .create_file("document1", link.id, FileType::Document)
+        .unwrap_err();
+
+    assert_eq!(result.kind, CoreError::FileNonexistent);
```
Parth, line 443: shouldn't this be: `FileNotFolder`?

Parth: prob a bug

Parth: probably we're doing something in the wrong order

ad-tra: the sharer is the one who is trying to create a document under a link. The link doens't exist on their tree, it only exists for the sharee. So `FileNoneexistent` is reasonable?

libs/core/tests/sharing_tests.rs
```diff
@@ -329,6 +329,213 @@ fn write_document_in_rejected_shared_folder_in_rejected_share_folder() {
...
+    cores[1].sync(None).unwrap();
+
+    let result: Vec<Uuid> = cores[1]
+        .get_and_get_children_recursively(link.id)
+        .unwrap()
+        .iter()
+        .map(|f| f.id)
+        .collect();
+
+    assert::all_recursive_children_ids(&cores[1], folder1.id, &result);
```
Parth, line 537: I think you want to assert something more specific here. Maybe a subsequent test has this, but you want to make sure:
+ `document` shows up in the recursive listing of `/`
+ `document` shows up in the recursive listing of `folder`

Parth: You probably want a similar thing for non-recursive get children and list metadata with tweaks respectively.

libs/core/tests/sharing_tests.rs
```diff
@@ -742,6 +949,43 @@ fn delete_link_to_share() {
...
+    assert::all_pending_shares(&cores[1], &[]);
+
+    cores[1].delete_file(folder.id).unwrap();
+
+    cores[1].sync(None).unwrap();
+    cores[0].sync(None).unwrap();
+
+    assert::all_paths(&cores[0], &["/", "/folder/"]);
+
+    assert::all_pending_shares(&cores[1], &["folder"]);
```
Parth, line 987: an FYI, deleting the link puts it back into a pending status for you, as this test demonstrates. 

deleting the actual file unshares it from you, it should also get rid of the link. Travis thinks there is a test for this behavior.

Parth (changes requested):
Yeah you're on the right track for sure:
+ [x] address any outstanding comments
+ [x] get CI to be happy
+ [x] add some tests to make sure that all the path logic is also resolving links correctly, you want to make sure the sharee can do things like `create_at_path` across the boundary of shared files.

libs/core/tests/sharing_tests.rs
Parth: oh true fine

libs/core/tests/sharing_tests.rs
ad-tra: okay i removed link swapping on `read_document`

libs/core/tests/sharing_tests.rs
ad-tra: okay i deleted that test

libs/core/tests/sharing_tests.rs
ad-tra: yeah the test already exists on `edited_document` ``sync_service_share_tests`

libs/core/tests/sharing_tests.rs
Parth: you should confirm that there is a test for both sides of this, deleting the link as well as deleting the share

tvanderstad: deleting a share deletes a link (demonstrated by test `delete_accepted_share`) but deleting a file does not and need not unshare it, since non-deletion updates to deleted files are thrown away anyway

tvanderstad: I'll write some good docs on sync and force the team to read and discuss them when the editor is in a better state 😉

Parth (changes requested):
fix assert_children in tests

Parth (approved):

.gitignore
```diff
@@ -109,3 +109,4 @@ libs/editor/SwiftEditor/Libs/
 libs/editor/egui_editor/include
 
 clients/apple/lockbook.xcworkspace/xcshareddata/swiftpm/Package.resolved
+
```
steverusso, line 112:
```suggestion
```

libs/core/libs/test_utils/src/assert.rs
```diff
@@ -7,6 +7,7 @@ use lockbook_shared::file_metadata::FileType;
 use lockbook_shared::path_ops::Filter::DocumentsOnly;
 use lockbook_shared::staged::StagedTreeLikeMut;
 use lockbook_shared::tree_like::TreeLike;
+
```
steverusso, line 10:
```suggestion
```

libs/core/tests/link_tests.rs
```diff
@@ -14,13 +14,112 @@ fn assert_valid_list_metadatas(c: &Core) {
...
+
+    cores[0].sync(None).unwrap();
+    cores[1].sync(None).unwrap();
+
+    cores[1].create_link_at_path("link", document.id).unwrap();
+
+    assert_valid_list_metadatas(&cores[0]);
+    assert_valid_list_metadatas(&cores[1]);
+    assert_eq!(cores[1].get_by_path("/link").unwrap().id, document.id);
+}
```
steverusso, line 44:
```suggestion
}

```

libs/core/tests/link_tests.rs
```diff
@@ -14,13 +14,112 @@ fn assert_valid_list_metadatas(c: &Core) {
...
+    cores[0].sync(None).unwrap();
+    cores[1].sync(None).unwrap();
+
+    cores[1].create_link_at_path("link", folder.id).unwrap();
+
+    let folder1 = cores[1].create_at_path("link/folder/").unwrap();
+
+    assert::all_ids(&cores[1], &[roots[1].id, folder1.id, folder.id]);
+    assert::all_paths(&cores[1], &["/", "/link/", "/link/folder/"]);
+}
```
steverusso, line 122:
```suggestion
}

```

libs/core/tests/sharing_tests.rs
```diff
@@ -329,6 +329,223 @@ fn write_document_in_rejected_shared_folder_in_rejected_share_folder() {
     let result = cores[1].write_document(document.id, b"document content by sharee");
     assert_matches!(result.unwrap_err().kind, CoreError::InsufficientPermission);
 }
+#[test]
```
steverusso, line 332:
```suggestion

#[test]
```

steverusso (changes requested):

PR #1588, opened 2023-02-18
server: hmdb -> db-rs

Parth:
More code can be deleted once we actually delete hmdb (next release) for now we need it for the migration.

salient changes to look over (@tvanderstad):
+ [x] migration
+ [x] tree impls
+ [x] admin_validate_server

other tests: 
+ [x] @SmailBarkouch can you just run all our billing tests when you get a moment (sanity check)

### DIFF
--- a/server/server/src/billing/billing_service.rs
+++ b/server/server/src/billing/billing_service.rs
@@ -14,6 +14,7 @@ use crate::schema::Account;
 use crate::ServerError::ClientError;
 use crate::{account_service, RequestContext, ServerError, ServerState, FREE_TIER_USAGE_SIZE};
 use base64::DecodeError;
+use db_rs::Db;
 use hmdb::transaction::Transaction;
 use libsecp256k1::PublicKey;
 use lockbook_shared::api::{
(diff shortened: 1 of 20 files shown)


server/server/src/billing/billing_service.rs
```diff
@@ -45,37 +46,42 @@ fn lock_subscription_profile(
...
-        if current_time - account.billing_info.last_in_payment_flow < state.config.billing.millis_between_user_payment_flows {
-            warn!(?owner, "User/Webhook is already in payment flow, or not enough time that has elapsed since a failed attempt");
-
-            return Err(ClientError(ExistingRequestPending));
-        }
+    let mut db = state.index_db.lock()?;
+    let tx = db.begin_transaction()?;
+    let mut account = db
+        .accounts
+        .remove(&owner)?
```
tvanderstad, line 53: Should this `remove` be a `get`? There are a few ways for this function to early terminate; is the transaction automatically aborted or something like that?

Parth: [215f6d8](https://github.com/lockbook/lockbook/pull/1588/commits/215f6d8ce2369a02b2aaf3a69aa778be19c770ef)

server/server/src/file_service.rs
```diff
@@ -330,195 +348,188 @@ pub async fn get_document(
...
+                        "attempted to disappear a file, a sharee didn't have it shared"
                     );
                 }
             }
 
-            Ok(docs_to_delete)
-        })?;
+            // maintain index: file_children
+            let parent = *meta.parent();
+            if !db.file_children.remove(meta.parent(), &id)? {
```
tvanderstad, line 510: this doesn't seem to match the previous implementation, which also removed the set for `meta.parent()` entirely

tvanderstad (approved):

PR #1589, opened 2023-02-18
core: cleanup old logs after a successful migration

Parth:
if a migration occurs, cleanup any old log files that may exist, ignore any errors if a cleanup doesn't work.

Tested manually by me from 0.5.11 using cli

### DIFF
--- a/libs/core/src/repo/mod.rs
+++ b/libs/core/src/repo/mod.rs
@@ -64,10 +64,12 @@ impl CoreV3 {
 
             let mut path = PathBuf::from(&config.writeable_path);
             path.push("lockbook_core__repo__schema_v2__CoreV2");
+            // ignore if file is missing
             drop(remove_file(path));
 
             let mut path = PathBuf::from(&config.writeable_path);
             path.push("lockbook_core__repo__schema__CoreV1");
+            // ignore if file is missing
             drop(remove_file(path));
         }
 
--- a/libs/core/src/repo/mod.rs
+++ b/libs/core/src/repo/mod.rs
@@ -1,6 +1,8 @@
 use db_rs::{Db, LookupTable, Single};
 use db_rs_derive::Schema;
 use hmdb::log::Reader;
+use std::fs::remove_file;
+use std::path::PathBuf;
 
 use lockbook_shared::account::Account;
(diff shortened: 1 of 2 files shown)


libs/core/src/repo/mod.rs
```diff
@@ -57,6 +59,16 @@ impl CoreV3 {
...
+
+            drop(source);
+
+            let mut path = PathBuf::from(&config.writeable_path);
+            path.push("lockbook_core__repo__schema_v2__CoreV2");
+            drop(remove_file(path));
+
+            let mut path = PathBuf::from(&config.writeable_path);
+            path.push("lockbook_core__repo__schema__CoreV1");
+            drop(remove_file(path));
```
steverusso, line 71: Reason for `drop`? Is this some trick / idiomatic practice on an `io::Result<()>` instead of just `let _ = `?
If not, I'd include a comment why.

steverusso (changes requested):

libs/core/src/repo/mod.rs
Parth: oh I just think it's prettier, I've seen it reasonably often

libs/core/src/repo/mod.rs
```diff
@@ -57,6 +59,16 @@ impl CoreV3 {
 
                 tx.drop_safely()?;
             }
+
+            drop(source);
+
+            let mut path = PathBuf::from(&config.writeable_path);
+            path.push("lockbook_core__repo__schema_v2__CoreV2");
+            drop(remove_file(path));
```
Parth, line 67:
```suggestion
            // ignore if file is missing
            drop(remove_file(path));
```

libs/core/src/repo/mod.rs
Parth, line 71:
```suggestion
            // ignore if file is missing
            drop(remove_file(path));
```

Parth: [30b2521](https://github.com/lockbook/lockbook/pull/1589/commits/30b2521aeffaeb73355649146fc14e5905684261)

steverusso (approved):

Issue #1570, opened 2023-02-10
cli + path not resolving link


Parth:
potentially a dupe
```
parth@Parths-MacBook-Pro ~/D/l/l/l/c/l/shared (integrate-db-rs)> lockbook sync
  [0s] [####################] 3/3 (0s) Sync complete.
parth@Parths-MacBook-Pro ~/D/l/l/l/c/l/shared (integrate-db-rs)> lockbook new
✔ Select a parent directory · /thoughts/discussions/Krishma/P&K/
✔ Choose a file name: · test/
error: a file in path '/thoughts/discussions/Krishma/P&K/' is a document being treated as a folder
```
```
parth@Parths-MacBook-Pro ~/D/l/l/l/c/l/shared (integrate-db-rs) [47]> lockbook debug info
✔ Select a file · /thoughts/discussions/Krishma/P&K/
File {
    id: 305190eb-d7fd-4e85-80b3-407153a3919b,
    parent: 91242034-f7e8-45c7-affe-dd0983afb757,
    name: "P&K",
    file_type: Link {
        target: bf25e2b3-14ed-4cf0-a741-62ebf625ebce,
    },
    last_modified: 1675995068809,
    last_modified_by: "parth",
    shares: [],
}
```
```
parth@Parths-MacBook-Pro ~/D/l/l/l/c/l/shared (integrate-db-rs) [1]> lockbook new -p bf25e2b3-14ed-4cf0-a741-62ebf625ebce -n test/
Folder created.
```

ad-tra:
Hey @Parth given the recent changes to the cli, I wasn't abe to reproduce the bug. 
I created a folder `test` under a link (internal `lockbook` shared folder in my case) without an error.
```
➜  lockbook git:(share-finalization) ✗ lockbook new /lockbook/test/
➜  lockbook git:(share-finalization) ✗ lockbook list -r | grep test 
test/   
```

When poking around this issue, I encountered another bug that might be related. 
```
➜  lockbook git:(share-finalization) ✗ lockbook list /lockbook
➜  lockbook git:(share-finalization) ✗ 
```
listing the contents of `/lockbook` yields nothing. It should return the children in that folder: `adam/, parth/, etc `

PR #1615, opened 2023-02-27
cli: fixed link resolving bug

ad-tra:
context: cli 
fixes: #1570

### DIFF
--- a/clients/cli/src/list.rs
+++ b/clients/cli/src/list.rs
@@ -101,7 +101,9 @@ fn get_children(
 ) -> Result<Vec<FileNode>, CliError> {
     let mut children = Vec::new();
     for f in files {
-        if f.parent == parent {
+        let is_parent_link = core.get_file_by_id(parent).unwrap().is_link();
+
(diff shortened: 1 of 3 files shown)


Parth (changes requested):
as @steverusso mentioned this is fundementally a core bug, not a CLI bug, possibly will be resolved by #1583

ad-tra:
I've written the breaking test. Here is my understanding of the issue based on this use case 

```mermaid
graph BT
subgraph sharer 
r["root 2"] ---> f["folder"] --> fi["file"]
end
subgraph sharee
r1["root 1"] ---> l["link"]
end
```

When the sharer shares a `folder` with the sharee, who accepts it as a link, this is the current behavior/ generated file tree

```mermaid
graph BT
subgraph sharee
r1["root 1"] ---> l["link"] --> f["folder"] --> fi["file"]
end
```

But the expected behavior is that the folder node would be replaced with a link. 
```mermaid
graph BT
subgraph sharee
r1["root 1"] ---> l["link"] --> fi["file"]
end
```

PR #1617, opened 2023-03-01
apple: remove navigation link file tree

### LINKED ISSUE #1529: new file weirdness iOS
create a new file, which would appear on the list at the bottom of the list, you need to scroll slightly for it to animate to it.

### LINKED ISSUE #1533: Share file sheet is not updated with the new sharee after a share is made

SmailBarkouch:
Traditionally, we used `NavigationView` to keep track of the file tree structure on iOS. The problem with this implementation was that `NavigationView` was not supposed to be used dynamically in this way. This caused a myriad of bugs (which are resolved now by the new implementation). This PR removes `NavigationView` in place of just mutating a list. Gestures and animation were manually added back, although are not the exact same. 

In addition, I added a file path breadcrumb. Every time a user opens a new folder, the old folder appears in the breadcrumb. If the path becomes larger than the width of the iOS device, the breadcrumb will automatically scroll to the latest folder.

**Screenshot**:
<details>

simulator_screenshot_BE615BBA-9325-468B-B55E-2E8E231D8BE1
</details>

fixes #1533
fixes #1529

### DIFF
--- a/clients/apple/Shared/Stateful Logic/FileService.swift
+++ b/clients/apple/Shared/Stateful Logic/FileService.swift
@@ -11,58 +11,36 @@ class FileService: ObservableObject {
     
     // File Service keeps track of the children and parent being displayed on iOS. Since this functionality is not used for macOS, it is conditionally compiled.
 #if os(iOS)
-    @Published var parent: File? = nil
-    @Published var children: [File] = []
-    
     @Published var path: [File] = []
+    var parent: File? {
+        get {
+            path.last
+        }
+    }
+    
+    func childrenOfParent() -> [File] {
+        if(path.isEmpty) {
+            if let realRoot = root {
+                path.append(realRoot)
+            }
+        }
+        
+        return childrenOf(path.last)
+    }
     
     func upADirectory() {
         path.removeLast()
-        refreshChildrenAtParent(parent?.parent)
     }
     
     func intoChildDirectory(_ file: File) {
-        if let realParent = parent {
-            path.append(realParent)
-        }
-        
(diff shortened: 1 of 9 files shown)


clients/apple/Shared/Stateful Logic/FileService.swift
```diff
@@ -3,12 +3,68 @@ import SwiftLockbookCore
...
+    var successfulAction: FileAction? = nil
 
     @Published var root: File? = nil
     @Published var files: [File] = []
-    var successfulAction: FileAction? = nil
+    
+    // File Service keeps track of the children and parent being displayed on iOS. Since this functionality is not used for macOS, it is conditionally compiled.
+#if os(iOS)
+    @Published var parent: File? = nil
+    @Published var children: [File] = []
```
Parth, line 15: can we keep track of less state in general? Couldn't we use `childrenOf` from below as we did before?

clients/apple/iOS/FilePathBreadcrumb.swift
```diff
@@ -0,0 +1,61 @@
...
+                                    fileService.pathBreadcrumbClicked(file)
+                                }
+                            }, label: {
+                                Text(file.name)
+                            })
+                            .id(file)
+                        }
+                        
+                        if(fileService.path.last != file) {
+                            Image(systemName: "arrow.forward")
```
Parth, line 35: can we use chevron there?

Parth (changes requested):

Parth (approved):

PR #1628, opened 2023-03-07
server: remove implicitly deleted files from server share index

tvanderstad:
also adds validation to admin cli


server/server/src/file_service.rs
```diff
@@ -63,6 +70,17 @@ pub async fn upsert_file_metadata(
             }
         }
 
+        let all_files: Vec<ServerFile> = tree.all_files()?.into_iter().cloned().collect();
```
tvanderstad, line 73: cloned because tree needs to be no longer used because tree has a mutable reference to shared files which we mutate in this loop

Parth (approved):

PR #1650, opened 2023-03-15
apple: path search

### LINKED ISSUE #840: explore an integration with spotlight search
https://developer.apple.com/documentation/corespotlight

SmailBarkouch:
Current implementation:
<details>

Screenshot 2023-03-27 at 1 10 13 PM
</details>

fixes #840


Parth (changes requested):
+ [ ] match color scheme used in xcode (selected file)
- [ ] open selected file in file tree

clients/apple/macOS/TreeView/FileTreeView.swift
```diff
@@ -65,6 +75,23 @@ struct FileTreeView: NSViewRepresentable {
             }
         }
     }
+    
+    func expandAncestorsOfDocument(file: File) {
+        if(treeView.row(forItem: file) != -1) {
```
Parth, line 80: If I'm reading this correctly, if a file is presently invisible we wouldn't expand it's parent?

SmailBarkouch: This `if` statement will execute if a file visible on the tree. That being, that all its parents are expanded. In that case, we don't need to expand any of its parents, and only need to scroll to it if it is presently not visible.

Parth (approved):

PR #1651, opened 2023-03-15
apple: fix ipad state issues

### LINKED ISSUE #1103: on ipad multitasking loses some state
If you're on an ipad and you go from full screen to split screen with another app, it's fine, current document stays open.

When you go back it doesn't maintain current open document. Cursory investigation of app state does not show that state is incorrect.

SmailBarkouch:
fixes #1103


Parth (changes requested):
failed QA on call

SmailBarkouch:
Will fix at a later date.

Issue #1639, opened 2023-03-08
metrics problems


Parth:
+ [ ] active users shouldn't include people who are just pushing the welcome doc
+ [ ] usage by user is counting shared activity
+ [ ] documents is double counting by shared

PR #1660, opened 2023-03-26
core: solve metrics problems 

ad-tra:
fixes: #1639 
 - [x] active users shouldn't include people who are just pushing the welcome doc
 - [x] usage by user is counting shared activity
- [x] documents is double counting by shared
- [x] qa by looking at `localhost/metrics` and clean up impl


server/server/src/metrics.rs
```diff
@@ -235,28 +233,45 @@ pub fn get_user_info(
...
-        }
-    }
+fn is_user_active<T: lockbook_shared::tree_like::TreeLike<F = ServerFile>>(
+    tree: LazyTree<T>, ids: &[Uuid],
+) -> Result<bool, ServerError<MetricsError>> {
+    let root = ids
+        .iter()
+        .find(|id| {
+            let file = tree.find(id).unwrap();
+            file.parent() == file.id()
```
Parth, line 255: there is an `.root() -> bool`

server/server/src/metrics.rs
```diff
@@ -235,28 +233,45 @@ pub fn get_user_info(
...
+        })
+        .unwrap();
+    let root_creation_timestamp = tree.find(root)?.clone().file.timestamped_value.timestamp;
```
Parth, line 258: almost certainly you don't need to `clone` to `copy` the `u64`

server/server/src/metrics.rs
```diff
@@ -235,28 +233,45 @@ pub fn get_user_info(
...
+        .max_by(|x, y| {
+            x.file
+                .timestamped_value
+                .timestamp
+                .cmp(&y.file.timestamped_value.timestamp)
+        })
+        .unwrap()
+        .file
+        .timestamped_value
+        .timestamp;
```
Parth, line 272: I think `last_seen` is still the right thing to use, not the most recent upsert.

Parth (changes requested):

server/server/src/metrics.rs
```diff
@@ -235,28 +247,14 @@ pub fn get_user_info(
...
-                })?;
-
-                total_bytes += usage;
-            }
-
-            total_documents += 1;
-        }
-    }
+    let total_bytes = get_usage_helper(db, &owner.0)
+        .unwrap()
```
Parth, line 253: no unwraps

server/server/src/metrics.rs
```diff
@@ -235,28 +247,14 @@ pub fn get_user_info(
...
+        .iter()
+        .map(|f| f.size_bytes)
+        .sum();
+    let total_documents = db.owned_files.data().get(&owner).unwrap().len() as i64;
```
Parth, line 257: honestly these should just be 2 different functions at this point

Parth (changes requested):

Parth (approved):

PR #1686, opened 2023-04-03
apple: iOS and iPadOS search

SmailBarkouch:
- [x] iOS document and path search
- [x] iPadOS document and path search

fixes: #843


Parth (approved):
what's the diff between end_search and stop_current_search? seems like we never actually call stop_current_search

Issue #1676, opened 2023-03-31
save open files before sync and as they are closed


steverusso:
- [x] when closed
- [x] before sync

PR #1690, opened 2023-04-04
egui: save tabs when closing them

steverusso:
There is still room for improving the uiux around saving, for example:
* keep the tab open while saving and then report any error back there.
* still close the tab (currently the case) and then report any error via dialog.

But for now, this prevents one form of data loss.

Checks one box in https://github.com/lockbook/lockbook/issues/1676.

@parth UAT if you can.


Parth (approved):
code looks good

made a change switched tabs, while on other tab closed the editor, reopened the editor the change persisted

PR #1698, opened 2023-04-08
editor: canonical input

tvanderstad:


libs/editor/egui_editor/src/editor.rs
```diff
@@ -21,6 +22,7 @@ pub struct Editor {
 
     // state
     pub buffer: Buffer,
+    pub pointer_state: PointerState,
```
tvanderstad, line 25: pointer state captures aspects of cursor which are not subject to undo history

steverusso (approved):

PR #1702, opened 2023-04-08
infra: macos bump bundle version (releaser), detect lockfile changes (dev-tool)

Parth:
# ci lockfiles

We were detecting lock file changes by doing a `cargo check --locked`. If you run this locally with/without a `.lock` it will behave as expected. However we invoke this at the end of our CI, at which file, if a lockfile change had not been pushed, the lockfile would have been updated by any of the commands that we run prior to it. My first thought was to just check this first. But the way we check -- inside a rust program that's a part of the workspace -- makes this infeasible.

So then I thought about just extracting out the `cargo check --locked` into the yml file for the CI job, but this needlessly overlaps with the `clippy` call we'll do later.

So I ultimately ended up asserting that after CI is done `git diff` must be empty.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -43,8 +43,8 @@ jobs:
     - name: Server Logs
       if: always()
       run: cargo run -p lockbook-dev -- print-server-logs
-    - name: Check Lock File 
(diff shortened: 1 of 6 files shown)


Parth:
sample lockfile failure
image

PR #1714, opened 2023-04-13
cli: improve delete warning messages

### LINKED ISSUE #1712: Inaccurate folder children count
The children count in the warning delete message is off by 1.
For example the folder `daily` is empty as shown in the `list daily/` yet the console message counts 1 children. This is a simple issue that can be fixed by calling `get_children` instead of `get_and_get_children_recursively`

### Screenshot 
image

ad-tra:
fixes #1712

### DIFF
--- a/clients/cli/src/main.rs
+++ b/clients/cli/src/main.rs
@@ -6,6 +6,7 @@ mod imex;
 mod list;
 mod share;
 
+use std::cmp::Ordering::{Equal, Greater, Less};
 use std::fmt;
 use std::io::{self, Write};
 use std::path::PathBuf;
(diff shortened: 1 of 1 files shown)


steverusso (changes requested):
Actually while we're here, let's also fix the grammar so it says `child` for `1` and `children` for more than `1`.

clients/cli/src/main.rs
```diff
@@ -135,9 +135,14 @@ fn delete(core: &Core, target: &str, force: bool) -> Result<(), CliError> {
 
     if !force {
         let mut phrase = format!("delete '{}'", target);
-        if f.is_folder() {
-            let count = core.get_and_get_children_recursively(f.id)?.len();
-            phrase = format!("{phrase} and its {count} children")
+        let count = core
```
Parth, line 138: could also save the recursive get children call for if we're actually dealing with a folder

steverusso: Good call

clients/cli/src/main.rs
```diff
@@ -135,9 +136,19 @@ fn delete(core: &Core, target: &str, force: bool) -> Result<(), CliError> {
...
+                .get_and_get_children_recursively(f.id)
+                .unwrap_or_default()
+                .len() as u64
+                - 1;
+
+            phrase = match count.cmp(&1) {
+                Less => format!("{phrase}"),
+                Greater => format!("{phrase} and its {count} children"),
+                Equal => format!("{phrase} and its 1 child"),
+            };
```
steverusso, line 151: If this necessitated a `match`, I'd just say no op the the `Less` case, but it doesn't. This is simply just and `if {} else if {}`. No `cmp`ing or `Ordering` necessary.

steverusso (approved):

steverusso:
Just for posterity's sake on why [c376c31](https://github.com/lockbook/lockbook/pull/1714/commits/c376c3139faa21e6a8a14243c8d094bdc069e06c) failed:
image

steverusso (approved):

Issue #1297, opened 2022-08-20
usage caps enforcement


Parth:
+ change-doc-content or upsert should not allow you to increase your usage beyond your cap
+ should gracefully inform the user as they're running out of space
+ should assign a constant (small) space cost per metadata

SmailBarkouch:
We should also determine the data cap of a user using the expiration dates present on each payment profile.

Parth:
*should assign a constant (small) space cost per metadata* 

disregards their deleted status

PR #1738, opened 2023-04-19
server: usage caps enforcement

ad-tra:
fixes: #1297

### DIFF
--- a/libs/core/libs/shared/src/api.rs
+++ b/libs/core/libs/shared/src/api.rs
@@ -96,6 +96,7 @@ pub enum ChangeDocError {
     NotPermissioned,
     OldVersionIncorrect,
     DiffMalformed,
+    UsageIsOverFreeTierDataCap,
 }
 
 impl Request for ChangeDocRequest {
--- a/server/server/src/file_service.rs
+++ b/server/server/src/file_service.rs
@@ -1,4 +1,4 @@
-use crate::account_service::is_admin;
+use crate::account_service::{get_usage, is_admin};
 use crate::schema::ServerDb;
 use crate::ServerError;
 use crate::ServerError::ClientError;
@@ -49,6 +49,7 @@ pub async fn upsert_file_metadata(
 
         let mut tree = tree.stage_diff(request.updates.clone())?;
         tree.validate(req_owner)?;
+
         let mut tree = tree.promote()?;
 
         for id in tree.owned_ids() {
@@ -169,7 +170,6 @@ pub async fn change_doc(
     if request.diff.diff() != vec![Diff::Hmac] {
         return Err(ClientError(DiffMalformed));
     }
-
     let hmac = if let Some(hmac) = request.diff.new.document_hmac() {
         base64::encode_config(hmac, base64::URL_SAFE)
     } else {
@@ -177,6 +177,27 @@ pub async fn change_doc(
     };
 
     let req_pk = context.public_key;
+
+    //todo: put this in a function called authorization check. refine the logic and call it on upsert_file_metadata
+    let usage_metrics = get_usage(RequestContext {
+        server_state: context.server_state,
+        request: GetUsageRequest {},
+        public_key: context.public_key,
+    })
+    .await
+    .unwrap();
+
+    let current_usage = usage_metrics
+        .usages
+        .iter()
+        .map(|f| f.size_bytes)
+        .sum::<u64>();
+    let new_usage = current_usage + request.new_content.value.len() as u64;
+
+    println!("current_usage: {}\nnew_usage: {}", current_usage, new_usage);
+    if new_usage > usage_metrics.cap {
+        return Err(ClientError(UsageIsOverFreeTierDataCap));
+    }
 
     {
         let mut lock = context.server_state.index_db.lock()?;
--- a/libs/core/tests/sync_service_tests.rs
+++ b/libs/core/tests/sync_service_tests.rs
@@ -1,8 +1,27 @@
+use image::EncodableLayout;
 use itertools::Itertools;
+use lockbook_core::Core;
 use lockbook_shared::file::ShareMode;
 use test_utils::*;
 
 /// Uncategorized tests.
+
+#[test]
+fn over_data_cap() {
+    let core: Core = test_core_with_account();
+    let document = core.create_at_path("hello.md").unwrap();
+    let free_tier_limit = 1024 * 1024;
+    let content: Vec<u8> = (0..(free_tier_limit * 2))
+        .map(|_| rand::random::<u8>())
+        .collect();
+    core.write_document(document.id, content.as_bytes())
+        .unwrap();
+
+    core.sync(None).unwrap();
+
+    let metrics = core.get_usage().unwrap();
+    assert!(metrics.server_usage.exact > metrics.data_cap.exact);
+}
 
 #[test]
 fn test_path_conflict() {


ad-tra:
why are billing tests ignored? f
fyi: i edited `downgrade_denied` to reflect the server enforcement of usage caps, and ran it locally. it passes.

server/server/src/account_service.rs
```diff
@@ -125,44 +127,71 @@ pub fn username_from_public_key(
...
+        &mut db.metas,
+    );
+
+    match tree {
+        Ok(t) => {
+            let usages = get_usage(&t.to_lazy(), db.sizes.data(), None)?;
+            Ok(GetUsageResponse { usages, cap })
+        }
+        Err(_) => Err(ClientError(GetUsageError::UserNotFound)),
+    }
```
ad-tra, line 152: is there a better way to propagate this error? ideally it would be `ServerTree::new()?.to_lazy()` but that throws an error conversion exception.

server/server/src/account_service.rs
```diff
@@ -325,22 +355,36 @@ pub async fn admin_get_account_info(
...
+                account: AccountInfo {
+                    username: account.username,
+                    root,
+                    payment_platform,
+                    usage: usage_str,
+                },
+            })
+        }
+        Err(_) => Err(ClientError(AdminGetAccountInfoError::UserNotFound)),
+    }
```
Parth, line 387: this isn't what this error means, same as above

server/server/src/billing/billing_service.rs
```diff
@@ -292,20 +294,37 @@ pub async fn cancel_subscription(
...
+                    .iter()
+                    .map(|a| a.size_bytes)
+                    .sum();
+
+                if usage > FREE_TIER_USAGE_SIZE {
+                    debug!("Cannot downgrade user to free since they are over the data cap");
+                    return Err(ClientError(CancelSubscriptionError::UsageIsOverFreeTierDataCap));
+                }
+            }
+            Err(_) => return Err(ClientError(CancelSubscriptionError::UserNotFound)),
```
Parth, line 326: not what this error means, like above

server/server/src/file_service.rs
```diff
@@ -41,22 +45,41 @@ pub async fn upsert_file_metadata(
...
-
         for id in tree.owned_ids() {
             if tree.calculate_deleted(&id)? {
                 current_deleted.insert(id);
             }
         }
 
+        tree.validate(req_owner)?;
+
+        let new_usage = get_usage(&tree, db.sizes.data(), Some(&current_deleted))
```
Parth, line 69: yeah so this tree captures the files that are hypothetically going to be deleted if allowed (because it's a staged tree).

and `sizes` already knows what files are previously deleted (because it will have no sizes for those files).

So you can probably get rid of the extra optional argument (`Some(&current_deleted))`).

lmk if you think I've missed something.

server/server/src/file_service.rs
```diff
@@ -202,6 +221,23 @@ pub async fn change_doc(
         )?
         .to_lazy();
 
+        let old_usage = get_usage(&tree, db.sizes.data(), None)
```
Parth, line 224: you probably don't need to answer the question in this complicated way here, you can just determine whether the new file is smaller than the older file (if they're over the cap)

Parth (changes requested):

libs/core/tests/get_usage_tests.rs
```diff
@@ -1,12 +1,14 @@
...
 use lockbook_shared::file_like::FileLike;
-use lockbook_shared::file_metadata::FileType;
 use lockbook_shared::file_metadata::FileType::Folder;
+use lockbook_shared::file_metadata::{FileType, METADATA_FEE};
 use test_utils::*;
 
 #[test]
 fn report_usage() {
-    let core = test_core_with_account();
+    let core: Core = test_core_with_account();
```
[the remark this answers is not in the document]
ad-tra: without setting `Core` type explicitly, i don't get helpful completion support from vscode. Does that happen to you on intellij?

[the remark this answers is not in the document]
Parth: does not happen for me on intellij, p sure does not happen for @tvanderstad on vscode either fwiw

libs/core/tests/get_usage_tests.rs
```diff
@@ -126,5 +145,103 @@ fn usage_new_files_have_no_size() {
...
+    core.sync(None).unwrap();
+
+    assert::all_paths(&core, &["/", "/hello.md"]);
+}
+
+#[test]
+fn old_file_and_new_large_one() {
+    let core = test_core_with_account();
+    let document1 = core.create_at_path("document1.md").unwrap();
+    let content: Vec<u8> = (0..((FREE_TIER_USAGE_SIZE as f64 * 0.8) as i64))
```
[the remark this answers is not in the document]
ad-tra: yeah i was aware of some compression/encryption happening. hence i used this snippet found in earlier tests to get the true storage size after a file is encrypted and compressed. `local_encrypted` accounts for compression right?
```rust
    let hmac = core
        .in_tx(|s| {
            Ok(s.db
                .base_metadata
                .data()
                .get(&file.id)
                .unwrap()
                .document_hmac()
                .cloned())
        })
        .unwrap();
    let local_encrypted = document_repo::get(&core.get_config().unwrap(), &file.id, hmac.as_ref())
        .unwrap()
        .value;
```

server/server/src/account_service.rs
```diff
@@ -125,44 +127,71 @@ pub fn username_from_public_key(
...
 
 #[derive(Debug)]
 pub enum GetUsageHelperError {
     UserNotFound,
 }
 
-pub fn get_usage_helper(
-    db: &ServerDb, public_key: &PublicKey,
-) -> Result<Vec<FileUsage>, GetUsageHelperError> {
+pub fn get_usage<T>(
```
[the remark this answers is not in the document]
ad-tra: but the passed tree is a bit general. It can be a `ServerTree` or a `StagedTree`

server/server/src/file_service.rs
ad-tra: if i do that, the file metadata cost would be disregarded.

server/server/src/account_service.rs
```diff
@@ -147,22 +153,41 @@ pub enum GetUsageHelperError {
...
+        .ids()
+        .iter()
+        .map(|&&file_id| {
+            let deleted = &tree
+                .implicit_deleted
+                .iter()
+                .filter(|&x| x.1 == &true)
+                .map(|x| *x.0)
+                .collect::<HashSet<Uuid>>();
+
```
Parth, line 172: you gotta additionally filter by `db.owned_files` for this user. Otherwise you'll incur the storage costs of the lockbook shared folder.

server/server/src/file_service.rs
```diff
@@ -202,6 +221,23 @@ pub async fn change_doc(
         )?
         .to_lazy();
 
+        let old_usage = get_usage(&tree, db.sizes.data())
+            .unwrap_or_default()
+            .iter()
+            .map(|f| f.size_bytes)
+            .sum::<u64>();
+        let new_usage = old_usage + request.new_content.value.len() as u64;
```
Parth, line 229: this is not correct, you're double counting the old size + the new size.

Parth: imo you should probably, just here, calculate the 
`old_size = sizes.get(id) + METADATA_FEE`, and `new_size = request.new.len() + METADATA_FEE`

don't want to incur the cost of sizing all metadatas for no reason.

ad-tra: i still believe calculating the cost of all metadata is necessary. even if you're increasing the size of one document, you have to be aware of the total current usage to determine if you're going over the total usage cap.

libs/core/libs/shared/src/api.rs
```diff
@@ -198,6 +207,7 @@ pub struct FileUsage {
 #[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone)]
 pub enum GetUsageError {
     UserNotFound,
+    NoRootFile,
```
Parth, line 210: why would this happen?

server/server/src/account_service.rs
```diff
@@ -153,31 +153,41 @@ pub enum GetUsageHelperError {
...
-                .filter(|&x| x.1 == &true)
-                .map(|x| *x.0)
-                .collect::<HashSet<Uuid>>();
-
-            let file_size = match deleted.contains(&file_id) {
+        .find(|file_id| match tree.find(file_id) {
+            Ok(f) => f.is_root(),
+            Err(_) => false,
+        })
+        .ok_or(GetUsageHelperError::UserNotFound)?;
```
Parth, line 169: it doesn't make sense to return this error if there's no root imo

server/server/src/account_service.rs
```diff
@@ -153,31 +153,41 @@ pub enum GetUsageHelperError {
...
+
+    let result = ids
```
Parth, line 171: I think you're better off relying on `db.owned_files` for a set of ids this person owns

server/server/src/account_service.rs
```diff
@@ -153,31 +153,41 @@ pub enum GetUsageHelperError {
...
+        .iter()
+        .filter_map(|&file_id| {
+            if let Ok(file) = tree.find(&file_id) {
+                if file.owner() != tree.find(root).unwrap().owner() {
```
Parth, line 175: I know you checked for root's existence but this is still a bad idea over time

server/server/src/file_service.rs
```diff
@@ -221,12 +221,15 @@ pub async fn change_doc(
         )?
         .to_lazy();
 
-        let old_usage = get_usage(&tree, db.sizes.data())
+        let old_usage = get_usage_helper(&mut tree, db.sizes.data())
             .unwrap_or_default()
```
Parth, line 225: I'm not sure it makes sense to `unwrap_or_default` here, I think you can `?` and return an internal error in this unexpected case.

Parth (changes requested):

server/server/src/account_service.rs
ad-tra: that causes some headache with the borrow checker. owned files are borrowed by the tree. passing the tree and db.owned_files will clash. 
What's wrong with using the tree to check if a file is owned by the user or not?

Parth: you can clone an array of ids before you create the trees, it's sort of the single place we're trying to maintain the idea of what files you're responsible for paying for. When that model evolves it'll be easier to change the code if there aren't multiple places we're expressing the same idea in slightly different ways.

It also eliminates finding the root O(n) and scanning the files, unwrapping and ultimately allocating your own info.

My gut says readability and performance win if you just clone the set of ids out of the db for a user

Parth (changes requested):

server/server/src/account_service.rs
```diff
@@ -147,22 +152,54 @@ pub enum GetUsageHelperError {
...
+where
+    T: TreeLike,
+{
+    let ids = tree.owned_ids();
+    let root = ids.iter().find(|file_id| match tree.find(file_id) {
+        Ok(f) => f.is_root(),
+        Err(_) => false,
+    });
+
+    if root.is_none() {
```
Parth, line 167: you can either `ok_or` or `let-else` to get rid of the optional status. Then you won't be unwrapping below. From a previous comment: 
image

server/server/src/file_service.rs
```diff
@@ -32,6 +34,8 @@ pub async fn upsert_file_metadata(
         let db = lock.deref_mut();
         let tx = db.begin_transaction()?;
 
+        let usage_cap = get_cap(db, &context.public_key).unwrap_or_default();
```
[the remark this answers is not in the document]
Parth: most of these unwrap_or_defaults I see in this file should be situations where we should be paging someone, not silently passing the error and creating a zero'd out struct.

Parth (changes requested):

Parth (approved):

PR #1747, opened 2023-04-28
egui: auto save

steverusso:


clients/egui/src/account/mod.rs
```diff
@@ -99,6 +133,23 @@ impl AccountScreen {
...
+                        }
+                    }
+                }
+                AccountUpdate::SaveResult(id, result) => match result {
+                    Ok(time_saved) => {
+                        if let Some(tab) = self.workspace.get_mut_tab_by_id(id) {
+                            tab.last_saved = time_saved;
+                        }
+                    }
+                    Err(_err) => {}
```
Parth, line 151: intentional?

clients/egui/src/account/mod.rs
```diff
@@ -29,38 +32,69 @@ pub struct AccountScreen {
...
     update_rx: mpsc::Receiver<AccountUpdate>,
 
+    save_req_tx: mpsc::Sender<SaveRequest>,
+
     tree: FileTree,
     sync: SyncPanel,
     usage: Result<Usage, String>,
     workspace: Workspace,
     modals: Modals,
+    shutdown: Option<AccountShutdownProgress>,
```
Parth, line 42: potentially could just be a bool?

Parth (approved):

clients/egui/src/account/mod.rs
steverusso: There'll be other things we add that we'd want to track on shutdown, for example, if there's a sync currently happening, finish up, etc.

clients/egui/src/account/mod.rs
steverusso: Good catch, https://github.com/lockbook/lockbook/pull/1747/commits/8f72edc6e3f9e60015bca36f59bf6b80720acf2a

PR #1748, opened 2023-04-29
apple: suggested documents and general design improvements

### LINKED ISSUE #1749: add suggested documents

SmailBarkouch:
This PR added suggested documents (through the `suggested_docs` endpoint) to all the apple clients. In addition, the file document icons have changed to reflect the documents extension (`.md`, `.draw`, and regular files).

### iOS:
- Suggested documents will only appear when the user is in their root folder.
<details>

</details>

### iPad:
<details>

Simulator Screen Shot - iPad (10th generation) - 2023-04-29 at 20 49 50
</details>

### macOS:
- Suggested documents and the file tree are both collapsable but open by default.
- Dividers and spacing adjustments were made to the document and path search results view.
<details>

Screenshot 2023-04-29 at 8 41 41 PM
</details>

fixes #1749


clients/apple/Shared/Views/DocumentView.swift
```diff
@@ -13,60 +13,64 @@ struct DocumentView: View {
...
+                    case .Unknown:
+                        Text("\(meta.name) cannot be opened on this device.")
+                            .title(meta.name)
+                    }
                 }
             }
         }
-        
+        .onDisappear {
+            DI.files.refreshSuggestedDocs()
```
Parth, line 72: have you confirmed this is ever called?

clients/apple/iOS/SuggestedDocs.swift
Parth, line 1: is it possible that this UI element will load and then cause someone to click on the wrong file?

Parth (changes requested):
+ thoughts on making macOS look closer to iOS? I would even be down for iPad to look closer to iPhone (more compact)
+ did you checkout what order sync is called on initial load compared to things like suggested docs / list files?

clients/apple/Shared/Views/DocumentView.swift
SmailBarkouch: I have confirmed.

clients/apple/iOS/SuggestedDocs.swift
SmailBarkouch: On initial load it will have a temporary view with the same size as the real one, so it will not. Only case it may is if there are no suggested docs to begin with, so the loaded view is nothing.

Parth (approved):
looks sick, can't wait to release this

Issue #1701, opened 2023-04-08
shift enter


steverusso:
Shift enter should insert a new line after the cursor

Parth:
fwiw the default system behavior for me is that shift enter works the same way as just enter on linux, though I think what you're suggesting is better

steverusso:
How are you coming about that? Would like to try myself

tvanderstad:
after the cursor as in, at the end of the cursor's line?

steverusso:
No just after the cursor.

PR #1778, opened 2023-05-30
apple: markdown toolbar

SmailBarkouch:
An implementation of a markdown toolbar for all the apple clients.

The toolbar actions and their corresponding shortcuts are as follows:
+ Headings (1-4)
    + Heading 1: `Cmd+Shift+H`
+ Bold: `Cmd+B`
+ Italic: `Cmd+I`
+ Inline Code: `Cmd+C`
+ Bullet List: `Cmd+Shift+B`
+ Number List: `Cmd+Shift+N`
+ Todo List: `Cmd+Shift+T`
+ Indent and De-indent (only available on iOS and iPadOS)

## Screenshots
iOS:
<details>

Screenshot 2023-05-26 at 11 10 05 AM
</details>

iPadOS:
<details>

simulator_screenshot_174BB0E6-C3EB-43C4-A287-B5D8E0081A99
</details>

macOS:
<details>

Screenshot 2023-06-12 at 3 22 10 PM
</details>

fixes https://github.com/lockbook/lockbook/issues/1801
fixes https://github.com/lockbook/lockbook/issues/1701 (unintuitively related)

Finishes a part of #1809


tvanderstad:
Valiant effort! You were ultimately doomed to fail because of a quirk with the modification system.

When you insert text, the cursor positions for subsequent modifications later in the document are adjusted. This makes it easy to e.g. move the cursor to an earlier place in the editor, insert text, then put it back. The value of the cursor in the "put it back" location is automatically adjusted so that it's selecting the same text it was originally selecting, rather than the same range of characters, which is different because inserting text earlier in the document moves everything over.

One thing to note about this system is that the place you're putting the cursor is evaluated according the the editor's contents at the start of the frame, not according to the editors contents at the time the modification is applied. This means e.g. that if in one frame you insert a lot of text, then move the cursor to the start of the line, it will move the cursor to the start of what _was_ the line when the frame started rather than what the line is after all that text is inserted. This is confusing but necessary because we can't tell what the lines will look like without rendering them, which only happens at frame boundaries and not in between modifications applied in one frame.

In the system before I rewrote it here, there was no way to insert text at the cursor and then put the cursor to where it was. That's because, in the document before any modifications are applied, there's just one place where the cursor is and there's no way to distinguish that you want the cursor to the start of the text you inserted instead of the end. The cursor would just always go to the end so that when you're typing normally and a character is inserted, the cursor is updated to the spot after the cursor.

In my rewrite here, I add support for a flag for whether to advance the cursor when doing a text insertion/replacement. That means you can insert text where the cursor is (e.g. for the trailing `_` when applying italics to the selection) without advancing the cursor past the text (so you can leave the cursor between the two new `_`'s). The flag is also easily used to support shift+enter which adds a newline without advancing the cursor so I went ahead and implemented that too.

libs/editor/egui_editor/src/apple/ios_ffi.rs
```diff
@@ -418,3 +420,13 @@ pub unsafe extern "C" fn clipboard_paste(obj: *mut c_void) {
...
+
+/// # Safety
+/// obj must be a valid pointer to WgpuEditor
+#[no_mangle]
+pub unsafe extern "C" fn indent_at_cursor(obj: *mut c_void, deindent: bool) {
+    let obj = &mut *(obj as *mut WgpuEditor);
+    obj.editor
+        .custom_events
+        .push(Modification::Indent { deindent });
+}
```
tvanderstad, line 432: weird that this is in a different file, but I think we'll be in a better place to organize this when we see what's re-used by an android integration, which might re-use more of the ios_ffi.rs functions

libs/editor/egui_editor/src/input/canonical.rs
```diff
@@ -144,7 +150,9 @@ pub fn calc(
                 text: "".to_string(),
             })
         }
-        Event::Key { key: Key::Enter, pressed: true, .. } => Some(Modification::Newline),
+        Event::Key { key: Key::Enter, pressed: true, modifiers, .. } => {
+            Some(Modification::Newline { advance_cursor: !modifiers.shift })
+        }
```
tvanderstad, line 155: implements shift+enter

libs/editor/egui_editor/src/offset_types.rs
```diff
@@ -414,16 +414,17 @@ impl Debug for RelCharOffset {
     }
 }
 
-pub trait RangeExt<Element>: Ord + Sized {
+pub trait RangeExt<Element: Sub<Element>>: Ord + Sized {
```
tvanderstad, line 417: fun rust syntax for "a trait implemented for types which can be subtracted from themselves and which implies `Ord` and `Sized`"

libs/editor/egui_editor/src/offset_types.rs
```diff
@@ -438,6 +439,10 @@ where
         *max(&self.0, &self.1)
     }
 
+    fn len(&self) -> <T as Sub>::Output {
```
tvanderstad, line 442: fun rust syntax for "a function of `&self` which returns whatever would be returned by subtracting one instance of the inner element type from another" (e.g. `RelCharOffset` when trait is implemented for `(DocCharOffset, DocCharOffset)`)

libs/editor/egui_editor/src/apple/ios_ffi.rs
SmailBarkouch: This is in a different file since it is not used by macOS.

libs/editor/egui_editor/src/ast.rs
```diff
@@ -33,6 +33,21 @@ pub fn calc(buffer: &SubBuffer) -> Ast {
...
+        for i in 0..self.nodes.len() {
+            if self.nodes[i].range.contains(offset)
+                && self.nodes[i].range.len().0 < smallest_chosen_ast_range
+            {
+                chosen = i;
+            }
+        }
+
+        chosen
+    }
```
tvanderstad, line 49: did you mean to keep track of the smallest range?
```suggestion
    pub fn ast_node_at_char(&self, offset: DocCharOffset) -> usize {
        let mut chosen = 0;
        let mut smallest_chosen_ast_range = usize::MAX;

        for i in 0..self.nodes.len() {
            if self.nodes[i].range.contains(offset)
                && self.nodes[i].range.len().0 < smallest_chosen_ast_range
            {
                chosen = i;
                smallest_chosen_ast_range = self.nodes[i].range.len().0
            }
        }

        chosen
    }
```

libs/editor/egui_editor/src/editor.rs
```diff
@@ -253,12 +262,59 @@ impl Editor {
...
+        let mut cursor_in_bold = false;
+        let mut cursor_in_italic = false;
+        let mut cursor_in_inline_code = false;
+
+        let ast_node_idx = self
+            .ast
+            .ast_node_at_char(self.buffer.current.cursor.selection.start());
+        let ast_node = &self.ast.nodes[ast_node_idx];
+
+        match ast_node.element {
```
tvanderstad, line 279: If you want this to indicate that you're in a heading when you're in a bold heading, you'll likely need to look at styled chunks which have a list of styles associated with text ranges. I'm down to merge this how it is and I'll implement that after some cleanup

libs/editor/egui_editor/src/editor.rs
```diff
@@ -28,6 +29,14 @@ pub struct EditorResponse {
...
     pub edit_menu_x: f32,
     pub edit_menu_y: f32,
+
+    pub cursor_in_heading: bool,
+    pub cursor_in_bullet_list: bool,
+    pub cursor_in_number_list: bool,
+    pub cursor_in_todo_list: bool,
+    pub cursor_in_bold: bool,
+    pub cursor_in_italic: bool,
+    pub cursor_in_inline_code: bool,
```
tvanderstad, line 39: I'm not sure what should be in the editor response and what should be a function invoked via ffi but this is easier and looks clean enough

SmailBarkouch: Yes I tried to keep it clean. Since this struct is returned through ffi to Swift, I kept it simple and just a series of booleans. This is also 1 to 1 with how I represent what markdown element button is pressed on the swift side as well.

libs/editor/egui_editor/src/editor.rs
SmailBarkouch: Not exactly sure how that would be done, so i'm fine merging this as what it is, and you can implement it after some cleanup.

Issue #1268, opened 2022-08-13
investigating lb-fs


Parth:
Being able to mount a user's directory as a file system would be really cool. It would allow someone inside a program like microsoft word to open files inside lockbook. It would also give us another strategy for opening foreign files from within lockbook. 

As far as I can tell, there's 2 broad strategies for doing this:
1. watching a file system and a destination for changes and reconciling them 
2. exposing a virtual filesystem which is a source of truth

To me, 2 seems more appealing. We won't need to duplicate data, which is appealing because we don't want to leave (potentially all) files unencrypted on disk. Especially in the case of failures where the sync is interrupted (laptop battery dying).

Parth:
On Linux you can use [fuse](https://www.kernel.org/doc/html/latest/filesystems/fuse.html). There are rust libs that [help](https://github.com/zargony/fuse-rs). While it seems like there's macOS support I wonder if it's going to be too much to ask our users to install something like macFUSE.

There are some platform specific implementations that may be better suited for those platforms long term:
+ [Apple](https://developer.apple.com/documentation/fileprovider/nsfileproviderreplicatedextension) 
+ [This discussion around some windows alternatives](https://stackoverflow.com/questions/13932953/creating-virtual-folders-and-hooking-them-into-the-file-system).

Parth:
If we're able to create a rust library called `vfsrs` that abstracts across the platform specific implementations I think it would be a really compelling addition to the rust ecosystem.

PR #1805, opened 2023-06-08
drive: poc

ssapra2:
Fixes: #1268


ssapra2:
- Start with a CLI that allows people to import an account and specify a location on disk which the drive will keep in sync with lockbook
- Look at place on file system, understand change, and put into lockbook
- Call sync function which will actually finalize the change in servers
- Changes coming from both lockbook as well as the file

ssapra2:
- Implement Notify into CLI and print all the changes it detects.

ssapra2:
Two areas of investigation:
- Use File-locks and try to determine the level of respect that file locks get.
- Lock the directory and see if possible to interact with that directory with cli, vscode, app
- Look for windows computer - which graphical programs respect file lock
- Linux - see if you can get graphical access, if not someone else

- Reconciling changes from sync back onto the file system
- Just determine what changed for now

Parth:
superseded by https://github.com/lockbook/lockbook/pull/2343

PR #1806, opened 2023-06-09
egui: improve drag and drop with file tree

### LINKED ISSUE #1682: drag and drop with file tree
- [ ] beginning a drag outside the tree and bringing it over the tree with pick up the drag process
- [x] dropping files where they are changes their icons to folders
- [ ] starting a drag while scrolled down the file tree will immediately scroll the tree to the top

ad-tra:
fixes #1682
- [x] beginning a drag outside the tree and bringing it over the tree with pick up the drag process
- [x] dropping files where they are changes their icons to folders
- [x] starting a drag while scrolled down the file tree will immediately scroll the tree to the top


Parth (commented):
There could be some egui weirdness going on (maybe the reason I don't see what I'm looking for in the initial implementation either cc: @steverusso).

But `response` has a number of things associated with the present drag event:

`dragged`, `dragged_by`, drag_started_by`, `on_hover_and_drag_cursor`, (and about 5 more).

Is there a reason we're not using those?

I imagine at the very least we won't be doing things like: `is_hovered && ui.input(|i| i.pointer.primary_down())`

Parth:
There could be some egui weirdness surrounding it, if that's the case let's just document the weird behavior

ad-tra:
okay i replaced my hacky `is_hovered` with the response state `response.dragged()`

Parth (approved):

PR #1859, opened 2023-06-21
egui: sharing support

ad-tra:


Parth (approved):
pending style changes

PR #1890, opened 2023-07-05
egui: suggested docs

ad-tra:


clients/egui/src/account/suggested_docs.rs
```diff
@@ -0,0 +1,131 @@
...
+            .unwrap_or_default()
+            .iter()
+            .filter_map(|id| {
+                let file = core.get_file_by_id(*id);
+                if file.is_err() {
+                    return None;
+                };
+                let path = core.get_path_by_id(*id).unwrap_or_default();
+
+                Some(SuggestedFile { name: file.unwrap().name, path, id: *id })
```
Parth, line 29: I'm skeptical of all these `unwrap_or_default`s as well as the straight up `unwrap`

clients/egui/src/account/suggested_docs.rs
```diff
@@ -0,0 +1,131 @@
+use std::sync::Arc;
+
+use eframe::egui;
+
+use crate::model::DocType;
+
+pub struct SuggestedDocs {
+    files: Vec<SuggestedFile>,
+}
+struct SuggestedFile {
```
Parth, line 10:
```suggestion

struct SuggestedFile {
```

clients/egui/src/account/suggested_docs.rs
```diff
@@ -0,0 +1,131 @@
+use std::sync::Arc;
+
+use eframe::egui;
+
+use crate::model::DocType;
+
+pub struct SuggestedDocs {
+    files: Vec<SuggestedFile>,
```
Parth, line 8: you can prob choose a better name so that you avoid `self.files` which is a bit ambiguous in our project -- `self.recs` for instance is a lot more readable

clients/egui/src/account/mod.rs
```diff
@@ -217,8 +230,14 @@ impl AccountScreen {
                         }
                     }
                 }
-                AccountUpdate::FileDeleted(f) => self.tree.remove(&f),
-                AccountUpdate::SyncUpdate(update) => self.process_sync_update(ctx, update),
+                AccountUpdate::FileDeleted(f) => {
+                    self.tree.remove(&f);
+                    self.suggested = SuggestedDocs::new(&self.core);
```
Parth, line 235: it's kinda weird to reset it to a new suggested docs imo. I think it's also part of the reason you're having a weird error handling experience. I think you probably want SuggestedDocs to take a context and the tx and have a function that requests a recalc. The recalc ideally happens on a different thread, updates the state of SuggestedDocs and requests a new frame from egui.

Parth (changes requested):

clients/egui/src/account/mod.rs
ad-tra: is a separate thread necessary considering that suggested docs call is sub 20ms if i recall correctly (at least for my case, not a lot of files).

clients/egui/src/main.rs
```diff
@@ -23,7 +23,7 @@ fn main() {
     // following issues:
     //  1. window decorations are non-native.
     //  2. dragging & dropping from the system doesn't work.
-    std::env::set_var("WINIT_UNIX_BACKEND", "x11");
+    std::env::set_var("WINIT_UNIX_BACKEND", "wayland");
```
Parth, line 26: ?

clients/egui/src/account/tree/mod.rs
```diff
@@ -45,33 +45,30 @@ impl FileTree {
...
-                } else {
-                    self.draw_drag_info_by_cursor(ui);
-                }
-            } else if is_hovered && ui.input(|i| i.pointer.primary_down()) {
-                // todo(steve): prep drag only if a file is clicked
-                self.state.dnd.is_primary_down = true;
-                if ui.input(|i| i.pointer.is_moving()) {
-                    self.state.dnd.has_moved = true;
-                }
+        ui.spacing_mut().item_spacing = egui::vec2(0.0, 0.0);
```
Parth, line 48: Feels like un un-related change snuck in here?

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+	"cSpell.words": ["recalc"]
```
Parth, line 2: I think this needs to be a local only config for you

Parth (changes requested):

clients/egui/src/main.rs
ad-tra: oops, i'm gonna do what you said in standup and only hard set that var if it's not defined by the user. it's getting a bit annoying.

clients/egui/src/account/tree/mod.rs
ad-tra: no that change is intentional, got rid of the `ui.vertical` because it's not needed any more and expanded the tree to take full height. 

without changes:
image

with change:
image

.vscode/settings.json
ad-tra: true my fault.

Parth (approved):

PR #1921, opened 2023-07-14
apple: log cleanup

tvanderstad:


libs/editor/SwiftEditor/Sources/Editor/macMTK.swift
```diff
@@ -155,7 +155,6 @@ public class MacMTK: MTKView, MTKViewDelegate {
     func setClipboard(){
         let pasteboardString: String? = NSPasteboard.general.string(forType: .string)
         if let theString = pasteboardString {
-            print("clipboard contents: \(theString)")
```
tvanderstad, line 158: this one in particular was putting my private key in the logs when I sign in

Parth (approved):

Issue #1809, opened 2023-06-12
implement egui-editor toolbar


smailbarkouch:
- [x] apple
- [ ] android
- [x] egui

ad-tra:
@smailbarkouch doesn't android already have a toolbar?

Parth:
closed due to change in strategy (#2280)

PR #1952, opened 2023-07-19
egui: markdown toolbar

ad-tra:
related to #1809


clients/egui/src/account/mod.rs
```diff
@@ -487,6 +487,67 @@ impl AccountScreen {
         }
     }
 
+    pub fn reload_markdown_tabs(&self, ctx: &egui::Context) {
```
ad-tra, line 490: there's duplicate code between `reload_markdown_tabs` and `refresh_tree_and_workspace`, not sure if i should merge the two functions and add conditionals to avoid running unnecessary code (like refreshing the tree, when i only need to reload the tabs) or if i can tolerate the duplication

clients/egui/src/account/modals/mod.rs
```diff
@@ -66,6 +66,10 @@ impl super::AccountScreen {
                 use SettingsResponse::*;
                 match inner {
                     SuccessfullyUpgraded => self.refresh_sync_status(ctx),
+                    ToggleToolbarVisibility(_) => {
+                        self.modals.settings = None;
```
ad-tra, line 70: avoids a node none existent bug thrown by accesskit which forbids hiding elements if they're disabled. I enable the toolbar by disabling the settings modal and then I reload the tab. this has the negative effect of closing the settings prematurely.

clients/egui/src/account/mod.rs
Parth: Am I understanding correctly that you reload the tabs if the toolbar is toggled on or off?

If that's the case it feels like a very heavy weight way to get the toolbar to disappear. 

Ideally you would reach into each toolbar and toggle it's visibility and request a new frame, instead of re-reading all the files. 

let me know if I'm reading the code wrong

Parth (changes requested):

clients/egui/src/account/mod.rs
ad-tra: correct, the tab reloads after the toolbar vis change. I modified it so that visibility changes from within without reloading.

Parth (approved):

PR #1978, opened 2023-08-02
egui: file content search

ad-tra:
# Demo

https://github.com/lockbook/lockbook/assets/66345861/b3498c2a-657c-4443-a30c-4125c18a400e


libs/core/src/service/search_service.rs
```diff
@@ -400,6 +400,20 @@ impl<Client: Requester, Docs: DocumentService> CoreState<Client, Docs> {
...
 }
 
+impl SearchResult {
+    pub fn get_score(&self) -> Option<i64> {
+        match self {
+            SearchResult::FileNameMatch { id: _, path: _, matched_indices: _, score } => {
+                Some(*score)
+            }
+            SearchResult::FileContentMatches { id: _, path: _, content_matches } => {
+                Some(content_matches[0].score)
```
smailbarkouch, line 410: Choosing the first item to represent the score of the entire content match list is going to give somewhat inaccurate results.

clients/egui/src/account/full_doc_search.rs
```diff
@@ -0,0 +1,295 @@
...
+                        Self::show_file(ui, file, path, self.x_margin);
+                    }
+
+                    FileContentMatches { id, path, content_matches } => {
+                        let file = &core.get_file_by_id(*id).unwrap();
+                        Self::show_file(ui, file, path, self.x_margin);
+                        ui.horizontal(|ui| {
+                            ui.add_space(15.0);
+                            ui.horizontal_wrapped(|ui| {
+                                self.show_content_match(ui, &content_matches[0], 15.0);
```
smailbarkouch, line 162: Only the first content match item for a file will be shown this way. Ideally, we want to be able to see each content match for a file as their own search result, with their own score.

Parth (changes requested):

clients/egui/src/account/full_doc_search.rs
ad-tra: how about showing the content match with the highest score and ignoring the other content matches (from the same file).
 
The goal of "file content search" is not to locate and jump between text content in a certain file (like in a simple `<C, f>`), but to locate _the file path_. we display the body content match just for context info. 

If multiple entries are added per file, that might crowd out the file path search results

smailbarkouch: I don't think showing just the highest score, at least with our current implementation of "file content search," is necessarily the best option. I return multiple content search results so that users can peer through them and see whether a file contains the specific body of text they want. This is a step more important than just functioning as a context info for a _file path_.

Although I do certainly see what you mean, and this very well can crowd up search results with multiple matches from the same file(s). I also have thoughts on how search can be improved, and we can definitely come back to this, but I feel for the sake of consistency with other platforms, its best to separate the results.

libs/core/src/service/search_service.rs
smailbarkouch: Although you separated the file content matches, their scores are all still reliant on the first item's score. This will push up more irrelevant matches from the same file. You should make each file content match from the same file be sorted by its own score.

smailbarkouch (approved):

Parth (approved):
My only concern is that on startup, we'll call start_search which will grab that lock, let's see if it becomes a problem.

PR #2001, opened 2023-08-12
cli: v3, use cli-rs

### LINKED ISSUE #878: man page and completion files

### LINKED ISSUE #1693: lockbook list -r hard to parse
image

### LINKED ISSUE #1723: cli: no way to export account without exposing private key
I expect `lockbook account export | pbcopy` to put my private key on my clipboard, but instead it hangs.

`lockbook account export` prints this prompt (and awaits user input)
```
your private key is about to be visible. do you want to proceed? [y/n]:
```

I tried `yes | lockbook account export | pbcopy` but it puts `your private key is about to be visible. do you want to proceed? [y/n]: <private key>` on my clipboard

### LINKED ISSUE #1776: sharee username cli completion

Parth:
still wip

+ [x] document all the args, specifically the `target`s

## improvements
+ eliminated much of the clap "routing" boiler plate where we'd be routing variants to functions.
+ simplified error handling ergonomics & experience.
  + impl `Display` for error messages in core with handling of backtrace & unexpected.
  + eliminated the need for `Console` vs `Silent` mainly due to cli-rs.
  + generally nicer errors for the users (no more debug output).
+ fixed tiny issues listed below.
+ completions for all the things.

## regressions:
+ there were some instances where if an argument wasn't provided we would fallback to stdin, for example with account creation, we no longer do this. In part because cli-rs doesn't support optional arguments yet. (https://github.com/lockbook/cli-rs/issues/6) 
+ `copy` used to take a `Vec<PathBuf>` but cannot do that because `cli-rs` doesn't support var-args yet. This is unfortunate because it's the mechanism that enables globbing. So for now it just takes a single file to copy in and you can glob into a folder if you must. (https://github.com/lockbook/cli-rs/issues/6) 

## thoughts:
+ I have mixed feelings about referring to things in core via `lb::`, I think cli could be a place people are going to be copy pasting from and it'll be confusing for them to get errors on things like `lb::Uuid`. So either we don't do that, or we refer to core as the `lb` lib (maybe something steve suggested, can't remember).
+ kept the project structure largely the same, I don't like how I have to refer to the main fn to see where program is implemented. But I understand how one command per file is problematic (v1). I did however make the singular purpose of main.rs the cli-rs spec.

## issues closed:
+ fixes #1723
+ fixes #1693
+ fixes #878


ad-tra:
solid work, pretty hype!

some remarks:
- bring help and error message aesthetics up to par with clap 

- `lockbook completions <tab>` has no completions
- would be nice to get the supported enum variants as part of the error message
image
- would be better and more conventional to have completion entries separated by new lines. (i can do this)
image
image
- would be nice to have syntactic sugar that shortens some command likes v2 did. for example `lockbook ls` instead of `lockbook list`

Parth:
help aesthetics for my reference

Parth:
Some updated aesthetics
image

Parth:
>  would be better and more conventional to have completion entries separated by new lines. (i can do this)

I'm not sure I see this everywhere: 
image

But I do have plans to make the static completions even better (with descriptions) so I'll tackle that then

Parth:
added tab completions
image

Issue #1913, opened 2023-07-08
implement a new new file experience


smailbarkouch:
- [x] apple
- [x] egui
- [ ] android

ad-tra:
@smailbarkouch the new new file experience is broken due to workspace on macOS

smailbarkouch:
@ad-tra I already have it tracked through here: https://github.com/lockbook/lockbook/issues/2529

ad-tra:
> @ad-tra I already have it tracked through here: #2529

ah ok my b,  glad you're aware

smailbarkouch:
Oh wait, you're right. This is the right issue (just made it lol): https://github.com/lockbook/lockbook/issues/2558

Parth:
This is broken on apple, unchecking it there

PR #2018, opened 2023-08-30
egui: new file experience 

ad-tra:
equi portion to #1913 

# Demo

https://github.com/lockbook/lockbook/assets/66345861/442da289-d2b6-4e1f-bf9e-bedc88b19a2f


clients/egui/src/account/mod.rs
```diff
@@ -703,6 +685,45 @@ impl AccountScreen {
...
+            let focused_parent = if focused_parent.file_type == FileType::Document {
+                focused_parent.parent
+            } else {
+                focused_parent.id
+            };
+            core.get_children(focused_parent)
+                .unwrap()
+                .iter()
+                .for_each(|f| {
+                    let re = Regex::new(r"^untitled-(\d+)\.\w*$").unwrap();
```
Parth, line 708: I don't think this qualifies as a good usecase for regex. There's ways to do a lot of what I'm seeing here in core already: https://github.com/lockbook/lockbook/blob/master/libs/core/libs/shared/src/filename.rs#L66. Poke around for more examples.

clients/egui/src/account/workspace.rs
```diff
@@ -2,6 +2,7 @@ use std::time::Instant;
 
 use eframe::egui;
 use egui_extras::RetainedImage;
+use egui_winit::egui::TextEdit;
```
Parth, line 5: You don't want `egui_winit`, ties us to `winit`, see other `use` statements.

Parth (changes requested):

clients/egui/src/account/mod.rs
```diff
@@ -703,6 +711,44 @@ impl AccountScreen {
...
+                });
+
+            let result = core
+                .create_file(
+                    max.generate_next().to_name().as_str(),
+                    focused_parent,
+                    lb::FileType::Document,
+                )
+                .map_err(|err| format!("{:?}", err));
+            update_tx.send(AccountUpdate::FileCreated(result)).unwrap();
```
Parth, line 749: I don't think this is the right strategy. Files with different prefixes will increment the variant. So if you have `image-conflict-3.png` as the only file in a folder you're going to create a file called `untitled-3.md`. You're also never going to create a file called `untitled-0.md`.

IMO the strategy should be something like:
start with `untitled.md`, check for conflicts with siblings, until there is no conflict continue to generate the `.next()`.

Parth (changes requested):

clients/egui/src/account/mod.rs
ad-tra: okay i fixed it so that only file names `untitled` are considered. 

here's the proposed procedure: 
1. filter children of parent and collect a vector of files whose name is `untitled`
2. sort the vector based on variant 
3. traverse the vector until you find a gap for insertion either in the start, middle, or end. for example. 
   - `1,2,3,4` the gap is at the start, create `untitled.md`
   - `0,1,2,4` the gap is after the 2, create `untitled-3.md`

Parth (approved):

Issue #2016, opened 2023-08-26
dynamic syntax capture interferes with selection


tvanderstad:
<img width="1800" alt="Screenshot 2023-08-25 at 11 50 54 PM" src="https://github.com/lockbook/lockbook/assets/6198756/29c8c86c-e5ea-405f-8208-c21ba18e6b9a">
<img width="1800" alt="Screenshot 2023-08-25 at 11 50 58 PM" src="https://github.com/lockbook/lockbook/assets/6198756/02662aa1-1bc1-465c-8193-6008208b6852">

PR #2038, opened 2023-09-04
editor: record and use pointer drag start offset

tvanderstad:
fixes https://github.com/lockbook/lockbook/issues/2016

https://github.com/lockbook/lockbook/assets/6198756/634aeaa3-737b-4757-b304-5eadd122d565

### DIFF
--- a/libs/editor/egui_editor/src/input/canonical.rs
+++ b/libs/editor/egui_editor/src/input/canonical.rs
@@ -251,7 +251,7 @@ pub fn calc(
         Event::PointerButton { pos, button: PointerButton::Primary, pressed: true, modifiers }
             if click_checker.ui(*pos) =>
         {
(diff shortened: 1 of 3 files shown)


libs/editor/egui_editor/src/input/canonical.rs
```diff
@@ -251,18 +251,18 @@ pub fn calc(
         Event::PointerButton { pos, button: PointerButton::Primary, pressed: true, modifiers }
             if click_checker.ui(*pos) =>
         {
-            pointer_state.press(now, *pos, *modifiers);
+            pointer_state.press(now, *pos, click_checker.pos_to_char_offset(*pos), *modifiers);
```
tvanderstad, line 254: the fix is to convert the pointer position to the character offset when the pointer is pressed instead of when it's dragged/released

Parth (approved):

PR #2054, opened 2023-09-07
core/content pane: reorganize libs folder

### LINKED ISSUE #1941: reorganize libs folder
`core/` -- contain everything related to core
`content`/ -- contain everything that a content pane can show

ad-tra:
fixes #1941


Cargo.toml
```diff
@@ -6,9 +6,9 @@ members = [
 	"libs/core",
```
Parth, line 6: core's path will also be `core/core`. Additionally I see no changes to makefiles which I know will cause problems

Parth (changes requested):

Cargo.toml
Parth: releaser will also need to be updated

libs/content/editor/SwiftEditor/Package.swift
Parth, line 1: This location changing will require a change to xcode, will push that commit shortly

Parth (changes requested):

Parth:
+ [ ] let's make sure we do this here as well https://github.com/lockbook/lockbook/commit/283986c589d292efc53e65fb35a0e3cb0d8ebf25
+ [ ] fix fuzzer if needed

PR #2067, opened 2023-09-12
editor: blinking cursor

tvanderstad:
fixes https://github.com/lockbook/lockbook/issues/2068


Parth:
@tvanderstad can you resolve conflicts here, I'll make sure energy usage is low on macOS then merge it.

tvanderstad:
this doesn't actually work due to some tricky timing issues with redraw_after so if we want to implement this we'll need a different approach

Issue #2049, opened 2023-09-06
tab indenting bulleted lists (and other list types) results in unexpected behavior


smailbarkouch:
Tab indenting on an empty list item will produce a graphical glitch initially:
<img width="62" alt="Screenshot 2023-09-06 at 12 48 58 PM" src="https://github.com/lockbook/lockbook/assets/20663038/7d18d454-ddc1-4f06-8583-a904f8b7addc">

Then when a character is entered on the glitched list item:
<img width="55" alt="Screenshot 2023-09-06 at 12 49 39 PM" src="https://github.com/lockbook/lockbook/assets/20663038/c88d2e2e-77db-40fe-8eef-aaad89f9c5b8">

tvanderstad:
Are you able to reproduce this on master? I might have fixed it inadvertently with another PR 

https://github.com/lockbook/lockbook/assets/6198756/87e663c9-7f58-4177-bd5f-e3c61a49fa1b

tvanderstad:
Managed to reproduce, this only affects iOS

PR #2091, opened 2023-09-17
editor: handle tab input as indent

tvanderstad:
fixes https://github.com/lockbook/lockbook/issues/2049 (graphical glitch and non-indenting-behavior)

### DIFF
--- a/libs/editor/egui_editor/src/apple/ios_ffi.rs
+++ b/libs/editor/egui_editor/src/apple/ios_ffi.rs
@@ -22,6 +22,10 @@ pub unsafe extern "C" fn insert_text(obj: *mut c_void, content: *const c_char) {
         obj.editor
             .custom_events
             .push(Modification::Newline { advance_cursor: true });
+    } else if content == "\t" {
+        obj.editor
+            .custom_events
+            .push(Modification::Indent { deindent: false });
     } else {
         obj.raw_input.events.push(Event::Text(content))
     }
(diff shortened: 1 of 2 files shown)


libs/editor/egui_editor/src/galleys.rs
```diff
@@ -121,7 +121,8 @@ pub fn calc(
             // see https://github.com/lockbook/lockbook/issues/1983, https://github.com/emilk/egui/issues/3203
```
tvanderstad, line 121: there's movement on this issue and the egui bug is tagged for egui's 0.23 release https://github.com/emilk/egui/issues/3203#issuecomment-1717408018

libs/editor/egui_editor/src/apple/ios_ffi.rs
```diff
@@ -22,6 +22,10 @@ pub unsafe extern "C" fn insert_text(obj: *mut c_void, content: *const c_char) {
         obj.editor
             .custom_events
             .push(Modification::Newline { advance_cursor: true });
+    } else if content == "\t" {
```
tvanderstad, line 25: this might be better handled on the integration side but it works for now

tvanderstad: related: https://github.com/lockbook/lockbook/issues/2092

Parth (approved):

PR #2134, opened 2023-09-30
editor: fix tab regression

tvanderstad:


libs/editor/egui_editor/src/input/mutation.rs
```diff
@@ -237,6 +235,9 @@ pub fn calc(
                     } else {
                         continue; // only process list items
                     };
+                if !indentation_processed_galleys.insert(galley_idx) {
+                    continue; // only process each galley once
+                }
```
tvanderstad, line 240: first, because this was happening before other skips, any galley intersecting in the cursor ended up in this set...

libs/editor/egui_editor/src/input/mutation.rs
```diff
@@ -472,7 +472,7 @@ pub fn calc(
                 }
             }
 
-            if processed_galleys.is_empty() && !deindent {
+            if indentation_processed_galleys.is_empty() && !deindent {
```
tvanderstad, line 475: ...so this wasn't empty, so we wouldn't insert a tab.

libs/editor/egui_editor/src/input/mutation.rs
```diff
@@ -327,17 +328,13 @@ pub fn calc(
 
             // always iterate forwards when renumbering because numbers are based on prior numbers for both indent
             // and deindent operations
-            processed_galleys.clear();
```
tvanderstad, line 330: second, because we cleared this, only when you renumbered a galley would it register as a change and skip inserting the tab, so when you indented a bullet it would still insert a tab...

libs/editor/egui_editor/src/input/mutation.rs
```diff
@@ -353,6 +350,9 @@ pub fn calc(
                     } else {
                         continue; // only process indented galleys
                     };
+                if !renumbering_processed_galleys.insert(galley_idx) {
```
tvanderstad, line 353: ...now, renumber processing is handed by a separate set, with the original indentation processing set being referenced to determine whether to insert a tab

Parth (approved):

PR #2199, opened 2023-10-16
egui: improve running out of storage experience

### LINKED ISSUE #2156: upgrade experience on egui is poor
IMG_1015

+ should have better error handling
+ a button that guides you into the payment flow

reported by @MehrotraRahul

ad-tra:
fixes #2156

## New approach:
if sync fails with `UsageIsOverDataCap` or `UsageIsOverFreeTierDataCap`,  display the progress bar filled with red instead of the usual accent color. Note that the percentage of fill might not show the bar full, it will display the percentage based on the storage in server, and not local. 
Upwards of the red progress bar there will be an upgrade button (in lieu of the error message) that takes the user to `settings -> usage`

### Possible improvements: 
just a red status bar, and upgrade button might not be enough to information communicated to the user. They might still be confused of what went wrong:
- still display the run out of storage error somewhere in the sidepanel 
- leverage toasts to notify the user  that they run out of storage. 
- deffer communication to the settings modal, and display an error message inside the settings usage tab.

### DIFF
--- a/clients/egui/src/account/modals/mod.rs
+++ b/clients/egui/src/account/modals/mod.rs
@@ -19,7 +19,7 @@ pub use file_picker::FilePickerAction;
 pub use help::HelpModal;
 pub use new_file::{NewFileParams, NewFolderModal};
 pub use search::SearchModal;
-pub use settings::{SettingsModal, SettingsResponse};
+pub use settings::{SettingsModal, SettingsResponse, SettingsTab};
 
 use crate::widgets::ToolBarVisibility;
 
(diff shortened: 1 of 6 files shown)


Parth:
Are you averse to augmenting `get_usage` with information that's present local-only?

Parth (approved):
I think you're going to run into problems where a UI element is defined by 2 sources of truth (sync / get_usage) use the usage_service.rs I sent you as a starting point to solve this problem, after that, this is good to go

Issue #2145, opened 2023-10-04
egui doesn't compile on macOS


Parth:
due to
+ [ ] @smailbarkouch's title things
+ [ ] @ad-tra's pdf things

Parth:
superseded by https://github.com/lockbook/lockbook/pull/2280

PR #2200, opened 2023-10-16
windows: windows frame

tvanderstad:


Parth (approved):

Parth:
Some non blocking problems you're going to want to fix:
+ [ ] add the windows crate to our cargo workspace (cargo.toml in proj root)
+ [ ] handle the compilation experience for someone trying to do `cargo check` on the root. Right now it's broken (#2145) for other reasons, but don't add to the pile. See how the "windows installer" no-ops' itself out of existence when we're not on windows. You probably want to do something similar.

Issue #1939, opened 2023-07-18
large files timeout during sync


Parth:

Parth:
#2798 should close this but I want to test a bit

ad-tra:
I imported the a large file on macos and got the beach ball that froze up my laptop. importing a large file on windows crashes the app. So is this issue really closed?

Parth:
well this is a pretty sync / lb-rs focused issue, and I believe that is closed because if you wait out the beach ball the uploads to succeed where perhaps they would not before. I made #4553 to track the apple one. But you should make a windows issue with your crash, I don't see it in server notifications. I also heavily invested in the clients, not sure if that was the source of the crash (did you use drag and drop? was that the source of problems?). Consider trying it on the builds from master on account of #4544

PR #2220, opened 2023-10-29
core: pull & push docs in parallel

### LINKED ISSUE #1940: sync should fetch and push files in parallel

Parth:
fixes #1940


Parth:
If you `clone` the client instead of creating a new one, the multithreaded solution is not any faster.

The way we're creating a new one right now is incompatible with the fuzzer, but is also not nearly as fast as it could be.

Considering spawning a tokio task and modeling the problem differently. Probably only right here for now.

Parth:
I suppose it's worth investigating what the solution to #1939 will be before investing further in a http-based solution.

Parth:
This will probably require a fresh start

PR #2240, opened 2023-11-08
egui: fix hover states and add pointer hand for tabs

lanesawyer:
# What
- Fixes a variety of pointer hover states in the egui version
- Adds pointer cursor for tab hover

# How
- Stops the editor from setting the default cursor, which was overriding other cursor styling in the app (search input, setting, sync, and collapse button)
- Adds pointer styling for tabs when any hovering occurs

# Screenshots
image
image
image

### DIFF
--- a/clients/egui/src/account/workspace.rs
+++ b/clients/egui/src/account/workspace.rs
@@ -138,6 +138,12 @@ fn tab_label(ui: &mut egui::Ui, t: &mut Tab, is_active: bool) -> Option<TabLabel
                 lbl_resp = Some(TabLabelResponse::Renamed(str.to_owned()))
             }
         } else {
+            if resp.hovered() {
+                ui.output_mut(|o: &mut egui::PlatformOutput| {
+                    o.cursor_icon = egui::CursorIcon::PointingHand
+                });
+            }
+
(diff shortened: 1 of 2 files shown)


lanesawyer:
This touches the editor so I'm afraid my fix breaks a necessary pointer setting for another platform, so extra scrutiny there would be great.

ad-tra (commented):
Hey Vari, Welcome to the project and thanks for the submitting the PR! This looks good and almost ready to go, If you want you can incorporate the following related feature:

The on hover cursor for a checkbook item in the editor should ideally be `PointingHand`. You can call  `fn checkbox` in `EditorClickChecker` to calculate whether the pointer is within the bounds of a checkbox.  

fyi: Egui by default doesn't assign a `PointingHand` on button hover, which is unfortunate. By contrast we have implemented a custom `Button` that implements a `PointingHand` on hover which we use in a couple of places including the settings icon that you highlighted in the screenshot. 

The editor code base is mostly worked on by @tvanderstad , so he can review to ensure the change wouldn't disrupt other clients, but as far as the egui client is concerned this is ready to merge.

tvanderstad (approved):
Looks good from an editor perspective! Appreciate the fix!

Issue #2194, opened 2023-10-13
editor does not detect a bare command keystroke


Parth:
the command key is supposed to un-reveal markdown characters, however the apple integration presently has no way to detect _only_ when a modifier key is pressed.

So in effect on macOS we un-hide the md characters when someone does command+left arrow which is unexpected and weird.

tvanderstad:
I'm pretty sure this is all it takes on windows, apple clients should be similar https://github.com/lockbook/lockbook/pull/2200/files#diff-f726ab1a38ef3aa414710848427d055973cce8d0296697a98f11cf65d4b06b6fR212-R219

PR #2246, opened 2023-11-10
macos: modifier-only events to editor

tvanderstad:
fixes https://github.com/lockbook/lockbook/issues/2194

https://github.com/lockbook/lockbook/assets/6198756/54bd3e97-5933-4526-a851-3b79125b47d8

### DIFF
--- a/clients/apple/Shared/LockbookApp.swift
+++ b/clients/apple/Shared/LockbookApp.swift
@@ -4,7 +4,7 @@ import SwiftLockbookCore
 import BackgroundTasks
 
 #if os(macOS)
-import AppKit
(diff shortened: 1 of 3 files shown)


clients/apple/Shared/LockbookApp.swift
```diff
@@ -4,7 +4,7 @@ import SwiftLockbookCore
 import BackgroundTasks
 
 #if os(macOS)
-import AppKit
```
tvanderstad, line 7: was getting a weird compilation error here, something to do with invalid characters

Parth (approved):

Issue #2152, opened 2023-10-06
multiple co-operative lb-rs instances 


Parth:
"be able to edit multiple notes side by side"

requested by @Luca409

I think this is something that is pretty easy with swiftui that we disabled. We could do something like single click opens in a new tab, double click opens in a popped out window.

note for smail:
In the past I was wondering whether the tab strip should get moved into egui, but I think this feature request seals the deal that these bits should be managed by the host application.

Parth:
Also requested by @Krishma-Bawa

Parth:
dupe of #2207

Parth:
This may be done on Apple thanks to #3853

Parth:
Tracking some issues that @smailbarkouch has discovered:
- workspace persistence in multiple windows will just use whichever window was the last to modify it's tab state. Seems fine to ship to me, but will probably require slightly deeper thinking for a long term solution (window id?)
- workspace presently disregards updates to documents made by `Actor::Workspace` which makes sense, but probably needs to be improved a bit. On init workspace can just hang on to a UUID and ignore updates specifically from the same UUID.

Parth:
The future is not yet, I don't like some of the behavior of this stuff, will be tweaking next release

Parth:
this is about to be a thing after #5078 what will remain after that is merged:
- [ ] cleaup how docs are handled in server and lb-rs
- [ ] populate keychain via db notifications
- [ ] forward db notifications to subscription stream

Issue #2104, opened 2023-09-18
egui tab refinements


tvanderstad:
Clicking a file should open a tab, then clicking another file should open a new tab in place of the first tab.

Double-clicking a file should open a tab, then clicking another file should open a new tab and leave the first one open.

related: https://github.com/lockbook/lockbook/issues/2103

Issue #2103, opened 2023-09-18
apple tab refinements


tvanderstad:
Clicking around in my files to browse them shouldn't result in many open tabs.

Clicking a file should open a tab, then clicking another file should open a new tab in place of the first tab.

Double-clicking a file should open a tab, then clicking another file should open a new tab and leave the first one open.

related: https://github.com/lockbook/lockbook/issues/2104
inspiration from vscode:

https://github.com/lockbook/lockbook/assets/6198756/90f1471d-ab6d-4ab1-be99-22bfda912f5b

Parth:
alternative acceptable ux would be to be able to close multiple tabs

Parth:
imo dupe of https://github.com/lockbook/lockbook/issues/2104

PR #2280, opened 2023-11-20
lb-workspace

Parth:
# Release blockers

+ [ ] horizontal scroll event
+ [x] egui app not initial syncing -- @ad-tra 
+ [x] Ugly -- @ad-tra 
+ [ ] Possible focus issue -- @tvanderstad / @ad-tra 
+ [x] Integration not sent to swift -- @Parth 
+ [x] Workspace current file not sent to swift -- @Parth 
+ [x] Sync status not sent to swift -- @Parth 
+ [ ] automatic renaming of files not working
+ [ ] settings hiding toolbar regression on egui

> its hard to decipher that the title of the page is the "title" bc it's the same font and styling as the rest of the text

# Merge blockers

+ [x] iOS integration -- @smailbarkouch 
+ [ ] Apple compatible pdf solution -- @Parth 

# Content Pane

The editor experiment has been a success, and this PR further extends the gains made by bringing more egui widgets to Apple.

+ We want to support many file types, and do not want to implement a custom editor for each file type on each platform. We want to implement it once and bring it to many places. Egui has already implemented support for the following file types:
  + Drawing
  + PDFs
  + Images
  + Plain Text
+ We also need to reduce the amount of duplicated effort we're performing. Quickly implementing these areas on various platforms served as a good proof of concept. But there are bugs and we can't afford the time to fix those bugs and unify the experience and create more complicated experiences for every platform.
  + tab strips
  + toolbars & shortcuts
  + auto save & auto sync

# Contract

Clients just tell content pane about file tree interactions (this UUID was clicked).

The following happens within content pane:
+ determining how to open a file and what to show
+ saving data
+ initial file names
+ determining when to kick off a sync
+ determining what content needs to be reloaded
+ tabs
+ toolbars

Content pane just needs to report back what file is currently selected (for nice file tree interactions). It can additionally, in the future when clients are ready, report the various tabs that are open and their dirtiness.

Content pane additionally probably needs to report back regarding whether it wants the software keyboard (and or styluses, etc).

# Considerations

We've had lots of conversations about the balance between native & cross platform UIs. The piece of this that gives me some pause is the tab strip. And primarily how it relates to #2152. However I think our current approach and it's future is on par with what we have and can achieve with egui. I do however really like the way macOS (not SwiftUI) native tabs feel, and would consider interacting with those in some future iteration of this work.

However, an important performance consideration, we don't want to initialize a content pane for every tab. We want mutually exclusive tabs (tabs on the same window) to share an underlying graphical instance. And only have multiple graphical instances for unique windows. I think this would require a medium-large sized technical lift no matter what approach we took. For this we may be in the territory of managing our own drag and drop events across two egui instances.

However, before we can even think about doing that, we need to more clearly understand what flow of editing the same note in two different windows would be. And so I think we can deliver 85% of the value of #2152 by deploying the following UX in the medium term:

You click on a note in a file tree, it opens it in a tab in the main window, whether that's a transient tab (#2104 / #2103) or not doesn't matter. You double click, and it will pop open a different window with just that note, you can tile it using your native window manager or whatever. If that note was already present in a tab you would be popping it out. Potentially there could be a pop out / pop in button or keyboard shortcut in either place.

The only workflow this solution is not compatible with is: viewing the same note side by side. But otherwise I think this is going to be our local optima for: team productivity, overall quality / reliability, and achieving the essence of that particular feature request.

# vague status

+ [x] pane
+ [x] pane-ffi
+ [x] eliminate ffi out of editor
+ [x] rework infra to reference "pane" instead of "editor"
+ [x] conform to contract inside rust and egui 
+ [x] conform to contract inside apple
+ [x] certainly a good amount of iOS work
+ [x] pdfium macos
+ [x] pdfium iOS
+ [x] possibly stylus events


tvanderstad:
Can you clarify what’s in the MVP vs what’s not? Checklist here is in the right direction.

Can toolbars be done as a later step? Surely multiple windows can be done as a later step as well - at least how Windows is looking, multiple window support is a sizable lift.

clients/apple/macOS/FileListView.swift
```diff
@@ -71,58 +73,58 @@ struct FileListView: View {
...
-                
-                Button(action: {
-                    currentSelection.isPendingSharesOpen = true
-                }) {
-                    pendingShareToolbarIcon(isPendingSharesEmpty: share.pendingShares.isEmpty)
-                        .imageScale(.large)
-                }
-            }
-        }
+//        .toolbar {
```
Parth, line 89: @smailbarkouch can you restore this

libs/content/workspace-ffi/SwiftWorkspace/Sources/Editor/macMTK.swift
```diff
@@ -0,0 +1,328 @@
...
+           let url = createTempDir() {
+            let imageUrl = url.appendingPathComponent(String(UUID().uuidString.prefix(10).lowercased()), conformingTo: .png)
+            
+            do {
+                try data.write(to: imageUrl)
+            } catch {
+                return false
+            }
+            
+            //            if let lbImageURL = workspaceState!.importFile(imageUrl) {
```
Parth, line 181: what's happening here @smailbarkouch?

libs/content/workspace-ffi/SwiftWorkspace/Sources/Editor/macMTK.swift
```diff
@@ -0,0 +1,328 @@
...
+            //                paste_text(wsHandle, lbImageURL)
+            //                workspaceState?.pasted = true
+            //
+            //                return true
+            //            }
+        } else if isDropOperation {
+            if let data = pasteBoard.data(forType: .fileURL) {
+                if let url = URL(dataRepresentation: data, relativeTo: nil) {
+                    //                    if let markdownURL = workspaceState!.importFile(url) {
```
Parth, line 190: also here @smailbarkouch

libs/content/workspace/src/workspace.rs
```diff
@@ -270,44 +221,43 @@ impl super::AccountScreen {
...
                 egui::Stroke { color: ui.visuals().widgets.active.bg_fill, ..Default::default() };
             ui.visuals_mut().widgets.hovered.fg_stroke =
                 egui::Stroke { color: ui.visuals().widgets.active.bg_fill, ..Default::default() };
-            if Button::default().text("New folder").show(ui).clicked() {
-                self.update_tx
-                    .send(OpenModal::NewFolder(None).into())
-                    .unwrap();
-                ui.ctx().request_repaint();
-            }
+            // if Button::default().text("New folder").show(ui).clicked() {
```
Parth, line 230: parth restore this

libs/content/workspace/src/workspace.rs
```diff
@@ -410,123 +358,321 @@ impl super::AccountScreen {
...
+            .name;
+
+        let fpath = self.core.get_path_by_id(id).unwrap(); // TODO
+
+        self.open_tab(id, &fname, &fpath, is_new_file);
+
+        let core = self.core.clone();
+        let ctx = self.ctx.clone();
+
+        // todo
```
Parth, line 416: restore this, but not for macOS

libs/lb/lb_external_interface/src/c_interface.rs
```diff
@@ -731,10 +731,32 @@ pub unsafe extern "C" fn time_ago(time_stamp: i64) -> *const c_char {
 /// Be sure to call `release_pointer` on the result of this function to free the data.
 #[no_mangle]
 pub unsafe extern "C" fn get_core_ptr() -> *mut c_void {
-    let obj = static_state::get().expect("Could not get core");
+    let obj = static_state::get().expect("Could not get core").core;
     Box::into_raw(Box::new(obj)) as *mut c_void
 }
 
+struct CameraHandle(*mut c_void);
```
Parth, line 738: parth remove this

libs/content/workspace-ffi/SwiftWorkspace/Sources/Editor/macMTK.swift
smailbarkouch: Weird, I already fixed all of paste import. I think someone forced pushed it out of the history, probably when we were trying to fix merge conflicts.

clients/apple/Shared/Stateful Logic/StatusService.swift
```diff
@@ -20,17 +20,18 @@ class StatusService: ObservableObject {
...
-            DispatchQueue.main.async {
-                switch localChanges {
-                case .success(let local):
-                    self.work = local.count
-                case .failure(let err):
-                    DI.errors.handleError(err)
-                }
-            }
-        }
+        // this whole file needs to be deleted
```
Parth, line 23: Delete this whole file eventually

clients/apple/Shared/Stateful Logic/SyncService.swift
```diff
@@ -113,7 +92,7 @@ class SyncService: ObservableObject {
     }
     
     func sync() {
-        if syncing {
+        if true {
```
Parth, line 95: Delete this function

PR #2285, opened 2023-11-27
apple: increase debounce for path search

### LINKED ISSUE #2282: search performance
+ make it better utilize hardware
+ better debounce on swift

smailbarkouch:
fixes #2282

Our previous debounce timer would wait `200 ms` before starting a search. This would cause a lot of searches to kick off while the user is typing, and since searches cannot happen concurrently, each search would have to wait for the subsequent search to complete (due to the mutex on core). 

So I increased the debounce duration to `550 ms`. This makes searching more forgiving for slow types, although the problem won't truly be fixed unless we can interrupt searches, or allow concurrent searches.

I also did a small refactor.

### DIFF
--- a/clients/apple/Shared/Stateful Logic/SearchService.swift
+++ b/clients/apple/Shared/Stateful Logic/SearchService.swift
@@ -132,7 +132,7 @@ class SearchService: ObservableObject {
             }
         }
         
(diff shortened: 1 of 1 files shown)


smailbarkouch:
Not needed anymore (we are going to improve cancelation speed and reduce debounce).

PR #2287, opened 2023-11-28
apple: incorporate document activity in search results

### LINKED ISSUE #1832: incorporate document activity in search results
When I search for my lockbook todo file, the results are always:

1. parth's todo list
2. raayan's todo list
3. smail's todo list
4. travis's todo list

Since I open and modify my own todo list all the time, I'd like that one to come up first, like it comes up in my suggested documents

smailbarkouch:
fixes #1832

### DIFF
--- a/libs/lb/lb-rs/src/service/activity_service.rs
+++ b/libs/lb/lb-rs/src/service/activity_service.rs
@@ -31,7 +31,7 @@ impl<Client: Requester, Docs: DocumentService> CoreState<Client, Docs> {
         Ok(())
     }
 
-    fn normalize(&mut self, docs: &mut [DocActivityMetrics]) {
+    pub(crate) fn normalize(&mut self, docs: &mut [DocActivityMetrics]) {
         let read_count_range = StatisticValueRange {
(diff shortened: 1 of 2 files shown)


libs/lb/lb-rs/src/service/activity_service.rs
```diff
@@ -31,43 +33,61 @@ impl<Client: Requester, Docs: DocumentService> CoreState<Client, Docs> {
         Ok(())
     }
 
-    fn normalize(&mut self, docs: &mut [DocActivityMetrics]) {
+    pub(crate) fn normalize_activity_metrics(
```
Parth, line 36: Not clear to me why you turned this into a `HashMap` the resulting code seems worse

smailbarkouch: I used a `HashMap` since it allowed me to get ids in quickly in search service, instead of sequentially searching for them. But on second thought, I can just turn the original `Vec` into a `HashMap` within search service, and make it a part of `SearchableFileInfo`.

PR #2291, opened 2023-12-09
core, apple: improved path and content search

### LINKED ISSUE #2216: search fails with "no search lock"

### LINKED ISSUE #2282: search performance

smailbarkouch:
fixes #2282
fixes #2216

# Changes

Significantly improved the speed of path and content search. In addition, there were small UI changes to the apple platforms for path search.

### DIFF
--- a/clients/egui/src/account/full_doc_search.rs
+++ b/clients/egui/src/account/full_doc_search.rs
@@ -147,7 +147,7 @@ impl FullDocSearch {
 
     fn send_search_results(&mut self, core: &lb::Core) {
         if let None = self.search_channel {
-            self.search_channel = Some(core.start_search(SearchType::PathAndContentSearch).unwrap());
+            self.search_channel = Some(core.start_search(SearchType::PathAndContentSearch));
         }
 
         let search_tx = self.search_channel.as_ref().unwrap().search_tx.clone();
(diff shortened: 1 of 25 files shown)


clients/egui/src/account/full_doc_search.rs
```diff
@@ -123,12 +145,17 @@ impl FullDocSearch {
         .inner
     }
 
-    fn send_search_results(&self) {
-        let search_tx = self.search_channel.search_tx.clone();
+    fn send_search_results(&mut self, core: &lb::Core) {
+        if let None = self.search_channel {
+            self.search_channel = Some(core.start_search(SearchType::PathAndContentSearch).unwrap());
```
ad-tra, line 150: The core call is expensive and blocks the ui thread causing the app to freeze momentarily when the user starts searching. It's a bit tricky to move this to a thread given that the result is stored in state. Other than this, the rest of changes on egui look good. 

some ui improvements to keep in mind, these are issues, I'll handle in the future:
- [ ] disallow search results from resizing the width of the sidebar
- [ ] throttle the spinner widget. you made search so fast that we're suffering from success lol. because it's so fast the spinner shows up for 200 ms or less and then flashes out of the screen. To remove this jarring effect, throttle the rendering of the spinner.

Parth (approved):

PR #2299, opened 2023-12-30
windows: event loop refinements

tvanderstad:
Fixes issues with cpu/energy consumption, reducing baseline from ~10% to ~1%. Redraw when necessary instead of redrawing continuously at max speed.

### DIFF
--- a/clients/windows/src/window.rs
+++ b/clients/windows/src/window.rs
@@ -14,12 +14,16 @@ use raw_window_handle::{
     HasRawDisplayHandle, HasRawWindowHandle, RawDisplayHandle, RawWindowHandle, Win32WindowHandle,
     WindowsDisplayHandle,
 };
-use std::time::{Duration, Instant};
+use std::time::Instant;
 use windows::{
     core::*, Win32::Foundation::*, Win32::Graphics::Direct3D12::*, Win32::Graphics::Dxgi::*,
-    Win32::System::LibraryLoader::*, Win32::System::Ole::*, Win32::UI::HiDpi::*,
-    Win32::UI::Input::KeyboardAndMouse::*, Win32::UI::WindowsAndMessaging::*,
+    Win32::Graphics::Gdi::*, Win32::System::LibraryLoader::*, Win32::System::Ole::*,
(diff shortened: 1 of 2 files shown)


clients/windows/src/window.rs
```diff
@@ -172,32 +169,18 @@ pub fn main() -> Result<()> {
...
-
-        // target framerate
-        let frame_period = Duration::from_micros(8333);
-        let now = Instant::now();
-        let elapsed = now - last_frame;
-        if elapsed < frame_period {
-            std::thread::sleep(frame_period - elapsed);
+    // "If the function retrieves the WM_QUIT message, the return value is zero."
+    // https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getmessagea
+    while unsafe { GetMessageA(&mut message, None, 0, 0) }.into() {
```
tvanderstad, line 174: PeekMessageA is non-blocking and best suited for games where you want max framerate. GetMessageA is blocking so nothing's happening when the UI is idle. It also makes the code simpler.

clients/windows/src/window.rs
```diff
@@ -294,30 +277,43 @@ fn handle_message(hwnd: HWND, message: Message) -> bool {
...
+                            unsafe { InvalidateRect(hwnd, None, false) };
                             input::mouse::handle_wheel(app, message, delta)
                         }
                         MessageAppDep::SetCursor => output::cursor::handle(),
                         MessageAppDep::Paint => {
+                            // "you'll find that your UI thread starts to burn 100% cpu core and your WM_PAINT handler
+                            // getting called over and over again... WM_PAINT is generated as long as the window has a
+                            // dirty rectangle, created by an InvalidateRect() call by either the window manager or
+                            // your program explicitly calling it. BeginPaint() clears it."
+                            // https://stackoverflow.com/questions/5841299/difference-between-getdc-and-beginpaint
```
tvanderstad, line 314: This is the magic ingredient I was missing (it's a code comment instead of a review comment! growth 🌱)

Parth (approved):

PR #2306, opened 2024-01-01
iOS, iPadOS: fix checkboxes

### LINKED ISSUE #2301: cannot reliably check checkboxes
`UITextInteraction` is likely consuming touch events and not letting them pass to the editor.

smailbarkouch:
fixes #2301 

# Details

Two gestures that are a part of `UITextInteraction`, which are `UITextMultiTapRecognizer` and `UITapAndAHalfRecognizer`, were consuming single tap events. These gestures are a part of a list of other gestures, all of which are private classes. So I had to do string comparisons to get these two gestures, and stop their gestures from canceling touches in the mtk view.

Even when a checkbox is checked with a tap, selection can change at the same time. That is because both operations are happening simultaneously, rather than depending on one another.


Parth (approved):

Parth:
If this came via some stackoverflow, please leave a link in the code

smailbarkouch:
No stackoverflow unfortunately. I figured it out myself.

PR #2343, opened 2024-01-26
lbfs

### LINKED ISSUE #1268: investigating lb-fs

Parth:
# status

+ [ ] better errors
+ [x] massively clean it up
+ [x] incorporate sync
+ [x] move to libs folder

# tests

+ [ ] windows: complex files, subdirs, changes show up post sync, apps refresh themselves
+ [x] linux: complex files, subdirs, changes show up post sync, apps refresh themselves
+ [x] macos: complex files, subdirs, changes show up post sync, apps refresh themselves

fixes #1268 

Thanks to @ssapra2 for all the experiments you ran in #1805, this probably couldn't have come to fruition without your efforts.


clients/cli/comp.fish
```diff
@@ -0,0 +1 @@
+complete -c lockbook -f --condition "not __fish_seen_subcommand_from file-command non-file-command" -a '(lockbook complete fish 0 (commandline -cp))'
```
Parth, line 1: don't merge whatever this is

libs/lb-fs/.gitignore
```diff
@@ -0,0 +1,2 @@
+test_dir/
```
Parth, line 1: don't merge

PR #2426, opened 2024-02-10
apple: realtime syncing status

### LINKED ISSUE #2392: separate sync message updates from frame output
Similar to what is done for import account sync (passing a closure).

### LINKED ISSUE #2402: workspace won't process data while backgrounded

smailbarkouch:
fixes #2392 
fixes #2402

### DIFF
--- a/libs/content/workspace-ffi/SwiftWorkspace/Sources/Editor/iOSMTK.swift
+++ b/libs/content/workspace-ffi/SwiftWorkspace/Sources/Editor/iOSMTK.swift
@@ -747,10 +747,8 @@ public class iOSMTK: MTKView, MTKViewDelegate {
     }
     
     func requestSync() {
-        withUnsafeMutablePointer(to: &workspaceState) { workspaceStatePtr in
-            request_sync(wsHandle, workspaceStatePtr, updateSyncMessage)
-            setNeedsDisplay(self.frame)
-        }
+        request_sync(wsHandle)
+        setNeedsDisplay(self.frame)
     }
     
     func docRenamed(renameCompleted: WSRenameCompleted) {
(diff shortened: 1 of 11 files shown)


Parth (changes requested):
The reason `perform_sync` doesn't take a closure is that it crafts pretty high level sync messages as part of the output. I understand how that could cause problems in the context of https://github.com/lockbook/lockbook/issues/2402. Can't we do frames regardless of whether the workspace is covered up or not?

libs/content/workspace-ffi/SwiftWorkspace/Sources/Editor/iOSMTK.swift
```diff
@@ -853,15 +851,17 @@ public class iOSMTK: MTKView, MTKViewDelegate {
...
-            let redrawIn = Int(truncatingIfNeeded: output.redraw_in)
+            var redrawIn = Int(truncatingIfNeeded: output.redraw_in)
             
-            if redrawIn != -1 {
-                let newRedrawTask = DispatchWorkItem {
-                    self.setNeedsDisplay(self.frame)
-                }
-                DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(redrawIn), execute: newRedrawTask)
-                redrawTask = newRedrawTask
+            if redrawIn == -1 {
```
Parth, line 856: Could this be less weird easily

Parth (changes requested):

libs/content/workspace-ffi/SwiftWorkspace/Sources/Editor/iOSMTK.swift
smailbarkouch: Made it less weird by using a `UInt64`, so I don't need to truncate.

Parth (approved):

PR #2436, opened 2024-02-12
canvas: pen tool bug where new path connects to old path  

### LINKED ISSUE #2342: when ending and beginning a touch in a drawing, the old stroke will connect to the new one

ad-tra:
when implementing some functionalities of Canvas, I've always assumed that if the window is in focus, there's a cursor with a hover position. this assumption is wrong for touch screens because when the user is not pressing or touching there's no hover cursor.  

This PR re-evaluates the assumption and fixes bugs for both pen and eraser tools.  
- pen (fixes #2342)
- eraser: there was an issue where if you erase paths, they would just turn gray but they would stay on the screen. It's now fixed.


smailbarkouch:
I still am experiencing the issue on iOS.

smailbarkouch (changes requested):
I am still experiencing the issue on iOS.

libs/content/workspace/src/tab/svg_editor/selection.rs
```diff
@@ -90,6 +90,7 @@ impl Selection {
                     } else {
                         self.selected_elements = vec![new_selected_el]
                     }
+                    self.current_op = SelectionOperation::Drag;
```
ad-tra, line 93: this is unrelated to this PR, but i noticed a small bug for the selection module when doing QA.
the bug is that when you press on the edge of an un-selected path you go directly into resize instead of grab mode.

smailbarkouch (approved):

PR #2442, opened 2024-02-13
docs: update apple build guide

CoreyCole:
* added step to run the workspace-ffi create_libs.sh
* added what to open with xcode
* added how to sign the app

### DIFF
--- a/docs/guides/build/apple.md
+++ b/docs/guides/build/apple.md
@@ -15,4 +15,7 @@ rustup target add aarch64-apple-ios x86_64-apple-ios aarch64-apple-darwin x86_64
 
 Steps:
 1. In `/libs/lb/lb_external_interface` run `make swift_libs` which will generate lb-rs libs and place them into the correct location within your Xcode project.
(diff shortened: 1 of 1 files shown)


Parth (approved):
🎉

CoreyCole:
In order to sign the app with my personal apple dev email, I needed to change the Bundle Identifier. This shouldn't cause any weirdness, right? Nothing is expecting an exact id of `app.lockbook`

CoreyCole:
image

changing the bundle id might have borked the subscription from the app store

Parth:
Potentially we could discuss this in an issue that is a little more centered around improving the external apple dev experience.

I would be curious to see what the error message was that prompted you to change the bundle identifier, just curious about it. I wonder if there was another solution to that problem, I believe we could author a run configuration which doesn't require signing to run. I'm not sure what the implications would be for running on physical devices, but I think it would be permissive and probably the best first approach.

Alternatively: I can't think of any other components apart from the billing related stuff that would care about this. Potentially we could author a run configuration in xcode that disables these things either under the condition.

PR #2444, opened 2024-02-14
apple: private key hint string for import

CoreyCole:
image


Parth (approved):

Parth:
I think this is a no brainer net positive improvement. Just so you're aware, however: there is a larger conversation around our use of words in code, marketing and the clients. I figured the first step was to create `docs/glossary.md` as an RFC style PR to drive consensus and identify target state. And then gradually migrate & refactor.

CoreyCole:
That makes sense. A glossary would also be first steps to any localization efforts.

PR #2507, opened 2024-02-19
apple/logout

CoreyCole:
- [x] logout confirmation window
- [x] logout resets app state and loops back to onboarding
- [x] delete local files so user is actually logged out


CoreyCole:
it's not pretty..... 😆 
image

CoreyCole:
Needs something like `deleteAccount()` that will delete the local files without deleting the account?
image

Parth:
imo deleting the lockbook folder itself is probably the way I'd go about it. Each core will have a config variable which stores where that core is initialized. You probably don't want to be in the business about reasoning about what to delete and probably want to take a large easy to maintain action. And then guard it as you have in the UI you've shown. Also I think that's a pretty reasonable way of going about warning the user.

CoreyCole:
I suggested just deleting the folder from swift, but travis took the opportunity to teach me how all FFI is hooked up. We went deep an implemented a `clear_local_db()` function for db-rs

But now I'm getting an error when I call it:
image

CoreyCole:
It works as far as my small brain can tell

CoreyCole:
image
image

CoreyCole:
TBD if I can add checkbox icons, but will likely save that for the future
image

CoreyCole:
image

smailbarkouch (dismissed):
Hey Corey, this code looks great, just some comments.

The log out button (in settings) on iOS could benefit from some theming to look more like a button. In addition, we don't have a logout button in settings for macOS. We should add it there (not directly under delete account, there should be some content/space so people don't click delete by accident).

Next, `exit(0)` [is not the way to go about ending the session](https://developer.apple.com/library/archive/qa/qa1561/_index.html#//apple_ref/doc/uid/DTS40007952). We should go back to the onboarding screen. Although this does take more effort since we need to appropriately reset each DI.

clients/apple/lockbook.xcodeproj/project.pbxproj
```diff
@@ -810,7 +820,7 @@
 				COMBINE_HIDPI_IMAGES = YES;
 				CURRENT_PROJECT_VERSION = 2;
 				DEAD_CODE_STRIPPING = YES;
-				DEVELOPMENT_TEAM = 39ZS78S25U;
+				DEVELOPMENT_TEAM = 4PN4ZH29UT;
```
smailbarkouch, line 823: We probably want to reset this back to the original value (we need to give you our team account access).

clients/apple/lockbook.xcodeproj/project.pbxproj
```diff
@@ -847,7 +857,7 @@
...
```
smailbarkouch, line 860: Same as the comment before.

clients/apple/lockbook.xcodeproj/project.pbxproj
```diff
@@ -979,7 +989,7 @@
...
```
smailbarkouch, line 992: Same as the comment before before.

CoreyCole:
I talked with @Parth about going back to onboarding vs exiting. I think exiting is OK for now given logout isn't really a typical use case for general usage. I will add a new issue for log out without exit.

CoreyCole:
@smailbarkouch 
image

Parth:
> I talked with @Parth about going back to onboarding vs exiting. I think exiting is OK for now given logout isn't typical for general usage. I will add a new issue for log out without exit.
> 
> 
> 
> > we don't have a logout button in settings for macOS
> 
> 
> 
> I put the logout for macos in the lockbook toolbar. Do we want it in both places or just the settings screen for macos?

I think you mean menu bar? If so I think that's a nice touch.

CoreyCole:
> I think you mean menu bar? If so I think that's a nice touch.

Yes, the lockbook section of the menu bar (see screenshots above). Right now it is in two places on macos if everyone is okay with that. 👍

smailbarkouch:
I'm personally fine with exiting, I think it helps clean up state a lot better (and results in less code). I'm just worried we could get rejected from the app store. To my knowledge, it has happened before with other apps. But after checking those apps again, it happened many years ago, so maybe that isn't the case anymore. So I'm cool with it.

The second screenshot in the settings looks great. Just one concern, someone will accidentally delete their account. I'm not sure how to separate it anymore than that. But if you and Parth are good with that design, I'm good with it too.

CoreyCole:
There is a confirmation dialog for the delete account button. I could make that confirmation dialog more extreme like my logout confirmation screen if we want

Parth (approved):
imo this is good to go 🚀

smailbarkouch:
> There is a confirmation dialog for the delete account button. I could make that confirmation dialog more extreme like my logout confirmation screen if we want

On second thought, its good. Feel free to merge 🚀

PR #2512, opened 2024-02-20
apple: route paste and drop operations through workspace

smailbarkouch:


libs/content/workspace-ffi/SwiftWorkspace/Sources/Editor/iOSMTK.swift
```diff
@@ -163,39 +163,34 @@ public class iOSMTKTextInputWrapper: UIView, UITextInput, UIDropInteractionDeleg
             self.mtkView.setNeedsDisplay()
         }
     }
+    
+    func sendImage(img: Data, isPaste: Bool) {
+        let imgPtr = img.withUnsafeBytes { (pointer: UnsafeRawBufferPointer) -> UnsafePointer<UInt8> in
+            return pointer.baseAddress!.assumingMemoryBound(to: UInt8.self)
```
tvanderstad, line 169: how is this returning something if this function has no return type 😵‍💫 swift is so confusing to me

libs/content/workspace-ffi/SwiftWorkspace/Sources/Editor/iOSMTK.swift
```diff
@@ -163,39 +163,34 @@ public class iOSMTKTextInputWrapper: UIView, UITextInput, UIDropInteractionDeleg
...
-                } catch {
-                    return false
-                }
-
-                if let lbImageURL = workspaceState!.importFile(imageUrl) {
-                    pasteText(text: lbImageURL)
-
-                    return true
-                }
+            if let img = image.pngData() ?? image.jpegData(compressionQuality: 1.0) {
```
tvanderstad, line 188: did you test with each kind of image? would be somewhat surprised if jpegs worked

libs/content/workspace-ffi/src/apple/common_ffi.rs
```diff
@@ -134,15 +135,41 @@ pub unsafe extern "C" fn clipboard_paste(obj: *mut c_void, content: *const c_cha
...
 ) {
     let obj = &mut *(obj as *mut WgpuWorkspace);
-    let content = std::slice::from_raw_parts(content, length).to_vec();
-    obj.context.push_event(workspace_rs::Event::Paste {
-        content: vec![ClipContent::Png(content)],
-        position: egui::Pos2::ZERO, // todo: cursor position
-    });
+    let img = std::slice::from_raw_parts(content, length).to_vec();
+    let content = vec![ClipContent::Png(img)];
+    let position = egui::Pos2::ZERO; // todo: cursor position
```
tvanderstad, line 144: At some point we'll support passing this in so we can put the image where you dropped it instead of at the cursor or whatever. In the meantime, I thought macos had behavior where if you drop an image into the file tree it imports it as a file and if you drop it into the editor it additionally inserts a link - is that right, and if so was that behavior preserved?

CoreyCole:
fixed image paste crash for me

libs/content/workspace-ffi/SwiftWorkspace/Sources/Editor/iOSMTK.swift
CoreyCole: that return value is going into the imgPtr I think? `->` is an inline function

smailbarkouch: Yeah its a little bit cryptic. The inner closure gives the outer closure its type.

libs/content/workspace-ffi/SwiftWorkspace/Sources/Editor/iOSMTK.swift
smailbarkouch: This was some old code from before, and I believe I tested it with tiff, jpeg, and png (back then at least). It works right now too, I just tried jpegs.

tvanderstad (approved):

libs/content/workspace-ffi/src/apple/common_ffi.rs
smailbarkouch: That behavior is still preserved. The difference is that if you drop an image into the file tree, it goes through entirely separate logic, and it uses the import functions from core.

Issue #2482, opened 2024-02-18
low fps 


ad-tra:
perf is not smooth whether that's
- scrolling a large md file
- drawing anything
- searching

PR #2517, opened 2024-02-21
windows: only reconfigure surface when resized

tvanderstad:
fixes https://github.com/lockbook/lockbook/issues/2482

me:
> I took some measurements and wgpu::SurfaceTexture.present() is taking >30ms per frame. Is that normal? Can you suggest troubleshooting next steps?

chatgpt:
> Experiment with disabling VSync or using adaptive sync techniques if your application's performance profile allows for it.

me:
> I disabled vsync by passing wgpu::PresentMode::Immediate. Now my timing statements indicate that configuring the surface is taking ~30ms

chatgpt:
> To optimize and avoid unnecessary reconfigurations (and thus minimize delays), you should track the current configuration of the surface and only reconfigure when there's an actual change in the parameters (e.g., width, height, or present_mode). As there's no direct wgpu API call to retrieve the current configuration of a surface, you'll need to manually track these parameters within your application's state

### DIFF
--- a/clients/egui/src/lib.rs
+++ b/clients/egui/src/lib.rs
@@ -147,6 +147,10 @@ pub struct WgpuLockbook {
     pub surface: wgpu::Surface,
     pub adapter: wgpu::Adapter,
 
+    // remember size last frame to detect resize
+    pub surface_width: u32,
+    pub surface_height: u32,
+
     pub rpass: egui_wgpu_backend::RenderPass,
     pub screen: egui_wgpu_backend::ScreenDescriptor,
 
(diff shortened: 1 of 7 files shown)


clients/egui/src/lib.rs
```diff
@@ -248,18 +252,23 @@ impl WgpuLockbook {
...
-            width: self.screen.physical_width,
-            height: self.screen.physical_height,
-            present_mode: wgpu::PresentMode::Fifo,
-            alpha_mode: CompositeAlphaMode::Auto,
-            view_formats: vec![],
-        };
-        if surface_config.width * surface_config.height != 0 {
+    pub fn configure_surface(&mut self) {
+        let resized = self.screen.physical_width != self.surface_width
+            || self.screen.physical_height != self.surface_height;
```
tvanderstad, line 257: the magic 🪄

Parth (approved):

PR #2518, opened 2024-02-21
windows: file drop fixes

tvanderstad:
fixes https://github.com/lockbook/lockbook/issues/2495
fixes https://github.com/lockbook/lockbook/issues/2358

image

### DIFF
--- a/clients/windows/src/input/file_drop.rs
+++ b/clients/windows/src/input/file_drop.rs
@@ -63,6 +63,7 @@ pub fn handle(app: &mut WgpuLockbook, object: Option<IDataObject>) -> bool {
         };
         let mut rgelt = [FORMATETC::default(); 1];
         loop {
+            println!("loop start");
             let mut fetched: u32 = 0;
             if unsafe { format_enumerator.Next(&mut rgelt, Some(&mut fetched as _)) }.is_err() {
                 break;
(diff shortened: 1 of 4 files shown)


clients/windows/src/input/file_drop.rs
```diff
@@ -71,12 +70,14 @@ pub fn handle(app: &mut WgpuLockbook, object: Option<IDataObject>) -> bool {
...
-            let mut format_name = [0u16; 1000];
             let is_predefined_format = format_str(format).is_some();
-            let is_registered_format =
-                unsafe { GetClipboardFormatNameW(format.0 as _, &mut format_name) != 0 };
-            if !is_predefined_format && !is_registered_format {
+            // use windows::Win32::System::DataExchange::*,
+            // let mut format_name = [0u16; 1000];
+            // let is_registered_format =
+            // unsafe { GetClipboardFormatNameW(format.0 as _, &mut format_name) != 0 };
+            if !is_predefined_format {
```
tvanderstad, line 80: app was getting data for an unnecessary format (probably the file data itself instead of the path) and it was taking a long time, causing the app to hang when you drop a file because it's importing the file itself (in addition to the path) on the UI thread

Parth (approved):

Issue #2487, opened 2024-02-19
long press not right clicking


Parth:
from #2413 QA session

tvanderstad:
this work broke something so I'm effectively reverting it in https://github.com/lockbook/lockbook/pull/2550

tvanderstad:
fixed by https://github.com/lockbook/lockbook/pull/2880 (see egui release notes - long presses now count as right clicks everywhere)

PR #2520, opened 2024-02-21
windows/egui: long press to right click

tvanderstad:
...and improve file scroll experience.

fixes https://github.com/lockbook/lockbook/issues/2487
also fixes an issue where dragging a file down past the bottom of the tree view would scroll indefinitely
also fixes an issue where dragging wouldn't register if you did it too quickly bc by the frame the drag registered, the cursor would already be over a different file (could be worked around by dragging sideways on the file first, then dragging it to destination)

demo (2nd half is touch-only):

https://github.com/lockbook/lockbook/assets/6198756/bb86c55c-fb83-4981-b9d1-65574c8231c6

### DIFF
--- a/clients/egui/src/account/tree/mod.rs
+++ b/clients/egui/src/account/tree/mod.rs
@@ -96,31 +96,17 @@ impl FileTree {
         end.x += 70.0;
         end.y += 50.0;
 
-        let response = ui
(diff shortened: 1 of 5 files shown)


ad-tra (approved):
nice!! thanks for fixing the infinite scroll issue too

Parth (approved):

PR #2589, opened 2024-03-20
android: workspace

smailbarkouch:
Workspace support for android.

Demonstration:

https://github.com/lockbook/lockbook/assets/20663038/8d9d6572-45ca-4b01-99ba-b98e3556783a

Known issues:
+ rotating (or any other configuration change) while having any image data in Workspace will crash the app


Parth (approved):

Parth:
Ideas on outstanding image problems:
1. add integration specific code to see if we can preserve more of wgpu's internal state, as the arrises from [here](https://github.com/hasenbanck/egui_wgpu_backend/blob/master/src/lib.rs#L387).
2. convert from bytes to a texture every frame
3. query to see if texture is still available at frame time, and regenerate it if not

PR #2598, opened 2024-04-02
Windows rescue

vvmirkovic:
Adding a feature flag for releaser to use to activate winstaller. Some uncertainties:

1. Does the name I chose for the feature flag make sense to you?
2. Is the windows flag now redundant in winstaller? I think not since the winstaller flag shouldn't ever be used if it's not on windows, but maybe it's worth keeping both?
3. How should I go about testing that the releaser isn't broken? Is that already integrated into ci?

### DIFF
--- a/clients/windows/src/input/mod.rs
+++ b/clients/windows/src/input/mod.rs
@@ -5,5 +5,4 @@ pub mod message;
 pub mod mouse;
 pub mod pointer;
 
-pub use message::Message;
 pub use message::Point;
--- a/libs/lb-fs/src/mount.rs
+++ b/libs/lb-fs/src/mount.rs
@@ -71,5 +71,6 @@ pub async fn umount() {
 
 #[cfg(target_os = "windows")]
 pub async fn umount() {
+    info!("todo");
     todo!()
 }
--- a/utils/releaser/src/windows/desktop.rs
+++ b/utils/releaser/src/windows/desktop.rs
@@ -29,7 +29,7 @@ fn build_x86() -> CliResult<()> {
 
     Command::new("cargo")
         .env("LB_TARGET", "x86_64-pc-windows-msvc")
-        .args(["build", "-p", "winstaller", "--release", "--target=x86_64-pc-windows-msvc"])
+        .args(["build", "-p", "winstaller", "--release", "--target=x86_64-pc-windows-msvc", "--feature", "releaser"])
         .assert_success();
 
     Ok(())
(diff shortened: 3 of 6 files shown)


clients/windows/src/input/mod.rs
```diff
@@ -5,5 +5,4 @@ pub mod message;
 pub mod mouse;
 pub mod pointer;
 
-pub use message::Message;
```
vvmirkovic, line 8: I checked out the parent module and other submodules and didn't see anything taking advantage of this. It gets imported as message::{Message,...

clients/windows/src/input/mod.rs
```diff
@@ -5,5 +5,4 @@ pub mod message;
...
 pub use message::Point;
```
vvmirkovic, line 8: Is this even worth keeping? I think the only place this is taken advantage of is here
https://github.com/lockbook/lockbook/blob/e92ed51631ff75ce67ea2527437c6f395dde0831/clients/windows/src/input/mouse.rs#L4

libs/lb-fs/src/mount.rs
```diff
@@ -71,5 +71,6 @@ pub async fn umount() {
 
 #[cfg(target_os = "windows")]
 pub async fn umount() {
+    info!("todo");
```
vvmirkovic, line 74: info is an unused important in the file when using windows. Stuck this in here since the other unmount functions used info in a similar place. 

Perhaps finishing this todo is a next action for me?

utils/releaser/src/windows/desktop.rs
```diff
@@ -29,7 +29,7 @@ fn build_x86() -> CliResult<()> {
 
     Command::new("cargo")
         .env("LB_TARGET", "x86_64-pc-windows-msvc")
-        .args(["build", "-p", "winstaller", "--release", "--target=x86_64-pc-windows-msvc"])
+        .args(["build", "-p", "winstaller", "--release", "--target=x86_64-pc-windows-msvc", "--feature", "releaser"])
```
vvmirkovic: I hope this works. Not sure how to test it

utils/winstaller/src/main.rs
```diff
@@ -1,6 +1,6 @@
 #![cfg_attr(not(debug_assertions), windows_subsystem = "windows")]
 
-#[cfg(windows)]
+#[cfg(all(windows, feature = "releaser"))]
```
vvmirkovic, line 3: windows flag might be superfluous now

PR #2599, opened 2024-04-02
Windows non-release compilation fix

vvmirkovic:
Adding a feature flag for releaser to use to activate winstaller. Some uncertainties:

1. Does the name I chose for the feature flag make sense to you?
2. Is the windows flag now redundant in winstaller? I think not since the winstaller flag shouldn't ever be used if it's not on windows, but maybe it's worth keeping both?
3. How should I go about testing that the releaser isn't broken? Is that already integrated into ci?


clients/windows/src/input/mod.rs
```diff
@@ -5,5 +5,4 @@ pub mod message;
 pub mod mouse;
 pub mod pointer;
 
-pub use message::Message;
```
vvmirkovic, line 8: I checked out the parent module and other submodules and didn't see anything taking advantage of this. It gets imported as message::{Message,...

clients/windows/src/input/mod.rs
```diff
@@ -5,5 +5,4 @@ pub mod message;
...
 pub use message::Point;
```
vvmirkovic, line 8: Is this even worth keeping? I think the only place this is taken advantage of is here
https://github.com/lockbook/lockbook/blob/e92ed51631ff75ce67ea2527437c6f395dde0831/clients/windows/src/input/mouse.rs#L4

libs/lb-fs/src/mount.rs
```diff
@@ -71,5 +71,6 @@ pub async fn umount() {
 
 #[cfg(target_os = "windows")]
 pub async fn umount() {
+    info!("todo");
```
vvmirkovic, line 74: info is an unused important in the file when using windows. Stuck this in here since the other unmount functions used info in a similar place.

Perhaps finishing this todo is a next action for me?

utils/releaser/src/windows/desktop.rs
```diff
@@ -29,7 +29,15 @@ fn build_x86() -> CliResult<()> {
...
-        .args(["build", "-p", "winstaller", "--release", "--target=x86_64-pc-windows-msvc"])
+        .args([
+            "build",
+            "-p",
+            "winstaller",
+            "--release",
+            "--target=x86_64-pc-windows-msvc",
+            "--feature",
+            "releaser",
+        ])
```
vvmirkovic, line 40: I hope this works. Not sure how this should be tested

utils/winstaller/src/main.rs
```diff
@@ -1,6 +1,6 @@
 #![cfg_attr(not(debug_assertions), windows_subsystem = "windows")]
 
-#[cfg(windows)]
+#[cfg(all(windows, feature = "releaser"))]
```
vvmirkovic, line 3: windows flag might be superfluous now

clients/windows/src/input/mod.rs
Parth: So I'm realizing the value of this PR is higher than I expected -- seems like we have a large body of windows code that was never being clippy'd.

So I imagine you came about this line via clippy? If clippy told you to remove this that's totally fine. Does the rest of clippy pass now?

libs/lb-fs/src/mount.rs
Parth: fwiw the `todo!()` will `eprintln` a panic message. The underlying windows library we use for this is this lib: https://github.com/xetdata/nfsserve

And it only supports windows pro, do you have windows pro? Have you also messed around with the `fs` command? I think there's going to be a longer term windows-centric way to do `fs` on windows.

If you find this body of work interesting we can talk about areas of inquiry. I would be curious about exploring something like this: https://learn.microsoft.com/en-us/windows/win32/projfs/projected-file-system

could be other approaches as well

utils/releaser/src/windows/desktop.rs
Parth: Yeah the easiest way would be to just run the releaser -- you can `cargo install --path .` in the `releaser` folder and verify that it _fails successfully_ lol. You can try it on `master`, it should compile a few things and then fail to upload to github (because you're not authorized, or because 0.9.0 exists already or other such error). 

Then you can: 
1. switch to your branch 
2. execute `cargo clean` 
3. comment out `--feature releaser` 
4. verify you get a different error message (probably something like the original error you were getting -- `LB_TARGET` not found)
5. bring back `--feature releaser`, run and get the github error again (the feature flag worked)

Parth: I think you telling me the outcome of ^ is all that stands in the way of this being merged

vvmirkovic: I'm not sure of a great way to test it. After setting GITHUB_TOKEN to a dummy value, the command that requires the feature flag does get run before failing at the github portion. 

However, it passes with and without the feature flag. One different outcome I noticed though is that winstaller file is much larger after running the releaser when compiled with the feature flag, compare to when it is compiled without the feature flag. 

Here are screenshots to show that
Result of releaser when compiled with the flag

Result of releaser when compiled without the flag
image

clients/windows/src/input/mod.rs
vvmirkovic: Yeah, clippy was telling me to remove it, and clippy now passes after I made all the instructed changes

clients/windows/src/input/mod.rs
vvmirkovic: Some clarification for my previous comment: When I said "this", I meant Point. Clippy was telling me to removed Message, but Point being used there allows shorthand use Point (Point instead of message::Point), However, it's only used in one place that doesn't actually make the line shorter. It's 
`use super::{message::MessageAppDep, Point}; `
instead of
`use super::message::{MessageAppDep, Point}; `

Hence, as far as I can tell, using Point is pointless

vvmirkovic: I'm going to remove it unless told otherwise

utils/releaser/src/windows/desktop.rs
Parth: a successful `winstaller` build contains the full `egui` client within it a la [`include_bytes!`](https://doc.rust-lang.org/std/macro.include_bytes.html). Like I mentioned [here](https://discord.com/channels/1014184997751619664/1222701324462854187/1227625533743824976) ideally the `277kb` outcome is an impossibility. But I know we're already juggling a lot of constraints at this point, and just getting a windows person to be able to clippy locally is more important.

Parth (approved):
🚀

PR #2639, opened 2024-04-20
canvas: fix pixelated output

ad-tra:
fixes: #2531

### DIFF
--- a/libs/content/workspace/src/tab/svg_editor/mod.rs
+++ b/libs/content/workspace/src/tab/svg_editor/mod.rs
@@ -140,19 +140,25 @@ impl SVGEditor {
             resolve_string: Self::lb_local_resolver(&self.core),
         };
 
-        let options =
-            usvg::Options { image_href_resolver: lb_local_resolver, ..Default::default() };
+        let options = usvg::Options {
+            image_href_resolver: lb_local_resolver,
+            image_rendering: usvg::ImageRendering::OptimizeSpeed,
+            dpi: 300.,
+            ..Default::default()
+        };
 
         let mut utree: usvg::Tree =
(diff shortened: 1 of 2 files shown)


libs/content/workspace/src/tab/svg_editor/mod.rs
```diff
@@ -140,19 +140,26 @@ impl SVGEditor {
...
+        let options = usvg::Options {
+            image_href_resolver: lb_local_resolver,
+            image_rendering: usvg::ImageRendering::OptimizeSpeed,
+            ..Default::default()
+        };
 
         let mut utree: usvg::Tree =
             usvg::TreeParsing::from_str(&self.buffer.to_string(), &options).unwrap();
-        let available_rect = ui.available_rect_before_wrap();
+        let high_res_factor = ui.input(|r| r.pixels_per_point) - 1.0;
```
Parth, line 151: why are you subtracting by 1? if this returns something like 1, you'll be multiplying by 0

Parth (changes requested):

libs/content/workspace/src/tab/svg_editor/mod.rs
ad-tra: multiplying by 0 is fine. that's the intended behavior for screens with 1-1 mapping between physical and logical pixels. 

```rust
        let available_rect = self.inner_rect.expand2(egui::vec2(
            self.inner_rect.width() * high_res_factor,
            self.inner_rect.height() * high_res_factor,
        ));
```

the size of the output `available_rect` is the size of the current screen + whatever expansion is needed to account for higher resolution screens. if the `high_res_factor` is 0 then the size of the image will be equal to the size of the current screens.

Parth:
superceded by #2654

PR #2678, opened 2024-06-11
apple: fix memory leaks

### LINKED ISSUE #2660: ws memory leak

smailbarkouch:
## How memory leaks happen
Memory leaks on our Apple platforms happen in one of two places...
+ within our Swift code 
+ at the boundary between Rust and Swift (FFI)

Swift manages memory using ARC (Automatic Reference Counting), and it fails to free memory when there are cyclical memory references. For instance, if I create a class that contains a closure referencing the class itself, neither will have their memory freed, since both "depend" on one another. There are other ways to have memory leaks, but this is the main one I fixed within this PR.

On FFI side, we allocate memory from within Rust, make Rust's borrow checker forget it, and trust that Swift will free it. When we don't free that memory (and forget about its reference), it becomes a memory leak.

## Memory leaks fixed

We had cyclical memory references...
+ when registering background tasks
+ within the closures in Workspace
+ within the closures in the WorkspaceWrappers

We didn't retain the event registering handle used for detecting modifier events (on macOS).

We didn't free the selection rects (used for UITextInteraction) provided by Rust.

We rendered a frame within our `init_ws` function, in which the IntegrationOutput (containing raw C Strings), was never given back to Swift to free.

We kept calling `get_core_ptr()` every Swift re-render of Workspace, and we didn't free those pointers.

In certain path and content search edge cases, the messages sent from Rust to Swift were not freed.

## Additional thoughts

Now, seemingly, there are no memory leaks when using Lockbook, except when you use search.

fixes #2660


clients/apple/Shared/Views/AppView.swift
```diff
@@ -17,6 +17,32 @@ struct AppView: View {
...
+                                DI.errors.errorWithTitle("Malformed link", "Cannot open file")
+                                return
+                            }
+        
+                            DispatchQueue.global(qos: .userInitiated).async {
+                                while !DI.files.hasRootLoaded {
+                                    if DI.accounts.calculated && DI.accounts.account == nil {
+                                        return
+                                    }
+                                }
```
Parth, line 31: my gut says this is a very aggressive way to express this idea, and is not friendly to the scheduler. I don't know the details of how a lot of this plays out in swift / swiftui, but generally adding a sleep(1) at the moment of loop iteration is enough of a hacky signal to the scheduler to try to do something else meanwhile.

Where this performs poorly is at startup the loop iterates like 10k times in 100ms making no progress while the scheduler doesn't understand that it's making no progress. This will manifest itself as startup lag.

libs/content/workspace-ffi/src/apple/common_ffi.rs
```diff
@@ -68,8 +68,6 @@ pub unsafe extern "C" fn init_ws(
         surface_height: 0,
     };
 
-    obj.frame();
```
Parth, line 71: I believe this used to exist as a mechanism for initial theme loading so that you weren't flash banged in dark mode by 1 frame of light mode. Could you do a quick sanity check to see if this is truly not needed?

libs/content/workspace-ffi/src/apple/ios_ffi.rs
```diff
@@ -707,7 +706,20 @@ pub unsafe extern "C" fn selection_rects(
         cont_start = new_end.selection.end();
     }
 
-    UITextSelectionRects { size: selection_rects.len() as i32, rects: selection_rects.as_ptr() }
+    UITextSelectionRects {
+        size: selection_rects.len() as i32,
+        rects: Box::into_raw(selection_rects.into_boxed_slice()) as *const CRect,
```
Parth, line 711: Why does this need to be heap allocated at all?

libs/content/workspace/src/lib.rs
```diff
@@ -6,6 +6,7 @@ pub mod theme;
 pub mod widgets;
 pub mod workspace;
 
+pub use lb_rs;
```
Parth, line 9: I'm fairly certain that this means the compiler is free to optimize lb_rs for workspace (and ultimately workspace_ffi) differently. It may not, but it may. Could you decouple this change from your PR for now?

You can feel free to continue doing this locally for now, and at release time it won't matter as both libraries will be built.

Your finding that the functions from core will be included twice is a good discovery, it'll simplify the design of workspace_ffi providing core once. This means the only additional thing build_swift_libs has to do is import the header file. That'll mean you won't ever have to build both again and we won't have to include all the static objects for core twice. It'll be a good cleanup in a few regards when we do it properly.

Parth (changes requested):

libs/content/workspace-ffi/src/apple/common_ffi.rs
Parth: (or the opposite)

libs/content/workspace-ffi/src/apple/ios_ffi.rs
Parth: aka why is it a `const *` instead of just included directly?

clients/apple/Shared/Views/AppView.swift
smailbarkouch: This isn't new to this PR, it's somewhat always how I have handled opening links within Lockbook. I agree with your sentiment, I can fix it in a future PR.

libs/content/workspace-ffi/src/apple/ios_ffi.rs
smailbarkouch: `Vec` doesn't have a C representation, so if I tried to return a Vec, Swift wouldn't be able to handle it.

Also, when I heap allocate it (done within the `into_boxed_slice`), it will remove all the excess space the `Vec` may have. It means the `.length` corresponds to the size of the array.

clients/apple/Shared/Stateful Logic/AccountService.swift
```diff
@@ -25,8 +30,62 @@ class AccountService: ObservableObject {
         }
         
         calculated = true
+        #if os(macOS)
+        registerBackgroundTasks()
+        #endif
     }
     
+    #if os(macOS)
```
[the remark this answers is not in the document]
smailbarkouch: @Parth Workspace has background sync?

libs/content/workspace-ffi/src/apple/common_ffi.rs
smailbarkouch: I checked, no flash bang.

libs/content/workspace/src/lib.rs
smailbarkouch: Sure, i'll change it back.

clients/apple/Shared/Stateful Logic/AccountService.swift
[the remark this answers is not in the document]
smailbarkouch: Okay, I removed it.

clients/apple/Shared/Views/AppView.swift
Parth: just add a sleep(1) for my soul

Parth (approved):

PR #2707, opened 2024-06-23
apple: fix the energy consumption problem

Parth:
we stopped return repaint_after and the integration fell back to continuous redraws

https://discord.com/channels/1014184997751619664/1026208854280777789/1253746606855950428


libs/content/workspace-ffi/SwiftWorkspace/Sources/Editor/iOSMTK.swift
```diff
@@ -720,7 +720,7 @@ public class iOSMTK: MTKView, MTKViewDelegate {
         
         dark_mode(wsHandle, isDarkMode())
         set_scale(wsHandle, Float(self.contentScaleFactor))
-        let output = draw_editor(wsHandle)
+        let output = draw_workspace(wsHandle)
```
smailbarkouch, line 723: Make sure to switch `draw_editor` to `draw_workspace` in `macMTK.swift` too.

PR #2708, opened 2024-06-24
canvas: bring back dark/light mode dynamic colors and support opacity

ad-tra:
https://github.com/lockbook/lockbook/assets/66345861/b4f05ef1-4e24-45f6-9b43-5d2ed4e1e05b


libs/content/workspace/src/theme/palette.rs
```diff
@@ -36,27 +36,29 @@ impl ThemePalette {
     };
 
     // todo: passing the is_dark_mode aram doesn't feel like good data modeling
```
tvanderstad, line 38: nit: can cleanup this comment now

PR #2719, opened 2024-06-26
editor: change range data model from `(x, y)` to `x..y`

tvanderstad:


tvanderstad:
moved to draft because backspace isn't working 😓

tvanderstad:
closing because value is unclear and I introduce some critical bugs I don't feel like fixing (backspace isn't working and neither is dragging backwards to create a selection)
* first complaint is that `std::ops::Range` doesn't implement `Copy` so gotta clone all over
* second complaint is that there isn't a mechanism to make sure `start <= end` and the subtle difference between `.start` and `.start()` (implemented via extension trait) wouldn't be clear

PR #2721, opened 2024-06-27
ios: floating cursor support

### LINKED ISSUE #1864: floating cursor problems
doesn't move the way it should, for instance auto scroll doesn't work. Very hard to navigate down

smailbarkouch:
Demonstration:

https://github.com/lockbook/lockbook/assets/20663038/f7eb9044-48c6-4fc3-99ff-45fcecb9bdf8

Notes
+ Suggestions are not working properly, and it will be fixed in a later PR altogether (since it has more issues on its own).

fixes #1864

### DIFF
--- a/libs/content/workspace-ffi/SwiftWorkspace/Sources/Editor/iOSMTK.swift
+++ b/libs/content/workspace-ffi/SwiftWorkspace/Sources/Editor/iOSMTK.swift
@@ -126,25 +126,6 @@ public class iOSMTKTextInputWrapper: UIView, UITextInput, UIDropInteractionDeleg
         setFloatingCursorLoc(point: point, animate: true)
     }
     
-    func setFloatingCursorLoc(point: CGPoint, animate: Bool) {
-        let pos = closestPosition(to: point)
-        let cursorRect = caretRect(for: pos!)
-        
-        lastFloatingCursorRect = cursorRect
-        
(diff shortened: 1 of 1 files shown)


Parth:
Some thoughts as I use it, feels better in some ways, but still has the same problems in other ways let me try to narrow in on what I'm experiencing. People (including me) rely on this interaction heavily and I think we do need to get the details right for note taking on iOS to be viable experience.

Fill up a document with varied text (headings, bullets, and longer paragraphs) in both notes and lockbook. Switch between the two of them to rapidly contrast your experience. 

# unreachable areas
1. Place your cursor at the first location in the document
2. Start FC (floating cursor)
3. In both apps at this point you're at the bottom of the screen and your cursor is at the top. In the notes implementation, you're able to exploit the clamp, by moving further in the direction of the first character to in some sense reset the thumb position. In our app you aren't able to do this. The animation seems to show a clamp as the title of [a5e7616](https://github.com/lockbook/lockbook/pull/2721/commits/a5e761600913c53110daa2e794b084468db8a422) suggests, but the underlying cursor does not behave as such.

My best attempt to show this

https://github.com/lockbook/lockbook/assets/821972/f09b1236-6147-462e-8a34-c34b77038ac5

# sticky

smailbarkouch:
Get rid of stickiness.

PR #2756, opened 2024-07-15
editor: support windows-style line endings

tvanderstad:
...also present in XCode log output. Pasting any multi-line content from XCode logs crashed the editor before this change.

### DIFF
--- a/libs/content/workspace/src/tab/markdown_editor/editor.rs
+++ b/libs/content/workspace/src/tab/markdown_editor/editor.rs
@@ -271,8 +271,7 @@ impl Editor {
                 &self.bounds.ast,
                 &self.appearance,
             );
-            self.bounds.paragraphs =
-                bounds::calc_paragraphs(&self.buffer.current, &self.bounds.ast);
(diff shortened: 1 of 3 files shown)


libs/content/workspace/src/tab/markdown_editor/editor.rs
```diff
@@ -271,8 +271,7 @@ impl Editor {
                 &self.bounds.ast,
                 &self.appearance,
             );
-            self.bounds.paragraphs =
-                bounds::calc_paragraphs(&self.buffer.current, &self.bounds.ast);
+            self.bounds.paragraphs = bounds::calc_paragraphs(&self.buffer.current);
```
tvanderstad, line 274: note: simplified paragraph calculation to always split on newlines. It checked with ast to ignore newlines that were part of captured syntax regions, but that was for an iteration of conditional character capture that would capture newlines e.g. to hide the whole syntax line for fenced code blocks

smailbarkouch (approved):

Issue #2760, opened 2024-07-19
Better Line rendering


ad-tra:
- thick start
- oscillating lines 
- drawing a dot is a bad time

Issue #2761, opened 2024-07-19
rich apple pencil support


ad-tra:
- [x] #2926
- [ ] #2849 
- [x] support apple pencil hover
- [ ] support squeeze 
- [x] support double tap
- [x] #2890

Parth:
All that remains is #3507 so we'll track this there

PR #2773, opened 2024-07-30
canvas: lyon based tessellation 

ad-tra:
fixes: #2760 #2761 #2689

# After
image

# Before
image


libs/content/workspace/src/tab/svg_editor/pen.rs
```diff
@@ -87,14 +102,35 @@ impl Pen {
...
                     self.path_builder.cubic_to(pos);
                     p.data = self.path_builder.path.clone();
+
+                    if let Some(f) = force {
+                        if let Some(pressure) = &mut p.pressure {
+                            pressure.push(f)
+                        } else {
+                            p.pressure = Some(vec![f]);
+                        }
+                    // sometimes the force is missing from the event so just autofill it based on the last force
```
tvanderstad, line 130: good to know, seems reasonable

libs/content/workspace/src/tab/svg_editor/renderer.rs
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    pub fn render_svg(&mut self, ui: &mut egui::Ui, buffer: &mut Buffer, painter: egui::Painter) {
+        let mut elements = buffer.elements.clone();
+
+        self.painter = Some(painter.clone());
+        let dark_mode_changed = ui.visuals().dark_mode != self.dark_mode;
+        self.dark_mode = ui.visuals().dark_mode;
+
+        // todo: should avoid runing this on every frame, because the images are allocated once
```
tvanderstad, line 60: I assume this is okay to do later since the PR's ready for review

libs/content/workspace/src/tab/svg_editor/renderer.rs
```diff
@@ -0,0 +1,268 @@
...
+                let mut first = None;
+                let mut i = 0;
+
+                while let Some(seg) = p.data.get_segment(i) {
+                    let pressure = if let Some(ref pressure) = p.pressure {
+                        let pressure_at_segment = pressure.get(i);
+                        if pressure_at_segment.is_some() {
+                            pressure_at_segment
+                        } else {
+                            pressure.get(i - 1)
```
tvanderstad, line 181: I thought these were filled in above but seems harmless

libs/content/workspace/src/tab/svg_editor/renderer.rs
```diff
@@ -0,0 +1,268 @@
...
+                        }
+                    } else {
+                        None
+                    };
+
+                    let normalized_pressure =
+                        if let Some(prsr) = pressure { *prsr + 0.5 } else { 1.0 };
```
tvanderstad, line 188: How'd you come up with `+ 0.5`? When I worked on the C# Windows client and tried to make drawings show up the same there as it did on Apple I just measured the pixel width of strokes of various pressure ended up with a `+ 0.16`, a `* 1.55`, and a clamp between 0.16 and 1.0

tvanderstad: https://github.com/lockbook/lockbook/commit/2991203d864fbb76972b96864817efbd0768cc06#diff-c2e5e440a219125a999af52a9e24f374447067b9411e86b3f01ce74060cf9e25 (clients/windows/core/Draw.cs if the link doesn't work)

libs/content/workspace/src/tab/svg_editor/selection/mod.rs
```diff
@@ -174,63 +196,64 @@ impl Selection {
...
-                                pos.y - selection.prev_pos.y,
-                            );
-                            selection.transform = selection.transform.post_concat(transform);
-                            match el {
-                                parser::Element::Path(p) => {
-                                    p.data.apply_transform(u_transform_to_bezier(&transform));
-                                }
-                                parser::Element::Image(img) => img.apply_transform(transform),
-                                parser::Element::Text(_) => todo!(),
+            if pos.is_some() {
```
tvanderstad, line 226: nit:
```suggestion
            if let Some(p) = pos {
```

Parth: non-nit imo this would eliminate an unwrap below

libs/content/workspace/src/tab/svg_editor/renderer.rs
Parth: worth calling out that ideally pressure will be a device specific knob which will result in a thickness stored in some svg complaint manner. It's a slightly different process as we'll likely be trying to match the pen *behavior* of the device, rather than lockbook's appearance from device to device

libs/content/workspace/src/tab/svg_editor/renderer.rs
ad-tra: yeah, not a heavy perf hit. I'm just scanning through the images and allocating textures to newly added ones. most of the frames, the images will have already been allocated and no duplicate texture will be reallocated. so this wastes an O(images) runtime.

libs/content/workspace/src/tab/svg_editor/renderer.rs
ad-tra: @tvanderstad  egui expects the force value to be normalized [between 0 and 1](https://docs.rs/egui/latest/egui/enum.Event.html#variant.Touch).
in this case i'm shifting the range form [0,1] to [0.5,1.5] because when you press very lightly it'll be jarring to have the thickness multiplied by a value very close to 0 and cause it to almost disappear from the screen.

Parth (approved):

PR #2781, opened 2024-08-12
ios: support for sharing into the app

### LINKED ISSUE #1951: share documents in and out of lockbook
Users should be able to share their files using the common sharing method on iOS, iPadOS, and macOS. Lockbook should also appear as an option when other apps are sharing their files (although this is not necessarily important for macOS)

- [x] share files out of lockbook
- [ ] share files into lockbook

smailbarkouch:
This PR includes:
+ adding support for sharing files, images, and videos into lockbook
+ redesigning the move and accept share sheet (and unifying the design)
+ redesigning the pending shares screen
+ shared folders have a special icon

(accept share) iOS:

https://github.com/user-attachments/assets/ba4a0ad2-1034-45df-9be9-75b61754aab1

fixes #1951


Parth:
Huge, can you update feature matrix in readme as well

tvanderstad (approved):
Didn't review too carefully but no red flags for me

Issue #2544, opened 2024-02-22
path and content search improvements


smailbarkouch:
Pertaining to macOS.

The search bar should be moved to the sidebar. In addition, search results who's direct parent is root will have empty text for their parent name (from within the search result).

Issue #2538, opened 2024-02-22
sometimes pressed search results don't open


smailbarkouch:
On iOS for content and path search

Issue #2431, opened 2024-02-11
sometimes search will crash


smailbarkouch:
Type any text, the crash will happen randomly (but pretty consistently).

Parth:
Also typing some test, and then closing the app, and coming back usually crashes too

Issue #2378, opened 2024-02-06
scrutinize cargo.toml


Parth:
specifically different packages importing `lb-rs` and `reqwest` in possibly different configurations. Possibly some of which are unchecked across unsafe boundaries (ffi).

PR #2798, opened 2024-08-18
core: bring tokio to core

Parth:
# What
+ introduce tokio
+ allow more of core to explicitly manage locks
+ cleanup lb-rs
+ cleanup ffi

# Why

tldr: **prep lb-rs for the changes expected over the next 12 months**

## Things benefiting from parallelism: 
+ usage calculations
+ mostly any interactions with the server (push / pull files in parallel)
+ searches
+ import / export

## There's also more and more opportunities for background handling of tasks:
+ speculatively fetching files based on platform
+ kicking off an index rebuild after a file has been edited
+ cleaning up old versions of files when the runtime is idle (instead of blocking a write for it)

## Other tokio benefits
Introducing a runtime also has some broad observability and stability benefits, even panicking code can be reliably recovered from in rust and wrapped in an `UnexpectedError`. Presently the app can detect and log this but not recover from it, the app will shut down. This keeps the app alive and lets us handle these scenarios gracefully. If this approach can be extended to workspace (not sure if overhead is prohibitive) then we'd know that all app crashes come from swift. This is one of the largest remaining unknowns for our customer crash debuggings sessions as the Apple-provided-dialogs and information usually lack the context we need to determine what happened. 

## Code cleanup benefits
Giving core the ability to manage it's own locks brings the benefits of concurrent syncs to more places. Generally the principle of don't hold a lock while talking to the network vastly improves the responsiveness of the app under many circumstances.

This code cleanup also allows us to reconsider the locks itself, using RWLocks allows a big category of read only operations to be completed in parallel.

The code cleanup also necessitates updating FFI so this PR addresses some longstanding tech debt around `lb_external_interface` (a json C API?!), and `c_interface_v2` an incomplete but fundamentally better approach to ffi. The current FFI forces the swift engineer to reimplement all the types in Swift. The proper use of FFI should allow them to depend much more on the code generation performed by `cbindgen`. For instance all error variants will be available as constants, as will a `File` structure.

Cleaning up our FFI in this manner is a prerequisite to a more sound (unexpected breaking changes, memory safety, and race coditions) ffi strategy overall. Fixing ffi also changes the contract around errors which will allow us to cleanup errors in the future (richer errors in rust, and more error work done in rust for ui developers).

# Merge blockers

+ [x] `master` is stable from #2779 
+ [x] rebase changes from #2779 
+ [x] lb_v3 compiles
+ [x] lb_v3 important tests pass
+ [x] lb_v3 all tests pass
+ [x] make cli compile
+ [x] make egui compile
+ [x] make workspace compile
+ [x] address fundemental problem with delete_unreferenced_file_versions
+ [x] lb-c
+ [x] lb-swift
+ [x] lb-jni
+ [x] lb-kt
+ [x] Apple QA
+ [ ] Android QA

# issues closed

fixes: https://github.com/lockbook/lockbook/issues/2371
fixes: https://github.com/lockbook/lockbook/issues/2431
fixes: https://github.com/lockbook/lockbook/issues/2538
fixes: https://github.com/lockbook/lockbook/issues/2544
fixes: https://github.com/lockbook/lockbook/issues/2629
fixes: https://github.com/lockbook/lockbook/issues/2899
fixes: https://github.com/lockbook/lockbook/issues/3000
fixes: https://github.com/lockbook/lockbook/issues/2378

# Release implications

This is a massive change and will probably require about a week of dog-fooding. At worst we should see easy to fix crashes due to the new way ffi functions. As such I have enabled debug info for release. As a fast follow I will expand debug info to include "last_crash" as I've wanted to for a while for quick fixes.

An additional area of risk is the increased possibility of deadlock due to manually managed locks. I've added some commented out code in repo/mod.rs incase you come across any of these instances. This code will timeout and generate a stacktrace. Armed with a stack trace, these deadlocks are pretty easy to find.

Fixing android is not a blocker for this PR to be merged. Android will get fixed incrementally while dogfooding happens and will possibly skip a release if needed.


libs/lb/lb-rs/tests/import_export_file_tests.rs
```diff
@@ -78,9 +83,11 @@ fn export_file_successfully() {
         path_copy.lock().unwrap().push(info);
     };
     core.export_file(file.id, tmp_path.clone(), false, Some(Box::new(export_progress.clone())))
+        .await
         .unwrap();
+    // todo(parth): fix clippy warning await_holding_lock
```
tvanderstad, line 88: note

tvanderstad:
integration tests handled ✅

tvanderstad:
https://github.com/lockbook/lockbook/issues/3000

Parth:
I implemented the pull portion of this (because it's easy now) #1940 to see what the sort of gains would be for me: 

https://github.com/user-attachments/assets/d20f137d-f6d8-4402-ba0d-08a5c7a159d5

https://github.com/user-attachments/assets/b3d90ad9-9bd7-416a-9dcc-6b5d46f4f5be

It's about a 4-5x improvement given my current café wifi. Could be more or less under different circumstances. I imagine when I'm plugged into ethernet it'll be dramatically more, and if I'm on LTE it'll be less. 

I think we now have the ability to mostly saturate a network connection if we want to.

I initially just spawned literally N parallel futures (the absolute naive implementation). And had lots of bad outcomes, mostly client side. I initially saturated my DNS lookup. After I added a no-regrets retry mechanism to our Client I then saturated the max number of open files.

I employed this pretty elegant solution for [limiting parallelism](https://stackoverflow.com/questions/70871368/limiting-the-number-of-concurrent-futures-in-join-all) so I don't have to implement any sort of buffered channel or a queue. (Basically what this is).

Further analysis could tell us what an appropriate value here is, but for now I default to [std available parallelism](https://doc.rust-lang.org/beta/std/thread/fn.available_parallelism.html), which seems like a reasonable default to me.

clients/apple/Shared/Views/AppView.swift
```diff
@@ -35,28 +35,20 @@ struct AppView: View {
             }
         }
             .alert(isPresented: Binding(get: { errors.globalError != nil }, set: { _ in errors.globalError = nil })) {
-                // TODO: Improve the UX of this
```
Parth, line 38: 4 year old todo
image

PR #2809, opened 2024-08-21
workspace tracing

### LINKED ISSUE #2790: Canvas tracing
build or buy tracing tools to figure out the performance breakdown

ad-tra:
fixes #2790
better observability for canvas. currently most of the added spans are traces that can be added to your lockbook.log if you ran the app with a trace log level env variable.

### DIFF
--- a/libs/content/workspace-ffi/Cargo.toml
+++ b/libs/content/workspace-ffi/Cargo.toml
@@ -18,6 +18,7 @@ lb_external_interface = { path = "../../lb/lb_external_interface" }
 raw-window-handle = "0.6"
 jni = "0.21.0"
 pollster = "0.2"
+tracing="0.1.5"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0.44"
 
--- a/libs/content/workspace-ffi/src/apple/ios/api.rs
+++ b/libs/content/workspace-ffi/src/apple/ios/api.rs
@@ -2,7 +2,7 @@ use egui::{Event, Key, Modifiers, PointerButton, Pos2, TouchDeviceId, TouchId, T
 use std::cmp;
 use std::ffi::{c_char, c_void, CStr, CString};
 use std::ptr::null;
-use tracing::{span, Level};
+use tracing::instrument;
 use workspace_rs::tab::markdown_editor::input::canonical::{
     Bound, Increment, Modification, Offset, Region,
 };
@@ -17,19 +17,14 @@ use workspace_rs::tab::TabContent;
 use super::super::response::*;
 use super::response::*;
 use crate::apple::keyboard::UIKeys;
-use crate::{ExtendedInput, WgpuWorkspace};
+use crate::WgpuWorkspace;
 
 #[no_mangle]
-pub extern "C" fn ios_frame(obj: *mut c_void) -> IOSResponse {
+#[instrument(level="trace", skip(obj) fields(frame = (&mut *(obj as *mut WgpuWorkspace)).context.frame_nr()))]
+pub unsafe extern "C" fn ios_frame(obj: *mut c_void) -> IOSResponse {
     let obj = unsafe { &mut *(obj as *mut WgpuWorkspace) };
 
-    let frame = obj.context.frame_nr();
-    let span = span!(target: "svg_editor", Level::TRACE, "request ios frame", frame);
-    let _ = span.enter();
-
     let res = obj.frame().into();
-
-    obj.context.push_span(span);
 
     res
 }
@@ -283,13 +278,9 @@ pub unsafe extern "C" fn end_of_document(obj: *mut c_void) -> CTextPosition {
 /// # Safety
 /// obj must be a valid pointer to WgpuEditor
 #[no_mangle]
+#[instrument(level="trace", skip(obj) fields(frame = (&mut *(obj as *mut WgpuWorkspace)).context.frame_nr()))]
 pub unsafe extern "C" fn touches_began(obj: *mut c_void, id: u64, x: f32, y: f32, force: f32) {
     let obj = &mut *(obj as *mut WgpuWorkspace);
-
-    let frame = obj.context.frame_nr();
-    let pos = egui::pos2(x, y);
-    let span = span!(target: "svg_editor", Level::TRACE, "UIEvent touch start", ?pos, frame);
-    let _ = span.enter();
 
     let force = if force == 0.0 { None } else { Some(force) };
     obj.raw_input.events.push(Event::Touch {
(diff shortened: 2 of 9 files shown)


libs/content/workspace-ffi/src/apple/ios/api.rs
```diff
@@ -16,12 +17,23 @@ use workspace_rs::tab::TabContent;
...
-use crate::WgpuWorkspace;
+use crate::{ExtendedInput, WgpuWorkspace};
 
 #[no_mangle]
 pub extern "C" fn ios_frame(obj: *mut c_void) -> IOSResponse {
     let obj = unsafe { &mut *(obj as *mut WgpuWorkspace) };
-    obj.frame().into()
+
+    let frame = obj.context.frame_nr();
+    let span = span!(target: "svg_editor", Level::TRACE, "request ios frame", frame);
```
smailbarkouch, line 27: Just some thoughts: should this span target here be `svg_editor` if it is around something with more scope (the entire frame)?

libs/content/workspace-ffi/src/apple/ios/api.rs
```diff
@@ -290,13 +308,23 @@ pub unsafe extern "C" fn touches_began(obj: *mut c_void, id: u64, x: f32, y: f32
...
 
 /// # Safety
 /// obj must be a valid pointer to WgpuEditor
 #[no_mangle]
 pub unsafe extern "C" fn touches_moved(obj: *mut c_void, id: u64, x: f32, y: f32, force: f32) {
     let obj = &mut *(obj as *mut WgpuWorkspace);
+
+    let frame = obj.context.frame_nr();
+    let pos = egui::pos2(x, y);
+    let span = span!(target: "svg_editor", Level::TRACE, "NSEevent touch move", ?pos, frame);
```
smailbarkouch, line 325: `UIEvent` is the right descriptor (also `NSEevent` should be `NSEvent`)

libs/content/workspace-ffi/src/apple/ios/api.rs
```diff
@@ -16,12 +17,23 @@ use workspace_rs::tab::TabContent;
...
+
+    let frame = obj.context.frame_nr();
+    let span = span!(target: "svg_editor", Level::TRACE, "request ios frame", frame);
+    let _ = span.enter();
+
+    let res = obj.frame().into();
+
+    let span_clone = span.clone();
+    drop(span);
+    obj.context.push_span(span_clone);
```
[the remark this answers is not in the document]
ad-tra: yeah that's a good callout, i don't need that hack anymore

libs/content/workspace-ffi/src/apple/ios/api.rs
```diff
@@ -16,12 +17,21 @@ use workspace_rs::tab::TabContent;
...
+use crate::{ExtendedInput, WgpuWorkspace};
 
 #[no_mangle]
 pub extern "C" fn ios_frame(obj: *mut c_void) -> IOSResponse {
     let obj = unsafe { &mut *(obj as *mut WgpuWorkspace) };
-    obj.frame().into()
+
+    let frame = obj.context.frame_nr();
+    let span = span!(target: "svg_editor", Level::TRACE, "request ios frame", frame);
+    let _ = span.enter();
```
smailbarkouch, line 28: Why not just have `span.enter()` (without the `let _ = `)

Parth: it'll get dropped and the span will close

libs/content/workspace-ffi/Cargo.toml
```diff
@@ -18,6 +18,7 @@ lb_external_interface = { path = "../../lb/lb_external_interface" }
 raw-window-handle = "0.6"
 jni = "0.21.0"
 pollster = "0.2"
+tracing="0.1.5"
```
Parth, line 21: optional:
we should `pub use` tracing out of `lb` to avoid defining this multiple times

libs/content/workspace-ffi/src/apple/ios/api.rs
Parth: I unresolved this comment, the targets are determined from the name of the crate, feels weird to break that pattern and call it svg_editor here this point is definitely more general than the SVG editor

libs/content/workspace-ffi/src/apple/ios/api.rs
```diff
@@ -275,6 +285,12 @@ pub unsafe extern "C" fn end_of_document(obj: *mut c_void) -> CTextPosition {
 #[no_mangle]
 pub unsafe extern "C" fn touches_began(obj: *mut c_void, id: u64, x: f32, y: f32, force: f32) {
     let obj = &mut *(obj as *mut WgpuWorkspace);
+
+    let frame = obj.context.frame_nr();
+    let pos = egui::pos2(x, y);
+    let span = span!(target: "svg_editor", Level::TRACE, "UIEvent touch start", ?pos, frame);
```
Parth, line 291: I think most if not all of your tracing invocations can get cleaned up by using the instrument macro like we do in LB: https://docs.rs/tracing/latest/tracing/attr.instrument.html

Includes printing of arguments automatically and other such a nice features

libs/content/workspace-ffi/src/apple/ios/api.rs
```diff
@@ -16,12 +17,21 @@ use workspace_rs::tab::TabContent;
...
     let obj = unsafe { &mut *(obj as *mut WgpuWorkspace) };
-    obj.frame().into()
+
+    let frame = obj.context.frame_nr();
+    let span = span!(target: "svg_editor", Level::TRACE, "request ios frame", frame);
+    let _ = span.enter();
+
+    let res = obj.frame().into();
+
+    obj.context.push_span(span);
```
Parth, line 32: I think you do not need to hang onto spans, I'm fairly certain all the behavior you're describing in this PR can happen automatically

libs/lb/lb-rs/src/service/log_service.rs
```diff
@@ -12,28 +12,39 @@ pub fn init(config: &Config) -> LbResult<()> {
         let lockbook_log_level = env::var("LOG_LEVEL")
             .ok()
             .and_then(|s| s.as_str().parse().ok())
-            .unwrap_or(LevelFilter::DEBUG);
+            .unwrap_or(LevelFilter::TRACE);
```
Parth, line 15: I don't think this is a mergable change, it will result in an explosion of log volume that I don't think most users want to incur.

Parth: You could consider overwriting this at configuration time any number of ways, but as it is, most users aren't going to want this, and it will force us to implement something like log rotation sooner

Parth (changes requested):
I think you should view this more like you're bringing logging to workspace rather than canvas, I think overall you should pursue a more idiomatic implementation and more flexible implementation that is suitable for more users

libs/content/workspace-ffi/src/apple/ios/api.rs
ad-tra: okay i changed it so that the target is workspace and not canvas

libs/content/workspace-ffi/src/apple/ios/api.rs
ad-tra: oh nice didn't know about that. I implemented it, thanks!

Parth (approved):

libs/content/workspace-ffi/src/apple/ios/api.rs
smailbarkouch: Right now we use the `workspace` span target name for stuff that exists in `workspace-ffi`. That feels a bit off, like there should be a separation between `workspace-ffi` and `workspace`.

Parth: they'll now show up as `workspace-ffi` and `workspace` in the logs

Parth: The filter just specifies anything that starts with workspace should end up in our logs

smailbarkouch: Okay that sounds good then.

PR #2811, opened 2024-08-22
core: 24 word secret key

smailbarkouch:


libs/lb/lb-rs/src/lib.rs
```diff
@@ -173,18 +173,42 @@ impl<Client: Requester, Docs: DocumentService> CoreLib<Client, Docs> {
             ])
     }
 
+    #[instrument(level = "debug", skip_all, err(Debug))]
```
Parth, line 176: I don't want the idea of v2 to leak up to the UI programmer as something to deal with. imo the complexity and key switching should be handled inside core

libs/lb/lb-rs/src/model/errors.rs
```diff
@@ -86,6 +86,9 @@ impl Display for CoreError {
             CoreError::InvalidAuthDetails => {
                 write!(f, "our server failed to authenticate your request, please try again")
             }
+            CoreError::KeyPhrasesMistyped => {
```
Parth, line 89: gonna revamp errors myself, so maybe we don't need to bother thinking too deep about this but Mistyped is like another word for invalid, we have a few variations of this sort of word already, idk feels weird

libs/lb/lb-rs/src/shared/account.rs
```diff
@@ -24,6 +29,97 @@ impl Account {
     pub fn public_key(&self) -> PublicKey {
         PublicKey::from_secret_key(&self.private_key)
     }
+
+    pub fn get_phrase(&self) -> SharedResult<[String; 24]> {
+        let key = self.private_key.serialize();
+        let key_bits = key.iter().fold(String::new(), |mut out, byte| {
+            let _ = write!(out, "{:08b}", byte);
```
Parth, line 36: what does this do?

libs/lb/lb-rs/src/shared/account.rs
```diff
@@ -24,6 +29,97 @@ impl Account {
...
+                .into_iter()
+                .fold(String::new(), |mut out, byte| {
+                    let _ = write!(out, "{:08b}", byte);
+                    out
+                });
+
+        let checksum_last_4_bits = &checksum[..4];
+        let combined_bits = format!("{}{}", key_bits, checksum_last_4_bits);
+
+        let mut phrase: [String; 24] = std::array::from_fn(|_| String::new());
```
Parth, line 51: I wonder if `= Default::default()` would just work here

libs/lb/lb-rs/src/lib.rs
```diff
@@ -173,18 +173,42 @@ impl<Client: Requester, Docs: DocumentService> CoreLib<Client, Docs> {
...
+    }
+
     #[instrument(level = "debug", skip_all, err(Debug))]
     pub fn export_account_qr(&self) -> Result<Vec<u8>, LbError> {
         self.in_tx(|s| s.export_account_qr())
             .expected_errs(&[CoreError::AccountNonexistent])
     }
 
+    #[instrument(level = "debug", skip_all, err(Debug))]
+    pub fn export_account_qr_v2(&self) -> Result<Vec<u8>, LbError> {
```
Parth, line 207: imo our qr code should just be base64 repr of the key and we're just going to support both perpetually

Parth (changes requested):
glad you got this working, excited to see this merged and see what my words are

did a first pass and left some structural comments, I think we want lb to handle key switching

libs/lb/lb-rs/src/shared/account.rs
smailbarkouch: `fold` lets you accumulate elements in an `Iterator`. Here I used it to transform the phrases into a byte string (and I do the same in my other use case). In my new commit, I removed this usage for a more concise method.

libs/lb/lb-rs/src/model/errors.rs
smailbarkouch: I switched the name.

libs/lb/lb-rs/src/shared/account.rs
smailbarkouch: I did this instead of collecting `format!(...)`s since Clippy gave me a warning (`fold` is more optimal).

smailbarkouch:
@Parth ready for review again. Do you think it would be better to have separate functions for exporting accounts, or use an enum to specify (like for import).

libs/lb/lb-rs/src/shared/account.rs
Parth: I mean `:08b`

libs/lb/lb-rs/src/service/account_service.rs
```diff
@@ -47,7 +49,14 @@ impl<Client: Requester, Docs: DocumentService> CoreState<Client, Docs> {
         Ok(account)
     }
 
-    pub(crate) fn import_account(&mut self, account_string: &str) -> LbResult<Account> {
+    pub(crate) fn import_account(&mut self, key: AccountKey) -> LbResult<Account> {
```
Parth, line 52: I think you misunderstood my intention. Presently the UI dev still has to try to determine if the thing the user put into the box is base64 or a 12 word phrase. 

I want core to just take a string and completely abstract that concept away from the consumer of core

Parth (changes requested):
handle key switching inside core

Parth:
I think there should be 3 export fns: phrase, compact, and qr (which is base64 as qr). enums are just going to add some hassle over ffi.

libs/lb/lb-rs/src/shared/account.rs
smailbarkouch: That formats a number into its 8 bit representation.

libs/lb/lb-rs/src/shared/account.rs
```diff
@@ -24,6 +29,96 @@ impl Account {
     pub fn public_key(&self) -> PublicKey {
         PublicKey::from_secret_key(&self.private_key)
     }
+
+    pub fn get_phrase(&self) -> SharedResult<[String; 24]> {
+        let key = self.private_key.serialize();
+        let key_bits = key.iter().fold(String::new(), |mut out, byte| {
+            let _ = write!(out, "{:08b}", byte);
```
Parth, line 36: If I understand this pipeline correctly you're going from: `[u8]` -> `String` -> `u16`. 

This is probably not the right approach, including the checksum. 

Working with bytes within db_rs and with crypto stuff I would expect to see you use one of:
- https://doc.rust-lang.org/std/primitive.u16.html#method.from_be_bytes
- https://doc.rust-lang.org/std/primitive.u16.html#method.from_le_bytes

to perform the `[u8]` to `u16` conversion.

I think the resulting implementation will be much clearer to read. And this is some of the most critical code to be as clear as possible. Everything that we do rests on this. 

Also we do want this operation to be as fast as possible for things like the fuzzer (although I can just use the compact key so possibly a moot point).

Parth (changes requested):
Expecting this to be the final round. Nice work. Very critical stuff.

libs/lb/lb-rs/src/shared/account.rs
Parth: I now understand that the algorithm requires us to take 11 bits at a time which is not as convenient as I initially expected.

libs/lb/lb-rs/src/shared/account.rs
```diff
@@ -24,6 +29,94 @@ impl Account {
...
+                .iter()
+                .fold(String::new(), |mut acc, byte| {
+                    acc.push_str(&format!("{:08b}", byte));
+                    acc
+                });
+
+        let gen_checksum_last_4 = &gen_checksum[..4];
+
+        if gen_checksum_last_4 != checksum_last_4_bits {
+            return Err(SharedErrorKind::KeyPhraseInvalid.into());
```
Parth, line 115: Just add some comments explaining 264, 253, 260, etc

Parth (approved):
tiny thing to comment, you can merge on your own after you throw the explanation of these values on there

libs/lb/lb-rs/src/shared/account.rs
```diff
@@ -24,6 +29,94 @@ impl Account {
...
+            .iter()
+            .map(|word| {
+                bip39_dict::ENGLISH
+                    .lookup_mnemonic(word)
+                    .map(|index| format!("{:011b}", index.0))
+            })
+            .collect::<Result<String, _>>()
+            .map_err(|_| SharedErrorKind::KeyPhraseInvalid)?;
+
+        if combined_bits.len() != 264 {
```
smailbarkouch, line 84: There are 24 words, and each word is 11 bits. So `24 x 11 = 264` bits after translating.

libs/lb/lb-rs/src/shared/account.rs
```diff
@@ -24,6 +29,94 @@ impl Account {
...
+            return Err(SharedErrorKind::Unexpected("the number of bits after translating the phrase does not equal the expected amount (264)").into());
+        }
+
+        for _ in 0..4 {
+            combined_bits.remove(253);
```
smailbarkouch, line 89: The 23 word ends at index 252. On 253, we have the addition 0s we don't need, 4 bits of them, so we must remove them.

Honestly though, maybe I don't need to remove them. I'll check that right now.

libs/lb/lb-rs/src/shared/account.rs
```diff
@@ -24,6 +29,94 @@ impl Account {
...
+        }
+
+        let key_bits = &combined_bits[..256];
+        let checksum_last_4_bits = &combined_bits[256..260];
```
smailbarkouch, line 93: The checksum is 4 bits and its right after the key, which is 256 bits.

Issue #2778, opened 2024-08-04
tabs should autoscroll


Parth:
open many documents tabs will keep piling on without scrolling automatically 

Requested by @Krishma-Bawa

PR #2834, opened 2024-09-13
workspace: fix tab shortcuts, scroll to active tab

tvanderstad:
fixes https://github.com/lockbook/lockbook/issues/2784
fixes https://github.com/lockbook/lockbook/issues/2778

https://github.com/user-attachments/assets/5e479060-6ddd-4ac7-94d1-d68e92282585

### DIFF
--- a/libs/content/workspace/src/workspace.rs
+++ b/libs/content/workspace/src/workspace.rs
@@ -21,6 +21,7 @@ pub struct Workspace {
 
     pub tabs: Vec<Tab>,
     pub active_tab: usize,
+    active_tab_changed: bool,
     pub backdrop: Image<'static>,
 
     pub ctx: Context,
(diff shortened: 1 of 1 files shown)


libs/content/workspace/src/workspace.rs
```diff
@@ -679,13 +691,16 @@ impl Workspace {
                     if let Some((name, id)) =
                         self.current_tab().map(|tab| (tab.name.clone(), tab.id))
                     {
-                        self.ctx.send_viewport_cmd(ViewportCommand::Title(name));
```
tvanderstad, line 682: this froze bc `send_viewport_cmd` ultimately tries to lock the same lock as `input_mut`

smailbarkouch (approved):

Issue #2853, opened 2024-09-19
gesture support


Parth:
either iOS gestures or procreate gestures for:
+ undo / redo
+ possible that each tool bar item, in addition to kb shortcuts have some gesture associated with them. Could be cool if I could always access "red" by tapping 3 fingers, and "green" by tapping 4 fingers, etc.

PR #2864, opened 2024-09-24
canvas: add shortcut gestures

### LINKED ISSUE #2851: zoom doesn't work as expected
there's some error in the math. draw two circles, put your fingers on those two circles and zoom around. your fingers should stay on the circles. Right now there's some weird pan in addition to good feeling zoom.

ad-tra:
fixes: #2853 and fixes #2851

two fingers to undo, three to redo


Parth (changes requested):
My ideal behavior here would be:
- on release undo happens automatically giving me the power to spam if i want to
- on hold acceleration behavior begins, although I am not at all tied to the accel behavior and would be fine if there was no accel.
- zoom feels good

PR #2911, opened 2024-10-08
editor: fix collaborative editing self-conflict

### LINKED ISSUE #2909: collaborative editing erroneous self-conflict
- new doc
- type `test`
- quickly open an existing doc that doesn't have a tab present
- switch back to tab
- content is `testtes`

tvanderstad:
fixes #2909

### DIFF
--- a/libs/content/workspace/src/tab/mod.rs
+++ b/libs/content/workspace/src/tab/mod.rs
@@ -26,28 +26,30 @@ pub struct Tab {
     pub last_saved: Instant,
 }
 
+#[derive(Debug, Default)]
 pub struct SaveRequest {
     pub id: lb_rs::Uuid,
-    pub hmac: Option<DocumentHmac>,
+    pub old_hmac: Option<DocumentHmac>,
+    pub seq: usize,
     pub content: String,
+    pub safe_write: bool,
 }
 
 impl Tab {
     pub fn make_save_request(&self) -> Option<SaveRequest> {
-        let mut hmac = None;
-        if let Some(tab_content) = &self.content {
-            let maybe_save_content = match tab_content {
-                TabContent::Markdown(md) => {
-                    hmac = md.editor.hmac;
-                    Some(md.editor.buffer.current.text.clone())
-                }
-                TabContent::Svg(svg) => Some(svg.get_minimal_content()),
-                _ => None,
-            };
-            maybe_save_content.map(|content| SaveRequest { id: self.id, content, hmac })
-        } else {
-            None
-        }
+        let tab_content = self.content.as_ref()?;
+        let mut result = SaveRequest { id: self.id, ..Default::default() };
+        match tab_content {
+            TabContent::Markdown(md) => {
+                result.old_hmac = md.editor.hmac;
+                result.seq = md.editor.buffer.current.seq;
+                result.content = md.editor.buffer.current.text.clone();
+                result.safe_write = true;
+            }
+            TabContent::Svg(svg) => result.content = svg.get_minimal_content(),
+            _ => return None,
+        };
+        Some(result)
     }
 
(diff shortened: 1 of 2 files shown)


libs/content/workspace/src/tab/mod.rs
```diff
@@ -26,28 +26,30 @@ pub struct Tab {
...
-            None
-        }
+        let tab_content = self.content.as_ref()?;
+        let mut result = SaveRequest { id: self.id, ..Default::default() };
+        match tab_content {
+            TabContent::Markdown(md) => {
+                result.old_hmac = md.editor.hmac;
+                result.seq = md.editor.buffer.current.seq;
+                result.content = md.editor.buffer.current.text.clone();
+                result.safe_write = true;
```
tvanderstad, line 47: The new version of this function contains an explicit flag for `safe_write`, set whenever the content is markdown, which avoids confusion when inferring write safety based on the presence of an hmac.

libs/content/workspace/src/workspace.rs
```diff
@@ -540,24 +540,16 @@ impl Workspace {
...
                     thread::spawn(move || {
-                        let content = save_req.content;
-                        let id = save_req.id;
-                        let hmac = save_req.hmac;
+                        let SaveRequest { seq, content, id, old_hmac, safe_write } = save_req;
 
-                        let result = if hmac.is_some() {
-                            core.safe_write(id, hmac, content.clone().into())
-                                .map(|hmac| (content, Some(hmac), Instant::now(), seq))
+                        let result = if safe_write {
```
tvanderstad, line 547: On master, when the initial content was empty, the document would be written unsafely. This is why the issue is only reproducible for new files before they have been saved. This issue is fixed using the explicit safe_write flag.

libs/content/workspace/src/workspace.rs
```diff
@@ -540,24 +540,16 @@ impl Workspace {
...
+                            core.safe_write(id, old_hmac, content.clone().into())
+                                .map(|new_hmac| (content, Some(new_hmac), Instant::now(), seq))
                         } else {
-                            // todo: what if editor opened a file with no hmac, then we overwrite outside edits here
                             core.write_document(id, content.as_bytes())
-                                .map(|_| (content, hmac, Instant::now(), seq))
+                                .map(|_| (content, None, Instant::now(), seq))
```
tvanderstad, line 552: In the old version, where the variables here were all named `hmac`, an unsafe write would pass on the old hmac in place of the new hmac which would have been returned by `core.safe_write()`. More descriptive variable names highlighted this issue.

This isn't related to the bug, though, because it causes the `SaveResult` to contain a `None` hmac on the first write which is the correct behavior.

libs/content/workspace/src/workspace.rs
```diff
@@ -786,14 +778,12 @@ impl Workspace {
                         match result {
                             Ok((content, hmac, time_saved, seq)) => {
                                 tab.last_saved = time_saved;
-                                if hmac.is_some() {
```
tvanderstad, line 789: On master, the editor would find out that the content had been saved (update to `md.editor.hmac` and call to `md.editor.buffer.saved()`) only if `hmac.is_some()`. On the initial write, `hmac` is `None`, and the editor is unaware that content has been saved. On the next load, the editor proceeds as if the version last agreed upon by editor and disk is the initial empty file contents, and therefore that the loaded content is a result of a concurrent edit. This triggers the conflict resolution logic when it shouldn't have been triggered at all.

The reason for the reload is unclear but it's known not to be a safe-write failure/re-read required. It's also unclear why this is more reproducible for Parth than for Travis, with Parth reproducing sometimes on the first attempt and Travis needing sometimes over 200 attempts.

Issue #2942, opened 2024-10-14
context menu doesn't open on Android


tvanderstad:

tvanderstad:
@smailbarkouch can you remind me the contract here?

On Apple Notes, double-tapping to select text opens a context menu, then tapping it again closes it. Should the editor indicate a context  menu in its response every frame, or just on the frame its open? In other words, is editor responsible for signaling when to close the menu?

smailbarkouch:
@tvanderstad  The editor should only indicate the frame in which the context menu was opened. I will open it that frame and keep it open, and I'll determine when it should be closed.

PR #2949, opened 2024-10-15
android: fix context menu

tvanderstad:
fixes: https://github.com/lockbook/lockbook/issues/2942

qa:
- [x] double tap opens a context menu in addition to selecting the tapped word
- [x] tapping the selection opens a context menu whether the selection is empty or not


smailbarkouch:
@tvanderstad it works, great job! Just a little nitpik: the context menu appears where you touched rather than next to the selected text. It feels a bit odd and isn't default behavior for context menus.

PR #2959, opened 2024-10-16
editor: smaller toolbar

tvanderstad:
fixes https://github.com/lockbook/lockbook/issues/1816
fixes https://github.com/lockbook/lockbook/issues/1933
fixes https://github.com/lockbook/lockbook/issues/1968
fixes https://github.com/lockbook/lockbook/issues/2954
fixes https://github.com/lockbook/lockbook/issues/2956
fixes https://github.com/lockbook/lockbook/issues/2933 by removing the setting
fixes https://github.com/lockbook/lockbook/issues/2935 by removing the setting
fixes https://github.com/lockbook/lockbook/issues/2896 by removing the floating toolbar

Screenshot 2024-10-16 at 5 40 24 PM
Screenshot 2024-10-16 at 5 39 55 PM
Screenshot 2024-10-16 at 5 43 39 PM
<video alt="Screenshot 2024-10-16 at 5 43 39 PM" src="https://github.com/user-attachments/assets/f9fee1e0-4ffa-478d-bed3-65b0e1b995f3">

PR #2960, opened 2024-10-16
workspace: prettier tab strip

tvanderstad:
fixes https://github.com/lockbook/lockbook/issues/2955

Screenshot 2024-10-16 at 7 33 25 PM
Screenshot 2024-10-16 at 7 34 46 PM

### DIFF
--- a/libs/content/workspace-ffi/src/lib.rs
+++ b/libs/content/workspace-ffi/src/lib.rs
@@ -58,17 +58,10 @@ impl<'window> WgpuWorkspace<'window> {
         self.raw_input.time = Some(self.start_time.elapsed().as_secs_f64());
         self.context.begin_frame(self.raw_input.take());
 
-        let workspace_response = {
-            let fill = if self.context.style().visuals.dark_mode {
-                egui::Color32::BLACK
-            } else {
-                egui::Color32::WHITE
-            };
-            egui::CentralPanel::default()
-                .frame(egui::Frame::default().fill(fill))
-                .show(&self.context, |ui| self.workspace.show(ui))
-                .inner
-        };
+        let workspace_response = egui::CentralPanel::default()
+            .frame(egui::Frame::default().fill(self.context.style().visuals.window_fill))
+            .show(&self.context, |ui| self.workspace.show(ui))
+            .inner;
 
         let full_output = self.context.end_frame();
 
--- a/libs/content/workspace/src/workspace.rs
+++ b/libs/content/workspace/src/workspace.rs
@@ -903,7 +903,7 @@ fn tab_label(
 
     let wrap_width = ui.available_width();
     let text: egui::WidgetText = (&t.name).into();
-    let text = text.into_galley(ui, Some(TextWrapMode::Extend), wrap_width, egui::TextStyle::Body);
+    let text = text.into_galley(ui, Some(TextWrapMode::Extend), wrap_width, egui::TextStyle::Small);
     let x_icon = Icon::CLOSE.size(16.0);
 
     let padding_x = 15.;
--- a/libs/content/workspace/src/workspace.rs
+++ b/libs/content/workspace/src/workspace.rs
@@ -901,13 +901,10 @@ fn tab_label(
 ) -> Option<TabLabelResponse> {
     let mut result = None;
 
-    let wrap_width = ui.available_width();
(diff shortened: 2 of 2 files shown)


ad-tra (approved):
this is sick. THANK YOU, finally getting rid of the old ugly tab strip.
LGTM but my two nits are:
- i think the text can be made smaller to further decrease the height of the toolbar.  this tab strip is thinner than the old one but still thicker than a lot of common apps like vscode and chrome. 
- the closing hover rect is quite ugly, it would be nice to fix it too. it's up to you if you wanna do that or not tbh. image

tvanderstad:
@ad-tra How's this look to you?

Screenshot 2024-10-16 at 10 01 13 PM

tvanderstad:
I made the text smaller but I like the tab strip height because it matches the [toolbar](https://github.com/lockbook/lockbook/pull/2959)

Screenshot 2024-10-16 at 10 04 54 PM

Parth:
nice, I think this is def better than what we have now

the tabs look super wide to me, how does vscode (or any other app you like) handle this? 

There's also something about the colors that give me a bit of "dated" vibes trying to think about what we can do about that

tvanderstad:
VSCode just makes the text smaller. FWIW, the screenshots have the app sized at 1/4 of my screen, so they may look wider than they really are (they're less wide than currently on master)

tvanderstad:
Chrome wraps the text so all the tabs are the same width which is nice if you want to close a bunch of tabs by clicking in the same place over and over

Parth:
uniformly sized tabs seems appealing, we can have some rules for a small amount of tabs

smailbarkouch:
Looks great! My only nit is that the toolbar isn't the same color as the toolbars for the PDF viewer or Canvas. It looks a bit odd.

For instance:
<details>

image
</details>

BuriedInTheGround:
Much cleaner, I love it! (The reviewed closing hover rectangle is definitely better with the lighter gray.)

### About horizontal size

I like uniformly sized tabs, with text fading on the left (possibly), like in Chrome or Firefox.

However, I think it's much easier to have many tabs in Lockbook than in a browser, especially if you don't bother closing them, so they may shrink quite fast and become very tiny (and basically indistinguishable since you can't read the titles). For this reason, I think that fixed sized tabs (I suggest around ~25 characters) might be a better option, along with keeping the current scrolling behavior.

Personally, I'd prefer uniformly sized tabs since scrolling horizontally annoys me, but I think that some sort of auto-close rule or something alternative should be implemented to deal with a fast-growing number of open tabs in an automatized manner.

smailbarkouch:
Another note, when you scroll down into a markdown document, the divider under the tab strip partially disappears (except on the left and right ends).

For instance:
<details>

image
</details>

tvanderstad:
> Another note, when you scroll down into a markdown document, the divider under the tab strip partially disappears (except on the left and right ends).

This issue tracked by https://github.com/lockbook/lockbook/issues/2956 and fixed by https://github.com/lockbook/lockbook/pull/2959 👍

tvanderstad:
In this iteration, tabs are a fixed width regardless of how many there are or how long their titles are. The text fills the tab and is truncated with '...'. When the close button is shown, it's truncated a little harder to make space for the button. If you add enough tabs, you still get a scrollbar (I don't love Chrome's behavior where the tabs get smaller and smaller). I do like how Chrome fades the text out instead of using '...' but this is what's built-in to egui so I think this is the effort it deserves for now.

For scale, I took this screenshot with the window at 1/2 width and 1/2 height of my screen. I'm on macOS with the smallest display scaling setting. The tab width here is 160px so there's no issue having 10 of them open if your window is maximized.
<details>
Screenshot 2024-10-17 at 5 24 04 PM
</details>

BuriedInTheGround:
I like the result and I think the sizing looks fine. Also, now that I see it, I don't really mind '...' that much, and I'm not sure that the fading would fit the overall look (tough I also like it on my browser).

Would it be possible to display the full title when hovering a tab? (See the screenshot below, on Firefox. You don't see my pointer, but it's there, my screenshotting tool hides it.)

image

tvanderstad:
> Would it be possible to display the full title when hovering a tab?

🫡 
Screenshot 2024-10-17 at 6 33 14 PM

Issue #2958, opened 2024-10-16
markdown code block refinements


tvanderstad:
- add a background
- don't change font color
- highlight language keywords (best effort) tracked by https://github.com/lockbook/lockbook/issues/2969
- copy button tracked by https://github.com/lockbook/lockbook/issues/2450

BuriedInTheGround:
> copy button

Is it possible to have this only when hovering the block? I suggest this to minimize the visual clutter, at least on devices with a pointer.

tvanderstad:
@BuriedInTheGround Yeah, certainly! We have a working example of that with the tab strip, where the close button for a tab only appears when hovering the tab, so should be no problem to get it working like that here as well

tvanderstad:
Speaking of visual clutter, would be great to get your feedback on some visual updates I put together yesterday:
* https://github.com/lockbook/lockbook/pull/2959 replaces the floating toolbar on desktop with the smaller one we're already using on mobile and adds features like making the 'B' for bold blue when your cursor's in bold text
* https://github.com/lockbook/lockbook/pull/2960 makes our tab strip more minimal and less cluttered
* https://github.com/lockbook/lockbook/pull/2961 gives code blocks a background so they're visually distinguished without needing to be a different font color

smailbarkouch:
@tvanderstad It's important to mention that hovers aren't communicated to the markdown editor on iPad due to some conflicts with Apple's own native cursor logic. So on that platform, you probably just want to keep it visible.

tvanderstad:
@smailbarkouch got it 👍 I don't have a pointer for my iPad so I treat it like a touch-only device by default

Issue #2450, opened 2024-02-17
editor copy code block feature


tvanderstad:
<img width="949" alt="Screenshot 2024-02-17 at 9 58 47 AM" src="https://github.com/lockbook/lockbook/assets/6198756/7ea4a28f-403b-4618-b38d-3b402088dcab">

Issue #2732, opened 2024-07-07
Markdown++


Parth:
I think we could exploring building on the markdown spec. Ideally in a way that degrades gracefully outside of Lockbook.

+ We could consider conforming closely with [Github Flavored Markdown](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax). Several solved problems in this domain: `hidden`, `img` metadata, [diagraming language](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-diagrams). Obviously a massively popular standard.
+ Would also be down to add little touches that make sense for our platform:
  + If todo is the only text inside a bold region down to make it louder (red background or something).
  + Can give "mentions" of other uses special support.

Parth:
Identifying this specification could be the basis for creating a parser

tvanderstad:
Github uses a [library](https://github.com/github/markup?tab=readme-ov-file#markups) that uses a [library](https://github.com/gjtorikian/commonmarker?tab=readme-ov-file) that uses a Rust [library](https://github.com/kivikakk/comrak) that has first-class support/interoperability with a [library](https://github.com/trishume/syntect) that I identified as probably the path forward for context-free syntax highlighting anyway. Those last two libraries are both Rust. Things are looking up for Markdown++.

Parth:
https://discord.com/channels/1014184997751619664/1026211549678936214/1297023224118378556

Kousay-Jebir:
intended to be fixed: 
"    a\r\na" a string example that causes the editor to crash related to parsing \r\n
(not anything that cotnains \r\n causes the crash)
found when pasting this in a markdown file

PR #2961, opened 2024-10-17
editor: pretty code blocks & block quotes

tvanderstad:
fixes https://github.com/lockbook/lockbook/issues/2450
fixes https://github.com/lockbook/lockbook/issues/2957
fixes https://github.com/lockbook/lockbook/issues/2958

<details>
Screenshot 2024-10-16 at 9 52 47 PM

Screenshot 2024-10-16 at 9 52 33 PM
</details>

### DIFF
--- a/libs/content/workspace/src/tab/markdown_editor/appearance.rs
+++ b/libs/content/workspace/src/tab/markdown_editor/appearance.rs
@@ -181,9 +181,9 @@ impl Appearance {
             return CaptureCondition::Never;
         }
         match node_type {
-            MarkdownNodeType::Block(BlockNodeType::ListItem(_)) => CaptureCondition::Always,
+            MarkdownNodeType::Block(BlockNodeType::ListItem(_))
+            | MarkdownNodeType::Block(BlockNodeType::Quote) => CaptureCondition::Always,
             MarkdownNodeType::Block(BlockNodeType::Heading(_))
-            | MarkdownNodeType::Block(BlockNodeType::Quote)
             | MarkdownNodeType::Block(BlockNodeType::Code)
             | MarkdownNodeType::Block(BlockNodeType::Rule)
             | MarkdownNodeType::Inline(InlineNodeType::Code)
--- a/libs/content/workspace/src/tab/markdown_editor/appearance.rs
+++ b/libs/content/workspace/src/tab/markdown_editor/appearance.rs
@@ -129,7 +129,7 @@ impl Appearance {
     }
 
     pub fn code(&self) -> Color32 {
-        self.code.unwrap_or(PINK).get(self.current_theme)
+        self.code.unwrap_or(BLACK).get(self.current_theme)
     }
 
     pub fn bold(&self) -> Color32 {
--- a/libs/content/workspace/src/tab/markdown_editor/ast.rs
+++ b/libs/content/workspace/src/tab/markdown_editor/ast.rs
@@ -577,7 +577,7 @@ impl AstTextRange {
             MarkdownNode::Block(BlockNode::Rule) => Some(Annotation::Rule),
             MarkdownNode::Block(BlockNode::Quote) => Some(Annotation::BlockQuote),
             MarkdownNode::Block(BlockNode::Code) => {
-                Some(Annotation::CodeBlock { start: node.range.start() })
+                Some(Annotation::CodeBlock { text_range: node.text_range })
             }
             _ => None,
         }
--- a/libs/content/workspace/src/tab/markdown_editor/ast.rs
+++ b/libs/content/workspace/src/tab/markdown_editor/ast.rs
@@ -563,7 +563,8 @@ impl AstTextRange {
     }
 
     pub fn annotation(&self, ast: &Ast) -> Option<Annotation> {
-        match self.node(ast) {
+        let node = &ast.nodes[self.ancestors.last().copied().unwrap_or_default()];
+        match node.node_type.clone() {
             MarkdownNode::Block(BlockNode::Heading(HeadingLevel::H1)) => {
                 Some(Annotation::HeadingRule)
             }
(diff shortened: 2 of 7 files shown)


BuriedInTheGround:
### Block quotes

I like the vertical line, as it distinguishes the block quote content from italic text. My only nit is with the color: I think a shade of gray would be a better choice, as I would reserve colors for [alerts](https://github.com/orgs/community/discussions/16925) from [GitHub Flavored Markdown](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax) (#2732).

### Code blocks

I appreciate the background (I would've asked for it in the near future!).

On a different note, I think the first line of a code block (touching the opening backticks) should be hidden when the cursor (and pointer) is outside, or at least displayed differently, as it usually indicates the "type" of code in the block.

code block example

tvanderstad:
I'm persuaded by your argument for gray block quote lines! Here's a demo of the code block copy button that only shows when hovered. I made the block quote lines gray as well. 

<details>

https://github.com/user-attachments/assets/61a75556-bc45-43b8-985e-8253c16a6813

</details>

The button is smaller than GitHub's button because theirs looks strange when the code block is empty:
```
```

tvanderstad:
> On a different note, I think the first line of a code block (touching the opening backticks) should be hidden when the cursor (and pointer) is outside, or at least displayed differently, as it usually indicates the "type" of code in the block.

When the syntax is hidden we call that 'character capture. It's received some improvements that aren't in the released version yet ([PR with demo](https://github.com/lockbook/lockbook/pull/2902)) but I don't think those address your ask. Since the update, the behavior is that whenever the cursor is in text in some markdown node, whether that's the syntax for the node or the text itself, the syntax for the node isn't captured. Some examples:
* `| **text**`: the syntax is captured because the cursor is outside
* ` |**text**`: the syntax isn't captured because the cursor borders the syntax
* ` **te|xt**`: the syntax isn't captured because the cursor is inside the markdown node

For code blocks the behavior hasn't received any special attention, so similar examples apply:
1. captured:
```
|
'''
text
'''
```
2. not captured:
```

|'''
text
'''
```
3. not captured:
```

'''
te|xt
'''
```

So the behavior is basically defined in terms of what cursor placements capture the syntax or don't. Do you think it'd be better for the syntax to be captured in example 3? How about the following border cases?

4.
```

'''
|text
'''
```
5.
```

'''
text|
'''
```

BuriedInTheGround:
> Do you think it'd be better for the syntax to be captured in example 3?

I don't get your question here. Do you mean like "language tag" instead of "syntax" (e.g. "rust")?

As for the examples 4 and 5, I think they should behave like 2 and 3. Even better, I think that "rust", in the examples here, when captured should either become hidden together with the backticks or become a badge indicating the language (see image to get an idea, even though in that position it would overlap with the copy button).

code snippet

### Example

#### Not captured
``` 
'''rust
te|xt
'''
```
Thus, it looks like (i.e. no change)
``` 
'''rust
te|xt
'''
```

#### Captured (hidden)
``` 
|
'''rust
text
'''
```
Thus, it looks like
``` 
|
text
```

#### Captured (badge)
``` 
|
'''rust
text
'''
```
Thus, it looks (more or less) like
``` 
|
[Rust]
text
```
And, if the language tag is empty, it falls back to look like
``` 
|
text
```

tvanderstad:
> I don't get your question here. Do you mean like "language tag" instead of "syntax" (e.g. "rust")?

I mean both the language tag ("rust") and the syntax (`'''`). When your cursor is in the code block like you're typing the code, would you prefer to see:
1. uncaptured
```
'''rust
text|
'''
```
2. captured, with a badge for the language
```
text|
```

BuriedInTheGround:
Oh, I get it, while typing the code I'd prefer to see the uncaptured version! :)

tvanderstad:
Another note on the layout: at one point we captured the whole first and last line of the syntax (`'''`) for a code block, but we changed our minds about it. The syntax is captured based on where the cursor is but you can also hover the pointer over a markdown node to reveal its syntax. With the whole lines being captured, the text would jump around as you moused over the document and lines were captured/uncaptured. Sometimes capturing a line changed what you were hovering, which caused the line to no longer be captured, and the capture would enter a loop alternating captured and not captured every frame... it was a whole mess haha.

```
''' <- these get captured but the line stays where it is
code
''' <- also captured
```

So it's easier for us to leave the line there even when the syntax is captured. That also leaves space at the top of the code block where we can put the language badge, which I think makes a lot of sense because that's approximately where it is in the source text. When you click on the badge to edit it and the syntax and language are un-captured, you'll be editing the language basically where it just was when it was captured and rendered as a badge.

I'm being a little wordy but here's a summary of what I have in mind:
* When your cursor is in the first or last line of a code block (the lines with `'''rust` or `'''`), we don't capture the text
```
'''rust|
code
'''
```
* When your cursor is in the code in the code block, I haven't decided, I'm open to capturing or not capturing the text (see comment above)
* When your cursor is outside the code block, the syntax and language will be captured but the lines will still be there as part of the code block. If there's a language, it'll be displayed on the top left in a badge. Clicking here will place the cursor on this line, un-capturing the syntax and language based on the behavior in the first bullet
```
[rust badge]
code

```

BuriedInTheGround:
I totally understand the problems you had with removing whole lines, so the current approach of keeping them empty is good and clear in its behavior in my opinion.

I like your direction! To me, when the cursor is in the code inside the code block, the behavior should match the one of other markup elements (e.g. bold multi-line text), with the language tag being considered part of the markup syntax (and thus getting captured along with backticks when they do). This would kind of contrast with my answer from earlier, if we consider the current behavior (it could be fine, though).

tvanderstad:
Great! I'll aim to merge this basically as-is. It'll circulate among the team for feedback and QA and go out in the next release in ~2 weeks. We'll discuss the capture rules for code blocks too and see what feels right

PR #3013, opened 2024-10-23
lbrs: reduce log volume by 99%

tvanderstad:
Looking at my own logs for the egui client, only 1% of characters remain after removing lines containing "DEBUG get":
```
~/.lockbook/egui ❯ cat uncolored-2024-10-23.log | wc -c
 59965612
~/.lockbook/egui ❯ grep -v -E "DEBUG get" uncolored-2024-10-23.log | wc -c
  697974
```

More analysis is required to demonstrate the bold claim of the PR title. I don't even see a fn `get` in lb-rs/lib.rs though I've been using other branches while developing recently.


tvanderstad:
Probably the frequent calls to the log-producing function are the issue; will revisit with a different solution

PR #3023, opened 2024-10-28
editor: visual refinements to paragraph spacing and checkboxes

tvanderstad:
<details>

Screenshot 2024-10-28 at 11 44 35 AM
Screenshot 2024-10-28 at 11 45 03 AM
</details>


ad-tra:
nit: love the new design of the checkbox, but i think  it would look even better if you add more padding between the check and container rect.
image

Issue #753, opened 2021-05-22
Apple spell check


Parth:
Does not spell check on old parts of the document, or copy and pasted parts of the document

Parth:
Requested by @krishma-bawa

smailbarkouch:
Implementing a [native cursor](https://developer.apple.com/documentation/appkit/text_display/adopting_the_system_text_cursor_in_custom_text_views) for macOS may lead to one solution. (it may integrate with [NSSpellChecker](https://developer.apple.com/documentation/appkit/nsspellchecker))

Parth:
Yeah I personally would like some smart interactions with existing dictionaries on platforms where it makes sense. @smailbarkouch there's a bunch of intelligence suggestions built into iOS & macOS these days (tab to complete the rest of your sentence). Do you know if these APIs are exposed to us in any capacity?

smailbarkouch:
Any plans for this

Parth:
we talk about this pretty often, I too want this. You think it's top 3 for you?

smailbarkouch:
Just seeing this - but I'd say so

PR #3038, opened 2024-10-29
editor: grammar checking POC with Harper

tvanderstad:
fixes https://github.com/lockbook/lockbook/issues/753

Demo:
<details>

https://github.com/user-attachments/assets/b65c4a73-4af4-4010-8cd4-06661abc7975
</details>

notes:
* ~~has some notable [false positives](https://github.com/elijah-potter/harper/issues/255)~~ (fixed!)
* ~~has notable [performance issues](https://github.com/elijah-potter/harper/issues/257)~~
* not integrated with system dictionary (probably an 'us' problem)
* does not ignore any parts of the document as mentioned in #753 (probably an 'us' problem)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -237,7 +237,7 @@ dependencies = [
  "ndk-context",
  "ndk-sys 0.5.0+25.2.9519653",
  "num_enum",
- "thiserror",
+ "thiserror 1.0.64",
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


elijah-potter:
Great to hear! If it helps at all, `IgnoredLints` is designed to be (de)serialized with `serde`. That's how we save lints in the Harper Obsidian plugin.

Parth:
I'm going to close both the spell check PRs (#3046). Seems like perf with Harper was acceptable but some unknowns remain:
- would be nice to be able to remember words added to the dictionary
- would be nice to see if we can interact with the system dictionary

Travis has an idea for a revision mode which is perhaps more tolerant to seeing red in words that are not errors, I think we could think about that more. I would personally be down with a right click "check spelling" and then a right click "hide spelling" as another short term avenue.

But mostly I'm closing this because #3530 is a re-write of the editor leaving significant conflicts, and we'll probably need to revisit the implementation anyways.

I would like to have spell check soon, maybe q4?

PR #3074, opened 2024-11-04
linux: support keymaps

### LINKED ISSUE #2982: Change of keyboard layout is not respected on Linux
On Linux, with Lockbook 0.9.11, when I use my keyboard shortcut to change the keyboard layout (e.g. from US English to Italian or vice versa) Lockbook still uses the primary layout anyway (for me, US English). This prevents me from inserting foreign characters, such as accented letters. This happens in general, e.g. both in the editor and in the text field when renaming a document. Note: This is not a keyboard shortcut problem, as switching to another application (e.g., Firefox) shows that the layout has indeed changed (also, this kind of keyboard shortcut should always have higher priority, as it is directly handled by the X server).

Parth:
<details>

<summary>Demo</summary>

https://github.com/user-attachments/assets/7517fa68-29e8-45c2-a99f-46164ff349b9
</details>

+ [x] cleanup
+ [x] spawn egui Text events
+ [ ] test

fixes #2982

### DIFF
--- a/clients/linux/src/input/key.rs
+++ b/clients/linux/src/input/key.rs
@@ -1,206 +1,188 @@
-use crate::input::{self, modifiers};
+use crate::input;
+use egui::Modifiers;
 use lbeguiapp::WgpuLockbook;
-use x11rb::{
-    protocol::xproto::{GetKeyboardMappingReply, KeyButMask},
-    xcb_ffi::XCBConnection,
-};
-use xkbcommon::xkb::{self, x11, Context, KEYMAP_COMPILE_NO_FLAGS};
+use x11rb::{protocol::xproto::KeyButMask, xcb_ffi::XCBConnection};
+use xkbcommon::xkb::{self, x11, Keycode};
+
+pub struct Keyboard {
+    state: xkb::State,
+}
+
+impl Keyboard {
+    pub fn new(conn: &XCBConnection) -> Self {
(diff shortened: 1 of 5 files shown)


Parth:
Key milestone: keyboard switching and utf-8 interpretation at runtime:

```
got map
key0: English (US) active true
key1: Italian active false
key: 47, s: ;
STATUS: true
3
got map
key0: English (US) active false
key1: Italian active true
key: 47, s: ò
STATUS: true
3
```

clients/linux/src/input/key.rs
```diff
@@ -1,158 +1,188 @@
...
+                .state
+                .mod_name_is_active(xkb::MOD_NAME_SHIFT, xkb::STATE_MODS_EFFECTIVE),
+            command: self
+                .state
+                .mod_name_is_active(xkb::MOD_NAME_CTRL, xkb::STATE_MODS_EFFECTIVE),
+            mac_cmd: false,
+        };
+
+        // todo: pressing a key then holding ctrl or alt while releasing may cause egui to think it is still pressed
+        if (modifiers.is_none() || modifiers.shift_only()) && utf8.chars().any(|c| !c.is_control())
```
tvanderstad, line 53: somewhat hacky but works well in my testing and cleaner than the alternative (checking against a hardcoded set of control chars)

Parth: Hard for me to determine the correctness of this

easy to determine the correctness of "this is how winit does it"

PR #3126, opened 2024-11-16
infra: bump rust version

### LINKED ISSUE #2381: update rust version
probably to 1.76 coming out shortly

ad-tra:
fixes #2381

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "1.77.0"
+channel = "1.82.0"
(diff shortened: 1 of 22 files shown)


Parth:
Unless it's intentional, I don't think you actually bumped the version file, there's a tool chain file in the root of the project. That's where you need to make the change if CI is happy and the version has been bumped then you're set.

rust-toolchain.toml
ad-tra, line 1: @Parth do you mean this file?

Parth: oh, idk why I didn't see this

Parth: yeah this is good

PR #3132, opened 2024-11-18
clients/egui: remove eframe as lib dependency

tvanderstad:


tvanderstad:
fixes https://github.com/lockbook/lockbook/issues/3131

PR #3160, opened 2024-11-23
all: no openssl

Parth:
try to use rustls-tls for everything. I believe we did native-tls for windows on arm but I see some language in the BUILDING.md that makes me feel like it shouldn't be a problem: https://github.com/briansmith/ring/blob/main/BUILDING.md

if we need to we can re-introduce native-tls just for windows, possibly just for windows on arm. But as it stood we were depending on native-tls for everything which is an annoying dependency to satisfy everywhere.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1556,7 +1556,7 @@ dependencies = [
  "bitflags 1.3.2",
  "core-foundation",
  "core-graphics-types",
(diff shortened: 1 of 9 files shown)


Parth:
I verified that this works on windows making sure it compiles on iOS then merging this

Issue #2604, opened 2024-04-08
persist ws tabs and corresponding file tree state across sessions


Parth:
requested by @alexwasserman

tvanderstad:
@Parth can you clarify?

Parth:
better?

tvanderstad:
yes thank you

Parth:
tangentially related #2936

Parth:
Although i said we should not close this due to file tree state, I think we should because file tree state will be synced to the open set of tabs

PR #3192, opened 2024-12-08
workspace: settings

### LINKED ISSUE #2936: workspace read settings
unblocks state persistence for disabled toolbar

ad-tra:
fixes #2936 #2604

for a start, the settings store last opened tabs info. This pr will preserve your open tabs and your active tab but it will not make any guarantees about the order of your tabs. 

## QA
- [ ] macOS
- [ ] iPad
- [ ] android
- [ ] iPhone
- [ ] linux
- [ ] windows

### DIFF
--- a/libs/content/workspace/src/workspace.rs
+++ b/libs/content/workspace/src/workspace.rs
@@ -75,25 +75,13 @@ impl Default for WsConfig {
 }
 
 impl WsConfig {
-    pub fn new(dir: String, auto_save: bool, auto_sync: bool, zen_mode: bool) -> Self {
-        let mut s = Self { path: dir, ..Default::default() };
-        s.update(auto_save, auto_sync, zen_mode);
-        s
-    }
-
-    pub fn from_file(path: PathBuf) -> Self {
+    pub fn new(path: PathBuf) -> Self {
         let mut s: Self = match fs::File::open(&path) {
             Ok(f) => serde_json::from_reader(f).unwrap_or_default(),
             Err(_) => Self::default(),
         };
         s.path = path.to_string_lossy().to_string();
         s
-    }
-
-    pub fn update(&mut self, auto_save: bool, auto_sync: bool, zen_mode: bool) {
-        self.auto_save.store(auto_save, Ordering::Relaxed);
-        self.auto_sync.store(auto_sync, Ordering::Relaxed);
-        self.zen_mode.store(zen_mode, Ordering::Relaxed);
     }
 
     pub fn update_last_open_tabs(&mut self, tabs: &[Tab], active_tab_index: usize) {
@@ -140,7 +128,7 @@ impl Workspace {
         let writeable_dir = Path::new(&writable_dir);
         let writeable_path = writeable_dir.join("ws_conf.json");
         Self {
-            cfg: WsConfig::from_file(writeable_path),
+            cfg: WsConfig::new(writeable_path),
             tabs: vec![],
             active_tab: 0,
             active_tab_changed: false,
@@ -668,8 +656,6 @@ impl Workspace {
                 .map_err(|err| format!("{:?}", err));
             update_tx.send(WsMsg::FileCreated(result)).unwrap();
         });
-
-        // self.cfg.update_last_open_tabs(&self.tabs, self.active_tab);
     }
 
     pub fn open_file(&mut self, id: Uuid, is_new_file: bool, make_active: bool) {
@@ -716,8 +702,6 @@ impl Workspace {
             self.active_tab = n_tabs - 1;
         }
         self.active_tab_changed = true;
-
-        // self.cfg.update_last_open_tabs(&self.tabs, self.active_tab);
     }
 
     pub fn process_updates(&mut self) {
--- a/libs/content/workspace/src/workspace.rs
+++ b/libs/content/workspace/src/workspace.rs
@@ -96,7 +96,7 @@ impl WsConfig {
         self.zen_mode.store(zen_mode, Ordering::Relaxed);
     }
 
-    pub fn update_last_open_tabs(&mut self, tabs: &Vec<Tab>, active_tab_index: usize) {
+    pub fn update_last_open_tabs(&mut self, tabs: &[Tab], active_tab_index: usize) {
         let mut active_tab = None;
         let mut tab_ids: Vec<Uuid> = tabs
             .iter()
--- a/libs/content/workspace/src/workspace.rs
+++ b/libs/content/workspace/src/workspace.rs
@@ -84,6 +84,12 @@ impl WsConfig {
         s
     }
 
+    pub fn update(&mut self, auto_save: bool, auto_sync: bool, zen_mode: bool) {
+        self.auto_save.store(auto_save, Ordering::Relaxed);
+        self.auto_sync.store(auto_sync, Ordering::Relaxed);
+        self.zen_mode.store(zen_mode, Ordering::Relaxed);
+    }
+
     pub fn update_last_open_tabs(&mut self, tabs: &[Tab], active_tab_index: usize) {
         let mut active_tab = None;
         let mut tab_ids: Vec<Uuid> = tabs
--- a/libs/content/workspace/src/workspace.rs
+++ b/libs/content/workspace/src/workspace.rs
@@ -5,9 +5,12 @@ use lb_rs::model::errors::LbErrKind;
 use lb_rs::model::file_metadata::FileType;
 use lb_rs::svg::buffer::Buffer;
 use lb_rs::Uuid;
+use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
+
+use std::{fs, io, mem, thread};
 
 use crate::output::{Response, WsStatus};
 use crate::tab::image_viewer::{is_supported_image_fmt, ImageViewer};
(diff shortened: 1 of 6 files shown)


libs/content/workspace/src/workspace.rs
```diff
@@ -113,18 +121,52 @@ impl WsConfig {
...
+    pub fn new(core: &Lb, ctx: &Context) -> Self {
         let (updates_tx, updates_rx) = channel();
         let background = BackgroundWorker::new(ctx, &updates_tx);
         let background_tx = background.spawn_worker();
         let status = Default::default();
         let output = Default::default();
 
+        let writable_dir = core.get_config().writeable_path;
+        let writeable_dir = Path::new(&writable_dir);
+        let writeable_path = writeable_dir.join("ws_conf.json");
```
Parth, line 167: I think there's a better path forward here. Here are some of my problems with stuff I see:

- skip_serialization is a data modeling smell
- path containing `""` as an initial value is a data modeling smell (that I introduced)
- similar fields getting copied into a few places with different variations seems risky. ws_conf.json is defined here.

So here is what I would suggest:
- move settings into it's own file
- keep the data getting written to it's own file in it's own, clean struct, without any Atomic business, document the fields figure out a good name for this thing
- wrap this thing in what hangs on to  it's `Path`, hangs on to the thing getting written in an `Arc<RwLock<Thing>>`.
- I picture an API that let's you read a one-off setting, ergonomic functions for mutating the settings that kickoff the file update. These functions should take care of the lock for me, and should be named in a way that's easy to discover.
- Writes to the file should happen in a background thread.

Look at all the things we want to do with the idea of [Workspace Settings](https://github.com/lockbook/lockbook/issues/2936) we probably need a version field and that clean struct should probably have a `*V1` with it's name. We'll need to think about migrating this file as we evolve it to be more sophisticated.

libs/content/workspace/src/workspace.rs
```diff
@@ -270,6 +312,12 @@ impl Workspace {
     }
 
     pub fn show(&mut self, ui: &mut egui::Ui) -> Response {
+        if self.ctx.frame_nr() == 0 {
+            self.cfg.last_open_tabs.clone().iter().for_each(|&file_id| {
+                self.open_file(file_id, true, true);
+            });
```
Parth, line 318: you probably don't need to clone the underlying collection here, and `Uuid` impls copy, this is forcing a heap allocation to happen when you can write it slightly differently and have a stack copy instead.

regardless, I see this being a moderately bad experience when you try to open a file that has been deleted as a result of a sync. I'm pretty sure you won't have the experience of being able to say X path was deleted as a result of sync, you won't have the filename or anything. You're just going to see `FileNonexistent`.

 We do want this codepath to be very fast, I think we could consider implementing a fast check for whether an ID exists in core or not and you could check that before opening a file.

Parth (changes requested):

Parth:
I think there's also a universe in which these things are saved in a file in core itself. Or a universe where core & server are more directly aware of various settings.

libs/content/workspace/src/workspace.rs
Parth: You could also consider passing a startup config into open_file which closes itself if FileNonexistent under this narrow circumstance.

Parth:
I don't believe this closes https://github.com/lockbook/lockbook/issues/2604 yet-- there's a file tree component to that request.

libs/content/workspace/src/workspace.rs
tvanderstad: Agree there are some smells including existing smells. This is a good time to do cleanup because we're introducing persistence and because we'll probably start persisting lots of things (so investments here will pay off quickly).

+1 for JSON so we don't have to do the versions and migrations thing (down to debate).

Hesitant to call this 'config' or 'settings' because lots of persistent state (what tabs are open, where they're scrolled to) isn't exactly one of those things. For settings, I have similar API desires - I imagine writing a button to change a single setting and would like that button-writing experience to be smooth. For other persistent things, I like the "just save everything" direction. Possibly worthwhile to split persistent state along those lines and have the API style that makes the most sense for each part.

libs/content/workspace/src/workspace.rs
tvanderstad: I'm skeptical of the use of `frame_nr()`. It shouldn't be breaking to miss the first frame - on the egui client, the first frames may be spent on the sign-in/sign-up flow. The code that initializes workspace seems a better fit

libs/content/workspace/src/workspace.rs
```diff
@@ -382,6 +430,10 @@ impl Workspace {
     fn show_tabs(&mut self, ui: &mut egui::Ui) {
         ui.spacing_mut().item_spacing = egui::vec2(0.0, 0.0);
 
+        if self.active_tab_changed {
+            self.cfg.update_last_open_tabs(&self.tabs, self.active_tab);
+        }
```
tvanderstad, line 435: Only update the persisted tabs when the active tab changes? Does closing an inactive tab correctly update the persisted tabs? I can also see us implementing features that would open a file without setting it active (think cmd+clicking a link in chrome)

tvanderstad: update: closing an inactive tab does work (I'm not sure how)

libs/content/workspace/src/workspace.rs
tvanderstad: update: this whole feature doesn't work at all on the egui client

libs/content/workspace/src/workspace.rs
Parth: +1 for JSON so we don't have to do the versions and migrations thing (down to debate).

good point I'm used to everything persistent not being like this

libs/content/workspace/src/workspace.rs
```diff
@@ -83,49 +84,100 @@ pub struct SaveResult {
...
+        match fs::File::open(&path) {
+            Ok(f) => WsPersistentStore {
+                path: path.to_string_lossy().to_string(),
+                data: Arc::new(RwLock::new(serde_json::from_reader(f).unwrap_or_default())),
+            },
+            Err(_) => WsPersistentStore {
+                path: path.to_string_lossy().to_string(),
+                data: Arc::new(RwLock::new(WsPresistentData::default())),
+            },
+        }
```
Parth, line 116: This is strange to me for a few reasons:
- you have the owned `PathBuf` you can just hang on to that and re-open the file as needed
- you could also just hang on to the file descriptor

Parth: It's also strange to me that if we failed to open / write to this file that we'll silently keep trying. I think at-least a log message is warranted

libs/content/workspace/src/workspace.rs
```diff
@@ -83,49 +84,100 @@ pub struct SaveResult {
...
-        self.auto_sync.store(auto_sync, Ordering::Relaxed);
-        self.zen_mode.store(zen_mode, Ordering::Relaxed);
+    pub fn update<F>(&mut self, updater: F)
+    where
+        F: FnOnce(&mut WsPresistentData),
+    {
+        let mut data_lock = self.data.write().unwrap();
+        let old_data = data_lock.clone();
+
+        updater(&mut data_lock);
```
Parth, line 126: I'm not convinced of the closure here. I think there's a number of simpler paths that won't involve a closure. I can be more explicit if you'd like.

libs/content/workspace/src/workspace.rs
```diff
@@ -142,7 +194,17 @@ impl Workspace {
             focused_parent: None,
             last_touch_event: None,
             out: output,
-        }
+        };
+
+        let last_open_tabs = ws.cfg.data.read().unwrap().last_open_tabs.clone();
```
Parth, line 199: from my last review: 

> I picture an API that let's you read a one-off setting, ergonomic functions for mutating the settings that kickoff the file update. These functions should take care of the lock for me, and should be named in a way that's easy to discover.

libs/content/workspace/src/workspace.rs
```diff
@@ -83,49 +84,100 @@ pub struct SaveResult {
...
+    pub path: String,
+    pub data: Arc<RwLock<WsPresistentData>>,
+}
 
-    pub auto_save: Arc<AtomicBool>,
-    pub auto_sync: Arc<AtomicBool>,
-    pub zen_mode: Arc<AtomicBool>,
+#[derive(Clone, Serialize, Deserialize, Eq, PartialEq)]
+pub struct WsPresistentData {
+    pub last_open_tabs: Vec<Uuid>,
```
Parth, line 94:
```suggestion
    pub open_tabs: Vec<Uuid>,
```

libs/content/workspace/src/workspace.rs
```diff
@@ -83,49 +84,100 @@ pub struct SaveResult {
...
+
+        updater(&mut data_lock);
+
+        // only write if the contents have been changed.
+        if old_data != *data_lock {
+            self.to_file();
+        }
+    }
+
+    pub fn update_last_open_tabs(&mut self, tabs: &[Tab], active_tab_index: usize) {
```
Parth, line 134:
```suggestion
    pub fn set_tabs(&mut self, tabs: &[Tab], current: usize) {
```

Parth: smaller better

libs/content/workspace/src/workspace.rs
Parth: I basically expect this to be `ws.cfg.get_tabs()` handling ownership and concurrency concerns inside the API

Parth (changes requested):
one step further api refinement needed

also could evaluate if `cfg` is the right identifier for this entity

PR #3231, opened 2024-12-20
lb-rs: perf inquiry ch 1

Parth:
# perf inquiry

Logs report `get_pending_shares` being called often and consuming a lot of resources while being calculated.

```
2024-12-20T17:39:11.300169Z DEBUG get_pending_shares: lb_rs::service::share: close time.busy=1.93ms time.idle=1.11µs
```

These are values on an ultra-powerful machine (i9-14), on Smail's M1 I saw values of ~20ms, I'm sure the situation is much worse on an intel Ultrabook. But the ratios of `busy / idle` is mostly the same, suggesting that we're not blocked on IO or anything like that. 

I believe this call is overall representative of most metadata operations, and we want these to be much faster than they are.

Let's see what the CPU spends most of it's time during this:

```rust
    let lb = Lb::init(Config::cli_config()).await.unwrap();
    for _ in 0..1000 {
        lb.get_pending_shares().await.unwrap();
    }
```

pasted_image_2024-12-20_18-12-11

There are 2 skill issues that dominate how we're spending our time:
+ We recompute cryptography for every core call. This accounts for 40% of our CPU time.
+ We pass around ids in owned (and cloned) `HashSet`s, the memory allocations (collect) and O(n) operations associated with this accounts for another 40% of our CPU time. 

# Making Crypto Faster

Let's start with making sure we don't recompute crypto (because it's easier).

`LazyTree` hangs on to some cached values to speed up some algorithms, especially ones that are related to tree traversal. For instance, to calculate whether something is deleted you need to walk up to the parents to make sure they're not deleted.

These calculations are usually pretty cheap, except for the cryptographic ones. Usually these calculations can change due to a number of factors well. For instance if you move a file somewhere you want to invalidate any stored ideas about whether or not this file has been deleted. So the lifetime of the `LazyTree` is always less than the lifetime of a _transaction_ within lb. While a `LazyTree` exists, you know nothing else is going on, because of the complexity of the computation. However, again the cryptograpic operations don't fall into this category.

So let's treat them differently, instead of tying the lifetime of our cryptographic cache to the lifetime of the `LazyTree` (a core call or smaller), let's tie it to the lifetime of `Lb` instead. 

pasted_image_2024-12-21_00-00-50

Great, cryptography doesn't even show up here anymore for `get_pending_shares`. And it's about 30% faster. Let's move onto the trickier optimization -- our memory management. Virtually all the remaining time is spent juggling memory.

# Onto Memory

Our `TreeLike` trait is a fundemental building block of the abstractions within the `lb` library. A "tree" in lockbook can take many shapes, it could be a `LookupTable` (piece of the local db), it could be a `Vec<FileMetadata>` that just came from our server, or it could be a list of hypothetical changes that we're checking for validity.

Very often we're iterating over a set of `Uuid`s (the fundemental identifier of any type of content in our platform: metadata or content, encrypted or decrypted). 

Part of the `TreeLike` contract requires these two implementations:
```rust
fn ids(&self) -> HashSet<&Uuid>;
fn owned_ids(&self) -> HashSet<Uuid>;
```

There's a number of things I don't like about this:
+ the word *owned* here refers to the borrow checker, which is confusing because we also model the idea of ownership of files for the purposes of access control and billing.
+ the only difference here is between whether or not `&Uuid` is borrowed, which `impl`s `Copy` which means that the "non-owned" version very likely performs similarly to the owned version.
+ unless however, `owned_ids` is implemented like this: `self.ids().iter().map(|id| **id).collect()` which it is, which makes `owned_ids` much slower.
+ finally the core purpose of this idea is to provide a base set for a programer to iterate over. Iterating through a `HashSet` is a bad time performance wise.

Here are some options that come to mind:
+ With very little effort we can address all of the above by making `ids(&self) -> Vec<Uuid>`. We lose the built in uniqueness, but in most cases the underlying data structure is a `HashMap`. We do still need some mechanism to enforce uniqueness for `StagedTree` situation. But inherent to the `StagedTree` situation -- changes are generally a smaller set of the overlying situation, so we can do some fast `Vec` scans instead of incurring the cost of populating a hashset.
+ There's probably some world where we can more heavily use `Iterators` (comments indicate that we think this is blocked on some fancy language features but we can probably expose some `&dyn impl Iter<Item = Uuid>`) which could provide genuine speedups for read only operations as we wouldn't need to allocate a container like `Vec`. I think this could be cool to explore in the future because we could make a lot of our data operations composable filters like: `files().documents().not_deleted()` etc. But I have the sense that something like `base.ids().zip(local.ids()).unique()` may perform poorly. 

All this is certainly worth revisiting in the future, but for now I believe the `Vec<Uuid>` gets us 80% of the value for 5% of the effort.

Even better, it looks like `StagedTree` has some constructs for hanging on to intermediate state (for things like `removed` files). I added a `new` file which keeps track of the new ids, so we don't even have to do that scan.

Most of the code changed over from `{ HashSet<&Uuid> | HashSet<Uuid> } -> Vec<Uuid>` 
pretty trivially. Our algorith is a good bit faster, and most of the allocations, collect, and clone statements have dissapeared from our flamegraph. Let's take a look at what remains.
# anything else?

pasted_image_2024-12-23_19-01-55

Looking at this flamegraph I see no obvious room for improvement. Our time is dominated by `linked_by` and `calculate_delete` which both involve some manner of tree-traversal. So as a last step I'll take a moment to audit what the code is doing.

```rust
        for id in tree.ids() {
            // file must not be deleted
            if tree.calculate_deleted(&id)? {
                continue;
            }
            // file must be owned by another user
            if tree.find(&id)?.owner() == owner {
                continue;
            }
            // file must be shared with this user
            if tree.find(&id)?.access_mode(&owner).is_none() {
                continue;
            }
            // file must not have any links pointing to it
            if tree.linked_by(&id)?.is_some() {
                continue;
            }

            let file = tree.decrypt(&self.keychain, &id, &db.pub_key_lookup)?;

            result.push(file);
        }
```

We're trying to figure out which files are pending shares, for almost every situation this is like finding a needle in a haystack. `calculate_deleted` is pretty expensive, and the ownership check right below it is significantly cheaper and would disqualify the vast majority of metadata, could we re-order these checks better?
+ ownership & access control is O(1)
+ calculate_deleted is O(log n)
+ linked_by is O(n)

Just re-ordering these checks cuts the execution time in almost half.

About 50% of the new remaining time is occupied by `linked_by`: 
pasted_image_2024-12-23_19-16-48

# linked by
linked_by is heavily used for remapping ids and for decryption to power the following UXs:
+ you can place a share wherever you want
+ you can name that share whatever you want

we call it twice, once explicitly to filter out any shares that have already been accepted, and then as a part of the final decryption routine for the client.

so we like that, let's checkout it's implementation to see if there's any low hanging fruit.

I think there is some low hanging fruit, I think linked ids are a good candidate for the lazy tree cache. In some sense a better candidate than deleted -- O(n) vs O(log n). The size of the cache is also going to usually be very miniscule. So great value cache item.

This change resulted in about a 3x speedup.

pasted_image_2024-12-23_19-56-32

the remaining time is taken up by:
+ `eq` on pubkey
+ `hash` on uuid

I have a strong desire to call it quits at this point, but I need to find out why `PublicKey`s `eq` has `serialize` in it's stack...

```rust
impl PartialEq for Owner {
    fn eq(&self, other: &Self) -> bool {
        self.0.serialize() == other.0.serialize()
    }
}
```

maybe we could not and just `#[derive(PartialEq)]`. Skipping this results in another roughly 2-3x speedup.

pasted_image_2024-12-23_20-51-21

At this point we're spending most of our time hashing UUIDs. That's probably a good place to leave it for now.

# Summary
Let's step through these changes and see what our overall time difference was. I'm going to bump the number of `pending_shares` calls up to 5k to make it easy to differentiate against the initial app startup: 
* baseline startup + 2500its: **5s**
* don't repeat crypto: **4s**
* owned_ids: **3.3s**
* linked_by: **0.25s**
* don't ser for eq: **0.16s**

Almost everything should yield signficant speeds to almost every core and server call, though there is still much more cruft to root out in this manner.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,8 +22,7 @@ members = [
 ]
 
 # restore for big binaries but good backtraces
- [profile.release]
- debug = true
-
+#  [profile.release]
+#  debug = true
 # [profile.bench]
 # debug = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ members = [
 ]
 
 # restore for big binaries but good backtraces
-#  [profile.release]
-#  debug = true
+[profile.release]
+debug = true
 # [profile.bench]
 # debug = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,5 +22,8 @@ members = [
 ]
 
 # restore for big binaries but good backtraces
-# [profile.release]
+ [profile.release]
+ debug = true
+
+# [profile.bench]
 # debug = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ members = [
 ]
 
 # restore for big binaries but good backtraces
-[profile.release]
-debug = true
+# [profile.release]
+# debug = true
 # [profile.bench]
 # debug = true
--- a/libs/lb/lb-rs/src/service/keychain.rs
+++ b/libs/lb/lb-rs/src/service/keychain.rs
@@ -45,33 +45,7 @@ impl From<Option<&Account>> for Keychain {
 
 impl Lb {
     pub fn get_account(&self) -> LbResult<&Account> {
-        self.keychain
-            .account
-            .get()
-            .ok_or_else(|| LbErrKind::AccountNonexistent.into())
-    }
-
-    pub fn get_pk(&self) -> LbResult<PublicKey> {
-        self.keychain
-            .public_key
-            .get()
-            .copied()
-            .ok_or_else(|| LbErrKind::AccountNonexistent.into())
-    }
-
-    #[doc(hidden)]
-    pub async fn cache_account(&self, account: Account) -> LbResult<()> {
-        let pk = account.public_key();
-        self.keychain
-            .account
-            .set(account)
-            .map_err(|_| LbErrKind::AccountExists)?;
-        self.keychain
-            .public_key
-            .set(pk)
-            .map_err(|_| LbErrKind::AccountExists)?;
-
-        Ok(())
+        self.keychain.get_account()
     }
 }
 
--- a/libs/lb/lb-rs/src/service/keychain.rs
+++ b/libs/lb/lb-rs/src/service/keychain.rs
@@ -1,15 +1,27 @@
+use std::{
+    collections::HashMap,
+    sync::{Arc, RwLock},
+};
+
 use crate::{
-    model::account::Account,
-    model::errors::{LbErrKind, LbResult},
+    logic::crypto::AESKey,
+    model::{
+        account::Account,
+        errors::{LbErrKind, LbResult},
+    },
     Lb,
 };
 use libsecp256k1::PublicKey;
 use tokio::sync::OnceCell;
+use uuid::Uuid;
+
+pub type KeyCache = Arc<RwLock<HashMap<Uuid, AESKey>>>;
 
 #[derive(Default, Clone)]
 pub struct Keychain {
     account: OnceCell<Account>,
     public_key: OnceCell<PublicKey>,
+    key_cache: KeyCache,
 }
 
 impl From<Option<&Account>> for Keychain {
@@ -18,8 +30,9 @@ impl From<Option<&Account>> for Keychain {
             Some(account) => {
                 let account = account.clone();
                 let pk = account.public_key();
+                let key_cache = Default::default();
 
-                Self { account: OnceCell::from(account), public_key: OnceCell::from(pk) }
+                Self { account: OnceCell::from(account), public_key: OnceCell::from(pk), key_cache }
             }
             None => Self::default(),
         }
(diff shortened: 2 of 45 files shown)


Cargo.toml
```diff
@@ -22,5 +22,8 @@ members = [
 ]
 
 # restore for big binaries but good backtraces
-# [profile.release]
+ [profile.release]
+ debug = true
+
```
Parth, line 27:
```suggestion
#  [profile.release]
#  debug = true
```

libs/lb/lb-rs/src/service/keychain.rs
```diff
@@ -57,3 +70,44 @@ impl Lb {
...
+
+    pub fn insert_aes_key(&self, id: Uuid, key: AESKey) -> LbResult<()> {
+        self.key_cache.write()?.insert(id, key);
+        Ok(())
+    }
+
+    pub fn get_aes_key(&self, id: &Uuid) -> LbResult<Option<AESKey>> {
+        Ok(self.key_cache.read()?.get(id).copied())
+    }
+}
```
Parth, line 113: evaluate code duplication here

libs/lb/lb-rs/tests/cli_data_tests.rs
```diff
@@ -0,0 +1,10 @@
+use lb_rs::{model::core_config::Config, Lb};
+
+#[tokio::test]
+// #[ignore]
```
Parth, line 4:
```suggestion
#[ignore]
```

libs/lb/lb-rs/src/service/sync.rs
```diff
@@ -187,11 +187,12 @@ impl Lb {
         }
 
         let mut base_staged = (&mut db.base_metadata).to_lazy().stage(None);
-        base_staged.tree.removed = prunable_ids.clone();
+        // todo this deserves more scrutiny before this is merged
```
Parth, line 190: scrutinize

Parth:
It was also seen at points that people had 100ms+ values in their logs. The idle time here was low. I suspect that the tokio mutex doesn't show up properly in logs.

It would be easy to verify and I don't think our use of the tokio mutex is essential, but I'm also skeptical this is the case because the same people make both. so further investigation is certainly required.

Parth:
Those several hundred millisecond log lines were misread by me -- they were list metas, which were genuinely taking that long, and are now generally less than 1ms (500+ x speedup).

PR #3242, opened 2024-12-27
workspace: task management refactor

tvanderstad:
This PR takes an escalated approach to our workspace data loss problems
- replaces workspace channels with a 'task manager' that's smarter and more transparent
- synchronously invokes operations that aren't file I/O or sync, including metadata and sync status
- replaces background work mechanism with simpler one based on `request_repaint_after()`
- moves draw-related code, unchanged, to a new file show.rs
- adds a status icon to the tab strip with a tooltip that shows when the tab was last saved (updates in real time)

<details>

https://github.com/user-attachments/assets/d9622334-d95e-4972-a554-e82f49fc4e1e

</details>


clients/egui/src/account/mod.rs
```diff
@@ -99,10 +98,7 @@ impl AccountScreen {
     pub fn begin_shutdown(&mut self) {
         self.shutdown = Some(AccountShutdownProgress::default());
         self.workspace.save_all_tabs();
-        self.workspace
-            .background_tx
-            .send(BwIncomingMsg::Shutdown)
```
tvanderstad, line 104: the handling of this message was ultimately a `todo` anyway

libs/content/workspace/src/task_manager.rs
```diff
@@ -0,0 +1,462 @@
...
+        } = cache.update();
+
+        for queued_load in loads_to_launch {
+            let request = queued_load.request.clone();
+            let in_progress_load = InProgressLoad::new(queued_load);
+            cache.in_progress_loads.push(in_progress_load);
+
+            let self_clone = self.clone();
+            let core = core.clone();
+            let ctx = ctx.clone();
```
Parth, line 353: would be reasonable for `TaskManager` to hang on to these for convenience

libs/content/workspace/src/task_manager.rs
```diff
@@ -0,0 +1,462 @@
...
+    pub in_progress_sync: Option<InProgressSync>,
+
+    // completions stored here then returned in the response on the next frame
+    completed_loads: Vec<CompletedLoad>,
+    completed_saves: Vec<CompletedSave>,
+    completed_sync: Option<CompletedSync>,
+}
+
+struct InnerResponse {
+    loads_to_launch: Vec<QueuedLoad>,
```
Parth, line 32: `loads_to_launch` 😏 

it's not clear to me why loading can't begin immediately making progress. Having to wait for a frame to do this (and has no implication for the UI) seems like a code smell to me.

In my head I'm thinking about something that looks mostly like this but operates in a more channel based manner. And you don't need the UI thread as this sort of touch point for progress to happen / state communication to be facilitated.

Would be mostly the same API (with many of the same wins of centrally managing this complexity here) but UI would be able to receive at it's convenience, and threads would make progress optimally.

Parth (changes requested):
even if crude, I would really like to see some visibility in the UI for what's taking place. Even if it's just a `*` character by the name of the tab.

it's obvious to me that this PR has a lot of value and improves the present situation in lots of ways, but it is still large, and touches some critical portions.

I think you've done a great job making the code as clear and simple as possible (apart from some relatively minor changes req'd). Making the UI represent the situation would cause significant anxiety reduction for me.

libs/content/workspace/src/task_manager.rs
```diff
@@ -0,0 +1,462 @@
...
+
+#[derive(Clone)]
+pub struct SaveRequest {
+    pub id: Uuid,
+    pub old_hmac: Option<DocumentHmac>,
+    pub seq: usize,
+    pub content: String,
+}
+
+// Telemetry
```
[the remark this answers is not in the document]
tvanderstad: Down to use whatever word, having timing info felt like a reasonable default. I'll add logging if an operation takes too long and differentiate between it being queued for too long vs in progress for too long.

libs/content/workspace/src/task_manager.rs
```diff
@@ -0,0 +1,462 @@
...
+pub trait FileCacheExt {
+    // pass-throughs
+    fn new() -> Self;
+    fn queue_load(&mut self, request: LoadRequest);
+    fn queue_save(&mut self, request: SaveRequest);
+    fn queue_sync(&mut self);
+
+    // non-pass-throughs
+    fn update(&mut self, ctx: &Context, core: &Lb) -> Response;
+}
```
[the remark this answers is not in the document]
tvanderstad: This way makes it less noisy to use the TaskManager because you don't have to reference the inner field to lock the mutex, but no strong feelings, I'll just change it if you prefer it that way

libs/content/workspace/src/task_manager.rs
tvanderstad: Okay I'll have another pass with attention to the involvement in the UI thread in making progress

Parth (approved):
This looks good to me. Some non blocking thoughts:
- I haven’t investigated too deeply but I have a vague recollection that interacting with time if we get no value out of is somewhat expensive. But if we even get slight value it’s worth it.
- If we’re just doing this to get timing info into logs (probably worthwhile, though watch out for log volume problems) `#[instrument]` is probably a better path forward.

Issue #1918, opened 2023-07-13
core shouldn't retrieve all files


Parth:
during a sync we pull any files indiscriminately, files we may not want to pull initially include:
+ very big files on phones
+ files for shares that haven't been accepted

Issue #3194, opened 2024-12-10
search results prioritization


smailbarkouch:
Commonly opened files should appear higher in the search results. This hints that document activity is incorrectly being factored into score rankings.

Parth:
the old code used to:
- get metrics
- normalize
- min(metrics.score, 600)

and the new code does absolutely none of that

tvanderstad:
I don't think it's just a problem with the incorporation of document activity. Check out these search results:

<img width="547" alt="Image" src="https://github.com/user-attachments/assets/1aeb27ba-5aa3-447b-8aee-ff48575c1f87" />

In this case I'm looking to open my dev journal (second entry). Here are some reasons it could/should be prioritized (in order of importance):
1. the file name starts with the search term rather than just containing it
2. the file name is shorter so more of it is comprised of the search term
3. the file has more activity
4. the file was more recently last modified
5. the file is my own
6. the file is editable

tvanderstad:
Last time I implemented a search, I had a few back-and-forths with users and eventually landed on this as the prioritization:
1. exact matches (e.g. search "2024-12.md" to pull up file "2024-12.md")
2. prefix matches (e.g. search "2024-12" to pull up file "2024-12.md")
3. substring matches (e.g. search "24-12" to pull up file "2024-12.md")
4. subsequence matches (e.g. search "2024md" to pull up file "2024-12.md")
5. fuzzy matches (e.g. search "2025md" to pull up file "2024-12.md")

I was not able to lean on my search library for this (I was implementing using trigram matching in Postgres) and I had to implement this prioritization manually. I thought we'd expect higher quality output from our search library here but it's looking like implementing this prioritization manually would make a noticeable improvement.

Parth:
This is useful product level info, I'm down to just do this and possibly something like #2718 falls into a type 1 result.

@tvanderstad how do you think #3255 interacts with this prioritization matrix?

Naively maybe it's below a fuzzy match, but that feels wrong / underestimating what an LLM could do for us. I suppose the question is: How do we detect that a user is trying to engage in a traditional search experience vs. as an LLM oriented style question.

Parth:
@tvanderstad would this make it to your top 3? I wonder if we didn't really complete #2544 to your satisfaction

Parth:
@ad-tra same question for you

PR #3250, opened 2024-12-29
lb: efficient search index rebuilds

Parth:
this pr makes search index rebuilds more efficient by only updating paths that changed if metas changed, and only re-reading the specific documents that have changed (either as a result of a sync or a direct write).

# context 
as `lb` evolves the number of parties (internally & externally) interested in keeping decrypted *views* of documents or metadata up to date is growing.

this pr allows such interested parties to *subscribe* to be notified when key events occur so they can take actions like:
- requerying `lb`
- updating internal data structures
- redrawing literal UI views that depend on those data structures

this api is coming now instead of later to support the following bodies of work:
- https://github.com/lockbook/lockbook/issues/1918: core will have a background service that will detach document syncing from the overall sync process. this will make a number of things feel significantly faster as well as reduce the current ux "penalty" for investing deeper into lockbook (and uploading images for instance).
- an upcoming status api which moves the core status computations into core (this api will be used to tell clients how to keep those parts of the app up to date).
- eliminating a number of situations in which clients are currently polling (see #3231, here clients are polling `get_pending_shares` every second to populate an icon).
- search index rebuilds are currently very expensive, and we perform full index rebuilds everytime a single document is updated. This subscription api would let search (an internal consumer of events) keep a view of decrypted documents up to date at all times. 
- similarly writes and reads don't have to wait for suggested docs, suggested docs can just subscribe to this channel and keep itself up to date

@tvanderstad has expressed the desire to have some rust lib (core, ws, or something else) manage the process of keeping a `Vec<File>` up to date at all times. I hope this api will have one of two outcomes:
- the implementation surface area of such an api is reduce to the point where it's not necessary 
- or, if still necessary, that an optimal implementation is trivial


Parth:
As I've written more of this PR I am sold on the idea @tvanderstad has been talking about. I think the following are good candidates for *this thing* that he's talking about:
- `Vec<File>`
- cached path computations
- status
- pending_shares
- usage calculations
- search index

for use by:
- ws
- egui
- cli
- lb-fs

Basically anything where the choice was: "do I impl this managing a tree myself, or can I just build my functionality on top of some other more fundamental core call?"

I'll continue to add to this list over time.

Ideally the contract would be something like `Arc<RwLock<View>>` or something like that, so for rust clients, even memory allocation can be minimized to what is absolutely needed.

libs/lb/lb-rs/src/service/search.rs
```diff
@@ -111,6 +110,15 @@ impl Lb {
             }
         }
 
+        // if the index is empty wait patiently for it become available
+        loop {
+            if self.search.index.read().await.is_empty() {
```
Parth, line 115: some sort of timeout here would be wise

libs/lb/lb-rs/src/service/events.rs
```diff
@@ -0,0 +1,60 @@
...
+    tx: Sender<Event>,
+}
+
+#[derive(Clone, Copy, Debug)]
+pub enum Event {
+    /// If any document or folder is birthed by this lb library
+    /// new files as a result of sync are communicated as MetadataChanged
+    /// documents that have new contents are communicated as DocumentWritten
+    NewFile(Uuid),
+
```
Parth, line 18: maybe just drop this

Parth:
should ideally sort this out here: https://github.com/lockbook/lockbook/issues/3194

PR #3272, opened 2025-01-04
marketing: new website

ad-tra:


libs/lb/lb-rs/Cargo.toml
```diff
@@ -24,32 +24,34 @@ bezier-rs = "0.2.0"
...
 tracing-appender = "0.2"
 sublime_fuzzy = "0.7.0"
 crossbeam = "0.8.1"
 qrcode-generator = "4.1.6"
-db-rs = "0.3.1"
-db-rs-derive = "0.3.1"
 strip-ansi-escapes = "0.2.0"
 chrono = "0.4"
 rand = "0.8.4"
+getrandom = { version = "0.2", features = ["js"] }
```
Parth, line 44: does this remove `rand`? `rand` does support wasm iirc. Rand has cryptography implications iirc. Will continue to review.

libs/lb/lb-rs/src/blocking.rs
```diff
@@ -30,10 +34,34 @@ use crate::{
...
+        block_on(future)
+    }
+
+    #[cfg(not(target_family = "wasm"))]
+    fn block_on<F>(&self, future: F) -> F::Output
+    where
+        F: Future,
+    {
+        self.rt.block_on(future)
+    }
```
Parth, line 62: I believe we could *just* use `futures::executor` for our blocking clients as well, but I'm not sure. I don't think we have anything specifically tied to `tokio` that couldn't be moved over. Would be nice to just have one of these. Makes it riskier tho. Any thoughts?

libs/lb/lb-rs/src/blocking.rs
```diff
@@ -30,10 +34,34 @@ use crate::{
...
 pub struct Lb {
     lb: crate::Lb,
+    #[cfg(not(target_family = "wasm"))]
     rt: Arc<Runtime>,
 }
 
 impl Lb {
+    #[cfg(target_family = "wasm")]
+    pub fn init(config: Config) -> LbResult<Self> {
+        let lb = crate::Lb::init_dummy(config)?;
```
Parth, line 44: I think this would ideally describe in what way it's a `dummy` `lb`.

libs/lb/lb-rs/src/lib.rs
```diff
@@ -35,13 +35,40 @@ pub struct Lb {
...
+            keychain: Default::default(),
+            db: Arc::new(RwLock::new(db)),
+            docs: AsyncDocs::from(&config),
+            search: Default::default(),
+            client: Default::default(),
+            syncing: Default::default(),
+            events: Default::default(),
+        })
+    }
+}
```
Parth, line 63: I don't think you need a dummy version of this function, just a dummy version of this config.

libs/lb/lb-rs/src/io/docs.rs
```diff
@@ -22,6 +28,34 @@ pub struct AsyncDocs {
     location: PathBuf,
 }
 
+#[cfg(target_family = "wasm")]
```
Parth, line 31: It should end up documented somewhere that lb-rs doesn't support wasm beyond a "demo" capacity. Not sure where, maybe in the config you define, see below comment.

libs/lb/lb-rs/src/lib.rs
```diff
@@ -35,13 +35,40 @@ pub struct Lb {
...
+
 impl Lb {
     #[instrument(level = "info", skip_all, err(Debug))]
     pub async fn init(config: Config) -> LbResult<Self> {
         logging::init(&config)?;
 
         let db = CoreDb::init(db_rs::Config::in_folder(&config.writeable_path))
-            .map_err(|err| LbErrKind::Unexpected(format!("{:#?}", err)))?;
+            .map_err(|err| LbErrKind::Unexpected(format!("db rs creation failed: {:#?}", err)))?;
```
Parth, line 71: when you rebase you'll have access to the `.unexpected()` implementation from the extension trait in `errors.rs`, I suggest changing this to that.

libs/lb/lb-rs/src/service/debug.rs
```diff
@@ -1,16 +1,25 @@
...
+#[cfg(not(target_family = "wasm"))]
+use {
+    std::env,
+    std::io::SeekFrom,
+    std::path::{Path, PathBuf},
+    tokio::{
+        fs::{self, File},
+        io::{AsyncReadExt, AsyncSeekExt},
+    },
+};
```
Parth, line 22: applies to a lot of files, but I imagine this is going to be annoying to maintain without CI or something.

libs/lb/lb-rs/src/service/debug.rs
```diff
@@ -27,6 +36,14 @@ pub struct DebugInfo {
...
 }
 
+#[cfg(target_family = "wasm")]
+impl Lb {
+    pub async fn debug_info(&self, _os_info: String) -> LbResult<String> {
+        Ok("NO LOGS FOUND".to_string())
+    }
+}
+
+#[cfg(not(target_family = "wasm"))]
```
Parth, line 46: I think we could provide a better debug info here, at the very least a more descriptive debug info "no debug info for wasm" wold be a better string. But I wonder if something like workspace should just have a wasm mode without the concept of Lb. Or if you should make an issue to support wasm in a more real way. Maybe a high quality github issue documenting current wasm implementation corners cut would be better.

public-site/install.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
Parth, line 1: maybe call this `setup_macos.sh`

public-site/static/android-screenshot.png
Parth, line 1: can we git-fls this?

public-site/static/banner-bg.svg
Parth, line 1: I like some prior iteration of this tesselation better, I think one with just key holes or something, this one is a lil busy.

public-site/static/favicon/favicon.svg
Parth, line 1: this favicon a lil hard to see, can we get keyhole here?

public-site/static/linux-screenshot.png
Parth, line 1: same for git-lfs for these assets, maybe all pngs in this folder. Or you can pursue more aggressive compression. The overall repo should be fast to clone. There's some historic cleanup we have to do anyways. Cloning the repo is something a lot of scripts / CI / installers do.

public-site/trunk/resources/canvas-demo.svg
Parth, line 1: 0 bytes?

public-site/trunk/src/app.rs
```diff
@@ -0,0 +1,122 @@
...
+        // fonts
+        //     .font_data
+        //     .insert("pt_mono".to_string(), FontData::from_static(martian));
+        // fonts
+        //     .font_data
+        //     .insert("pt_bold".to_string(), FontData::from_static(martian_bold));
+
+        // fonts
+        //     .families
+        //     .insert(FontFamily::Name(Arc::from("Bold")), vec!["pt_bold".to_string()]);
```
Parth, line 51: intentional?

public-site/trunk/src/app.rs
```diff
@@ -0,0 +1,122 @@
...
...013183594 L 66.89703369140625 397.31663513183594 L 69.01812744140625 394.76585388183594 L 71.73297119140625 392.04710388183594 L 74.44781494140625 389.32835388183594 L 78.60406494140625 383.26976013183594 L 80.38922119140625 380.05491638183594 L 81.57672119140625 377.67210388183594 L 83.02203369140625 374.82444763183594 L 83.95953369140625 372.94163513183594 L 83.95953369140625 376.40647888183594 L 83.10797119140625 379.37913513183594 L 82.20172119140625 383.00022888183594 L 81.74859619140625 386.62132263183594 L 81.29547119140625 390.24241638183594 L 80.49859619140625 395.05882263183594 L 80.49859619140625 398.03147888183594 L 80.49859619140625 400.41038513183594 L 80.49859619140625 402.78929138183594 L 80.49859619140625 405.16819763183594 L 83.36578369140625 406.55882263183594 L 86.33843994140625 406.98069763183594 L 89.31109619140625 406.98069763183594 L 92.28375244140625 406.55491638183594 L 95.45172119140625 404.74241638183594 L 98.81500244140625 401.85569763183594" stroke-width='5' stroke='rgba(0,0,0,1)' fill='none' id='3a59007b-5ecb-4118-9b2d-65722148c1eb' transform='matrix(1 0 0 1 0 0)'/><g id="lb_master_transform" transform="matrix(1 0 0 1 602.1928 -12.461166)"></g> </svg>
```
Parth, line 91: include_bytes / include_string?

public_site/CNAME
```diff
@@ -1 +0,0 @@
-lockbook.net
```
Parth, line 1: just make sure gcs doesn't want this, don't think it does

Parth (changes requested):
- I think a mild readme in pub site would go a long way describing the overall build, tools used, etc.
- I would prefer to move the public_site into the docs folder.
- how much effort would light mode / dark mode take? can happen after merge.
- we should think about what deployment will take, can happen after merge.

libs/content/workspace/src/tab/markdown_editor/images.rs
```diff
@@ -148,11 +149,11 @@ pub fn calc(
     result
 }
 
-fn download_image(
-    client: &reqwest::blocking::Client, url: &str,
-) -> Result<Vec<u8>, reqwest::Error> {
-    let response = client.get(url).send()?.bytes()?.to_vec();
-    Ok(response)
+fn download_image(client: &reqwest::Client, url: &str) -> Result<Vec<u8>, reqwest::Error> {
```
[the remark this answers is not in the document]
ad-tra: yeah it broke the editor. but i fixed it and also fixed on wasm so we can include images in the demo

libs/lb/lb-rs/src/blocking.rs
ad-tra: yeah maybe we can unify our approach at another stage. this pr is already too big

Parth:
Some ideas on roadmap, happy to discuss in this PR:

# Completed
+ End to End encryption
+ Collaboration
+ Offline Support
+ Markdown and SVG editing support
+ PDF and Image Preview
+ CLI support

# In Development
+ Stability and Scaling improvements
+ Large File Support
+ PDF editing
+ File System Integration

# Coming Soon
+ Document History
+ Transcription / Audio support
+ Mentions, Comments and Notifications
+ Browser Extension for "Clippings"
+ LLM Integration

libs/lb/lb-rs/src/blocking.rs
ad-tra: futures::executor compiles for wasm but i recently found out that it panics at runtime, i haven't found a way to implement blocking behavior in wasm yet

libs/lb/lb-rs/src/lib.rs
ad-tra: i need a non async init function

libs/lb/lb-rs/src/service/debug.rs
ad-tra: yeah it would be nice to add wasm as a compilation target in ci to ensure that it doesn't brake.

Parth:
Would like to see an "author" field for all the blog posts

.gitattributes
```diff
@@ -2,3 +2,6 @@ graphics/** filter=lfs diff=lfs merge=lfs -text
 docs/design-ui/**/*.png filter=lfs diff=lfs merge=lfs -text
 docs/design-ui/**/*.jpg filter=lfs diff=lfs merge=lfs -text
 docs/design-ui/**/*.jpeg filter=lfs diff=lfs merge=lfs -text
+*.jpg filter=lfs diff=lfs merge=lfs -text
+*.png filter=lfs diff=lfs merge=lfs -text
+*.jpeg filter=lfs diff=lfs merge=lfs -text
```
Parth, line 7: I think these are ideally constrained to the public_site directory I don't want to break workspace logo or something like that

libs/lb/lb-rs/src/service/debug.rs
Parth: maybe just create an issue to track follow up work like this

server/Cargo.toml
```diff
@@ -47,7 +47,7 @@ itertools = "0.10.1"
 reqwest = { version = "0.11", default-features = false, features = ["json", "rustls-tls"] }
 jsonwebtoken = "8.2.0"
 x509-parser = { version = "0.15.0", features = ["verify", "validate"]}
-db-rs = "0.3.1"
+db-rs = {git="https://github.com/lockbook/db-rs"}
```
Parth, line 50: 0.3.2 was published you can probably refer to this normally now

public-site/static/favicon/favicon.svg
Parth: Do we want to try an un-skewed keyhole?

Parth:
@ad-tra can you go through and resolve what you feel like you've completed? I checked a few things and found mixed results

libs/lb/lb-rs/Cargo.toml
```diff
@@ -24,32 +24,34 @@ bezier-rs = "0.2.0"
 bincode = "1.3.3"
 time = "0.3.20"
 diffy = "0.3.0"
+db-rs = {git="https://github.com/lockbook/db-rs"}
```
tvanderstad, line 27: is this intended to stay this way? also, nit: formatting

libs/lb/lb-rs/src/blocking.rs
```diff
@@ -33,10 +37,34 @@ use crate::{
...
+        let lb = crate::Lb::init_dummy(config).unwrap();
+        Ok(Self { lb })
+    }
+
+    #[cfg(target_family = "wasm")]
+    fn block_on<F>(&self, future: F) -> F::Output
+    where
+        F: Future,
+    {
+        block_on(future)
```
tvanderstad, line 56: nit: I prefer to qualify instead of import a raw fn like this. There are a lot of `block_on`'s happening around here
```suggestion
        futures::executor::block_on(future)
```

libs/lb/lb-rs/src/io/mod.rs
```diff
@@ -66,31 +67,31 @@ impl<'a> LbTx<'a> {
 
 impl Lb {
     pub async fn ro_tx(&self) -> LbRO<'_> {
-        let start = std::time::Instant::now();
+        let start = Instant::now();
 
-        // let guard = tokio::time::timeout(std::time::Duration::from_secs(1), self.db.read())
+        // let guard = web_timetimeout(std::time::Duration::from_secs(1), self.db.read())
```
tvanderstad, line 72: nit: probably-unintended comment change

libs/content/workspace/src/tab/markdown_editor/images.rs
```diff
@@ -130,9 +68,19 @@ pub fn calc(
                             *image_state.lock().unwrap() = ImageState::Failed(err);
                         }
                     }
+                });
 
-                    // request a frame when the image is done loading
-                    ctx.request_repaint();
+                #[cfg(not(target_arch = "wasm32"))]
+                thread::spawn(move || {
```
tvanderstad, line 74: Could we have used something from `futures`, the way you managed the blocking behavior in lb-rs, to make the functions in this file async whether they're wasm or not? Then we wouldn't need two versions and I imagine this diff would be much cleaner

tvanderstad (commented):
Overall, this feels like it creates a small mess in order to deliver the feature quickly. I'm on board if we can align on a target state and a timeline to achieve it, and I'm down to do my part.

Here's a pitch to get us started:
- Let's aim to drop WASM support for `lb-rs` and workspace, where the undue portion of the mess is concentrated imo, and instead ship just the editor and canvas modules (with simple wrappers if necessary).
- I can start by factoring dependencies on `lb-rs` and `workspace` out of the editor. I'll just have it accept an optional `FilePathResolver` of some kind so we can resolve relative links in our apps and not depend on `lb-rs` in WASM (a feature I want to add anyway). I'll move any workspace helper functions to wherever I need to to make it happen. I'll aim to finish by the end of this quarter (2025 Q2).
- You can apply the same pattern to canvas like we did with collaborative editing, then revert the `lb-rs` and `workspace` changes that were needed to support those in WASM, aiming to finish by end of Q3.

These timelines are approximate. Basically I see some of the changes as tech debt - the conditional compilation forces me to basically maintain two programs simultaneously which has been a bad time in the past - but we only pay interest on that tech debt when we're maintaining or building on that code. For example, if we want to up our persistence game and persist scroll positions in open files, we'll be reminded that PDFs are only present in the workspace content types enum for some targets. So here I'm saying Q2 or Q3 as a guess for "when is the next significant change to something affected by the debt" or possibly "when is the next engineer to onboard to the code affected by the debt".

libs/lb/lb-rs/Cargo.toml
Parth: not intended, the change has been released

Parth:
Alternatively -- we could explicitly support wasm. I think there are situations where a web app makes sense and it would be annoying to factor all those things out of the editor and canvas for effectively no reason long term.

I think this forces us to do things the wasm way for the wasm situation, but my ideal investment would be to understand if these are just fundamentally the right ways to do these things anyways? Is it more correct for something like lb-rs (which has no main method) to just depend on a general `futures` implementation rather than forcing a tokio one, for instance.

But broadly I'm aligned on mess reduction happening in this quarter if this is merged.

I think this would still be a reasonable state for the standalone editor, lb-rs would go to resolve a link and find that you're not logged in and would not resolve the link if it's lb://.

Parth:
And I'm down to make the lb-rs support of wasm my problem.

Parth:
we chatted after standup and basically I think both things are going to happen
- for standalone editor he'll abstract out lb, because 
- lb will also just support wasm
- canvas can go either direction


but either way you should rebase and merge this

Parth:
@tvanderstad I'm assuming you're certifying the editor changes

Parth:
@ad-tra I would ideally merge this after wed's release

tvanderstad (approved):

Parth:
nix-release depends on `public-site/favicon.svg` existing, I can update this before you merge but we must correct it before a release.

Issue #3271, opened 2025-01-04
workspace produces self file write conflicts


tvanderstad:
Observed with markdown and canvas, iPadOS and macOS. When observed in markdown this erroneously inserted a bunch of words. When observed in a ~30MB canvas it froze the UI thread for a few seconds.

Issue #3270, opened 2025-01-04
workspace saves and syncs excessively


tvanderstad:
Observed when panning around a large canvas - sync status seems to get stuck syncing forever and/or saves and syncs alternate for a while. Maybe workspace is not managing tasks correctly or maybe canvas is requesting saves too often.

Issue #3108, opened 2024-11-12
logs in more situations


Parth:
+ when a test fails print out the logs
+ show the logs in xcode
+ support logcat

Parth:
@smailbarkouch & I paired on logcat support
@tvanderstad & I pared on xcode support

so I think we're good

PR #3275, opened 2025-01-07
workspace: task management refinements

tvanderstad:
**changes**
- tab content is now copied for save when the save is launched (rather than when it's queued) and canvas contents are now serialized on a background thread
- tasks once again rely on periodic check-ins from the UI thread via workspace (required to prevent file self-conflicts)
- introduces a tab `is_closing` ("purgatory") state which collapses the tab's tab to just a status indicator and removes it when all dependent tasks are complete (i.e. if a save was in progress, finish saving the tab before closing)
  - closing a tab now generally just moves it to this state
  - we check each frame if it's ready to close
  - some interactions (like opening the file again) will remove the state leaving the tab as a normal tab
- clean ups including removal of unpopulated/unused fields `workspace.out.error` and `workspace.status.usage`
- logging improvements
  - logs from `lbrs` and `dbrs` are now visible in stdout
  - workspace `info!`-logs task updates like queues, launches, and completions for saves, loads, and syncs
  - workspace `warn!`-logs when tasks take a long time (>1s for async tasks, >100ms for work on UI thread)
  - workspace `error!`-logs when tasks fail
  - lbrs `warn!`-logs when locks are held for a long time (>100ms)
  - lbrs `warn!`-logs when network requests are large (>1MB) or take a long time (>1s)
- fixes https://github.com/lockbook/lockbook/issues/3270
- fixes https://github.com/lockbook/lockbook/issues/3271

**notes for reviewers**
- review the interesting bits I called out in my self-review
- QA a big svg; try closing it while it's still saving
- logging
  - is my usage of tracing good? should I be using more spans (difficult when tasks cross threads or otherwise aren't single fns)? should I be providing format arguments differently?
  - any opportunity for helper fn's here? unsure how to create a helper for logging warnings when tasks take a long time without requiring closures (prob not worth)

**todo**
- [x] fix autosave is_dirty mechanism so it doesn't queue a save if the previous save hasn't finished (with the same content)
- [x] fix sync status refresh sometimes taking >2s on the UI thread

### DIFF
--- a/libs/content/workspace/src/show.rs
+++ b/libs/content/workspace/src/show.rs
@@ -417,7 +417,7 @@ impl Workspace {
             Icon::SCHEDULE.size(icon_size)
         } else if self.tasks.load_or_save_in_progress(tab.id) {
             Icon::SAVE.size(icon_size)
-        } else if tab.is_dirty() {
+        } else if tab.is_dirty(&self.tasks) {
             Icon::CIRCLE.size(icon_size)
         } else {
             Icon::CHECK_CIRCLE.size(icon_size)
@@ -588,7 +588,7 @@ impl Workspace {
                     "save queued"
                 } else if self.tasks.load_or_save_in_progress(tab.id) {
                     "save in progress"
-                } else if tab.is_dirty() {
+                } else if tab.is_dirty(&self.tasks) {
                     "unsaved changes"
                 } else {
                     "all changes saved"
--- a/libs/content/workspace/src/show.rs
+++ b/libs/content/workspace/src/show.rs
@@ -273,6 +273,9 @@ impl Workspace {
                                 }
                                 TabLabelResponse::Closed => {
                                     self.close_tab(i);
+                                }
+                                TabLabelResponse::Removed => {
+                                    self.remove_tab(i);
 
                                     let title = match self.current_tab() {
                                         Some(tab) => tab.name.clone(),
@@ -405,23 +408,23 @@ impl Workspace {
     fn tab_label(
         &mut self, ui: &mut egui::Ui, t: usize, is_active: bool, active_tab_changed: bool,
     ) -> Option<TabLabelResponse> {
-        let t = &mut self.tabs[t];
+        let tab = &mut self.tabs[t];
         let mut result = None;
 
         let icon_size = 16.0;
         let x_icon = Icon::CLOSE.size(icon_size);
-        let status_icon = if self.tasks.load_or_save_queued(t.id) {
+        let status_icon = if self.tasks.load_or_save_queued(tab.id) {
             Icon::SCHEDULE.size(icon_size)
-        } else if self.tasks.load_or_save_in_progress(t.id) {
+        } else if self.tasks.load_or_save_in_progress(tab.id) {
             Icon::SAVE.size(icon_size)
-        } else if t.is_dirty() {
+        } else if tab.is_dirty() {
             Icon::CIRCLE.size(icon_size)
         } else {
             Icon::CHECK_CIRCLE.size(icon_size)
         };
 
         let padding_x = 10.;
-        let w = 160.;
+        let w = if tab.is_closing { 40. } else { 160. };
         let h = 40.;
 
         let (tab_label_rect, tab_label_resp) = ui.allocate_exact_size(
@@ -442,8 +445,31 @@ impl Workspace {
             tab_label_resp.scroll_to_me(None);
         }
 
+        // closing (just draw status icon)
+        if tab.is_closing {
+            if !self.tasks.load_or_save_queued(tab.id)
+                && !self.tasks.load_or_save_in_progress(tab.id)
+            {
+                result = Some(TabLabelResponse::Removed);
+            }
+
+            let icon_draw_pos = egui::pos2(
+                tab_label_rect.min.x + padding_x,
+                tab_label_rect.center().y - status_icon.size / 2.0,
+            );
+
+            let icon: egui::WidgetText = (&status_icon).into();
+            let icon = icon.into_galley(
+                ui,
+                Some(TextWrapMode::Extend),
+                status_icon.size,
+                egui::TextStyle::Body,
+            );
+            ui.painter()
+                .galley(icon_draw_pos, icon, ui.visuals().text_color());
+        }
         // renaming
-        if let Some(ref mut str) = t.rename {
+        else if let Some(ref mut str) = tab.rename {
             let res = ui
                 .allocate_ui_at_rect(tab_label_rect, |ui| {
                     ui.add(
@@ -481,7 +507,7 @@ impl Workspace {
 
             // release focus to cancel ('esc' or click elsewhere)
             if res.lost_focus() {
-                t.rename = None;
+                tab.rename = None;
             }
         } else {
             // interact with button rect whether it's shown or not
@@ -494,7 +520,7 @@ impl Workspace {
                     .expand(2.0);
             let close_button_resp = ui.interact(
                 close_button_rect,
-                Id::new("tab label close button").with(t.id),
+                Id::new("tab label close button").with(tab.id),
                 Sense { click: true, drag: false, focusable: false },
             );
 
@@ -554,15 +580,15 @@ impl Workspace {
                 .style_mut(|s| s.visuals.menu_rounding = (2.).into());
             ui.interact(
                 status_icon_rect,
-                Id::new("tab label status icon").with(t.id),
+                Id::new("tab label status icon").with(tab.id),
                 Sense { click: false, drag: false, focusable: false },
             )
             .on_hover_ui(|ui| {
-                let text = if self.tasks.load_or_save_queued(t.id) {
+                let text = if self.tasks.load_or_save_queued(tab.id) {
                     "save queued"
-                } else if self.tasks.load_or_save_in_progress(t.id) {
+                } else if self.tasks.load_or_save_in_progress(tab.id) {
                     "save in progress"
-                } else if t.is_dirty() {
+                } else if tab.is_dirty() {
                     "unsaved changes"
                 } else {
                     "all changes saved"
@@ -573,7 +599,7 @@ impl Workspace {
                 ui.add(egui::Label::new(text));
 
                 let last_saved = {
-                    let d = time::Duration::milliseconds(t.last_saved.elapsed().as_millis() as _);
+                    let d = time::Duration::milliseconds(tab.last_saved.elapsed().as_millis() as _);
                     let minutes = d.whole_minutes();
                     let seconds = d.whole_seconds();
                     if seconds > 0 && minutes == 0 {
@@ -595,7 +621,7 @@ impl Workspace {
             });
 
             // draw text
-            let text: egui::WidgetText = (&t.name).into();
+            let text: egui::WidgetText = (&tab.name).into();
             let wrap_width = if show_close_button {
                 w - (padding_x + status_icon.size + padding_x + padding_x + x_icon.size + padding_x)
             } else {
@@ -608,7 +634,7 @@ impl Workspace {
             text_rect.max.x = close_button_rect.min.x;
             ui.interact(
                 text_rect,
-                Id::new("tab label text").with(t.id),
+                Id::new("tab label text").with(tab.id),
                 Sense { click: false, drag: false, focusable: false },
             )
             .on_hover_ui(|ui| {
@@ -679,6 +705,7 @@ impl Workspace {
 enum TabLabelResponse {
     Clicked,
     Closed,
+    Removed,
     Renamed(String),
 }
 
--- a/libs/content/workspace/src/show.rs
+++ b/libs/content/workspace/src/show.rs
@@ -23,7 +23,7 @@ impl Workspace {
 
         self.process_updates();
         self.process_keys();
-        self.status.populate_message();
+        self.status.message = self.status_message();
 
         if self.is_empty() {
             self.show_empty_workspace(ui);
--- a/libs/content/workspace/src/show.rs
+++ b/libs/content/workspace/src/show.rs
@@ -6,6 +6,7 @@ use egui::{EventFilter, Id, Key, Modifiers, Sense, TextWrapMode, ViewportCommand
 use std::collections::HashMap;
 use std::mem;
 use std::time::{Duration, Instant};
+use tracing::instrument;
 
 use crate::output::Response;
 use crate::tab::{TabContent, TabFailure};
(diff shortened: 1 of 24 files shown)


libs/content/workspace/src/show.rs
```diff
@@ -458,8 +461,31 @@ impl Workspace {
...
+            let icon: egui::WidgetText = (&status_icon).into();
+            let icon = icon.into_galley(
+                ui,
+                Some(TextWrapMode::Extend),
+                status_icon.size,
+                egui::TextStyle::Body,
+            );
+            ui.painter()
+                .galley(icon_draw_pos, icon, ui.visuals().text_color());
+        }
```
tvanderstad, line 486: in `is_closing` state, tab closes itself if there are no tasks associated and draws with just the status icon in the tab strip

libs/content/workspace/src/task_manager.rs
```diff
@@ -80,9 +87,6 @@ pub struct LoadRequest {
 #[derive(Clone)]
 pub struct SaveRequest {
     pub id: Uuid,
-    pub old_hmac: Option<DocumentHmac>,
-    pub seq: usize,
-    pub content: String,
```
tvanderstad, line 85: queue a save request with just the id; task manager will grab the rest when it's time to launch

libs/content/workspace/src/task_manager.rs
```diff
@@ -273,6 +382,15 @@ impl TaskManager {
             if tasks.load_or_save_in_progress(id) {
                 continue;
             }
+            if tasks
+                .completed_saves
+                .iter()
+                .any(|completed_save| completed_save.request.id == id)
+            {
+                // result of completed save must be processed before another save to the same file; this guarantees
+                // that the hmac from the completed save is used for the next, preventing a ReReadRequired error
```
tvanderstad, line 391: note

libs/content/workspace/src/task_manager.rs
```diff
@@ -343,55 +490,91 @@ impl TaskManager {
...
-            // content cloned; one copy sent to disk and other retained in UI to represent on-disk version for merge
-            // first step to alleviate: https://github.com/lockbook/lockbook/issues/3241
             let request = queued_save.request.clone();
             let in_progress_save = InProgressSave::new(queued_save);
+            let content = {
+                let mut result = None;
+                for tab in tabs {
+                    if tab.id == request.id {
+                        let start = Instant::now();
+                        result = tab.content.clone();
```
tvanderstad, line 528: note: `tab.content` cloned on the UI thread

libs/content/workspace/src/task_manager.rs
```diff
@@ -343,55 +490,91 @@ impl TaskManager {
...
-                let new_hmac_result = self_clone.core.safe_write(
-                    request.id,
-                    request.old_hmac,
-                    request.content.into(),
-                );
+                let (old_hmac, seq, content) = match content {
+                    TabContent::Markdown(editor) => {
+                        (editor.hmac, editor.buffer.current.seq, editor.buffer.current.text)
+                    }
+                    TabContent::Svg(svg) => (svg.buffer.open_file_hmac, 0, svg.buffer.serialize()),
```
tvanderstad, line 570: note: `svg.buffer` serialized on a background thread

libs/content/workspace/src/task_manager.rs
```diff
@@ -249,10 +285,83 @@ impl TaskManager {
...
+    }
+
     /// Launches whichever queued tasks are ready to be launched, moving their status from queued to in-progress.
     /// In-progress tasks have status moved to completed by background threads. This fn called whenever a task is
-    /// queued (on the UI thread) or completed (on the background thread).
-    fn check_launch(&self) {
+    /// queued or explicitly - background threads will not call it and will instead only call request_repaint() when
+    /// done - so it's the UI's responsibility to check in on it from time-to-time. This is necessary so that the UI
+    /// can interject between tasks that are queued and tasks that they are queued behind i.e. to provide the latest
+    /// hmac and file content so that a save succeeds.
```
tvanderstad, line 363: note

tvanderstad (commented):
interesting bits for review

libs/content/workspace/src/syncing.rs
```diff
@@ -3,78 +3,65 @@ use lb_rs::model::work_unit::WorkUnit;
...
         let server_ids = work.work_units.iter().filter_map(|wu| match wu {
             WorkUnit::LocalChange { .. } => None,
             WorkUnit::ServerChange(id) => Some(*id),
         });
 
         for id in server_ids {
             for i in 0..self.tabs.len() {
-                if self.tabs[i].id == id {
+                if self.tabs[i].id == id && !self.tabs[i].is_closing {
+                    tracing::debug!("Reloading file after sync: {}", id);
```
Parth, line 44: nit (with `use tracing::*` for uniformness)
```suggestion
                    debug!("Reloading file after sync: {}", id);
```

libs/content/workspace/src/tab/markdown_editor/editor.rs
```diff
@@ -397,3 +397,17 @@ impl Editor {
...
+        Self::new(
+            self.core.clone(),
+            &self.buffer.current.text,
+            self.file_id,
+            self.hmac,
+            self.needs_name,
+            self.appearance.plaintext_mode,
+        )
+    }
+}
```
Parth, line 413: I'm not sure this is suitable behavior for `Clone`. I imagine this resets some portion of the state of the editor, and I would not expect clone to behave this way. I expect `Clone` to be a deep copy of all state unless there's some smart pointer business going on.

Parth: Easy enough to implement a fn that describes whatever you're trying to do.

libs/content/workspace/src/tab/mod.rs
```diff
@@ -78,6 +60,18 @@ impl std::fmt::Debug for TabContent {
...
+impl Clone for TabContent {
+    fn clone(&self) -> Self {
+        match self {
+            TabContent::Markdown(editor) => TabContent::Markdown(editor.clone()),
+            TabContent::Svg(svg) => TabContent::Svg(svg.clone()),
+            TabContent::Image(_) => unimplemented!("images aren't cloneable"),
+            TabContent::Pdf(_) => unimplemented!("pdfs aren't cloneable"),
+        }
+    }
+}
```
Parth, line 73: This seems more acceptable to me, but I would also be pretty surprised to find a `.clone()` crashing my program.

libs/content/workspace/src/task_manager.rs
```diff
@@ -343,55 +490,91 @@ impl TaskManager {
...
+                for tab in tabs {
+                    if tab.id == request.id {
+                        let start = Instant::now();
+                        result = tab.content.clone();
+                        let time = Instant::now().duration_since(start);
+                        if time > Duration::from_millis(100) {
+                            warn!(
+                                "spent {}ms on UI thread cloning content for file {} for background thread to save it",
+                                request.id,
+                                time.as_millis()
```
Parth, line 534: I believe duration has put some effort into readability by impl'ing a display, could be wrong

libs/content/workspace/src/task_manager.rs
```diff
@@ -304,9 +429,31 @@ impl TaskManager {
...
+                "launching {} loads, {} saves, {} syncs, and {} sync status updates; {} loads, {} saves, {} syncs, and {} sync status updates remain queued",
+                loads_to_launch.len(),
+                saves_to_launch.len(),
+                sync_to_launch.is_some() as usize,
+                sync_status_update_to_launch.is_some() as usize,
+                tasks.queued_loads.len(),
+                tasks.queued_saves.len(),
+                tasks.queued_syncs.len(),
+                tasks.queued_sync_status_updates.len()
+            );
```
Parth, line 455: for your consideration, you may find the following syntax more ergonomic / readable: https://docs.rs/tracing/latest/tracing/macro.debug.html#examples

libs/content/workspace/src/task_manager.rs
```diff
@@ -457,30 +649,87 @@ impl TaskManager {
...
-                        .timing
-                        .started_at
-                        .duration_since(completed_sync.timing.queued_at);
-                    if in_progress_time > Duration::from_secs(1) {
-                        warn!("Sync took {}s", in_progress_time.as_secs());
+                    let timing = CompletedTiming::new(in_progress_sync.timing);
+                    let in_progress_time = timing.completed_at.duration_since(timing.started_at);
+                    if let Err(err) = &status_result {
+                        error!("sync failed ({}ms): {:?}", in_progress_time.as_millis(), err);
+                    } else if in_progress_time > Duration::from_secs(1) {
```
Parth, line 659:
```suggestion
                    } else if in_progress_time > Duration::from_secs(5) {
```

my gut on what I would find "alarming", but even still people have all sorts of networks 🤷

libs/content/workspace/src/workspace.rs
```diff
@@ -435,31 +453,92 @@ impl Workspace {
...
+            warn!("processing completed sync took {:?}", start.elapsed());
+        }
+
+        let start = Instant::now();
+        if let Some(update) = completed_sync_status_update {
+            self.sync_status_update_done(update)
+        }
+        if start.elapsed() > Duration::from_millis(100) {
+            warn!("processing completed sync status update took {:?}", start.elapsed());
+        }
```
Parth, line 480: this amount of timing is making this portion pretty unreadable but your call on whether worthwhile. Maybe just representative of some suffering you incurred.

log messages will all have timestamps and you could always just `instrument` the underlying function. So although the warning helps stick out in the log I would go for overall less log volume (probably the thing that's causing you to add this in the first place).

But actually up to you as the potential debugger of this stuff.

libs/lb/lb-rs/src/repo/mod.rs
```diff
@@ -58,23 +58,35 @@ impl<'a> LbTx<'a> {
...
         //     .await
         //     .unwrap();
 
         let mut guard = self.db.write().await;
         let tx = guard.begin_transaction().unwrap();
 
+        if start.elapsed() > std::time::Duration::from_millis(100) {
+            tracing::warn!("readwrite transaction lock acquisition took {:?}", start.elapsed());
+        }
+
```
Parth, line 89: nice touch

you may want to move this up one line to just time the `write().await`

libs/lb/lb-rs/src/service/logging.rs
```diff
@@ -42,7 +42,9 @@ pub fn init(config: &Config) -> LbResult<()> {
                 .with_target(false)
                 .with_filter(lockbook_log_level)
                 .with_filter(filter::filter_fn(|metadata| {
-                    metadata.target().starts_with("workspace")
+                    metadata.target().starts_with("lb_rs")
+                        || metadata.target().starts_with("dbrs")
+                        || metadata.target().starts_with("workspace")
```
Parth, line 47: while desirable I believe this will make our CLI unusable. However I moved the entire cli config into core, and you could see if those flags are already suitable.

I think they're not, you'll probably want to add a `stdout_logs` flag and `false` it on cli.

but it will be nice to `true` it on the tests.

libs/lb/lb-rs/src/service/network.rs
```diff
@@ -68,7 +68,13 @@ impl Network {
             client_version: client_version.clone(),
         })
         .map_err(|err| ApiError::Serialize(err.to_string()))?;
+
+        if serialized_request.len() > 1024 * 1024 {
+            tracing::warn!("making network request with {} bytes", serialized_request.len());
+        }
```
Parth, line 74: I think canvas users will see this for practically every `svg`. Maybe 10mb is a more suitable threshold

Parth: also maybe you want 10mb in plaintext somewhere for ease

Parth (changes requested):
- Didn't get a chance to think too hard if your changes are breaking across ffi anywhere like android just want to flag for someone to double check (like @smailbarkouch)
- Can we get some screenshots of the various states a tab could be in (I imagine you'll find it easy to advertise this change to stakeholders with this accessible)
- could we benefit from some level of de-duplication? does it ever really make sense to launch back to back syncs

libs/content/workspace/src/workspace.rs
tvanderstad: What I'm going for here is that when anyone on the team introduces performance issues, the app starts spitting out warnings with a decent indicator of the nature of the issue. Is that different from what you thought? I don't know how log messages having timestamps helps me with that and I'm not sure how to accomplish that by instrumenting a function.

Parth: yeah I think that is a good idea. There is no way to achieve this trivially, maybe I could add it to our own implementation of a tracing subscriber. 

I just wonder if we've already achieved a good amount more visibility by having the logs in our face in more situations (possibly closing https://github.com/lockbook/lockbook/issues/3108). With a bit of coaching we could prompt team members to look at these in more situations.

Possibly also, for dev-time builds I would be down to timeout lock acquisition as a stronger kick in the butt if the team introduces regressions. 

mainly I'm concerned about the readability hit.

tvanderstad: feels like we're still pretty early in our journey of logging useful things, is it cool if we do this the messy way until we iron some patterns out, then invest in readability? I am willing to pay the price in readability in this instance to make the logs something I want to have open and watch while I'm developing

maybe a helper like this would go a long way?
```
trait InstantExt {
    fn warn_if_elapsed(&self, duration: Duration)
} 

impl InstantExt for std::time::Instant {
    fn warn_if_elapsed(&self, duration: Duration) {
        if self.elapsed() > duration {
            warn!() // todo maybe this whole thing needs to be a macro so caller can put what they want here
        }
    }
}
```

Parth: I am down to iron out later

I think this is good use case for that kind of extension trait and helper if you wanted to use it. I think that reduces enough noise to make it a viable pattern immediately

on board with a macro so long as the surrounding block doesn't end up in the macro just the timing statement.

also on board with what you have to just take an `&str` for now, iirc that's all you need at the moment.

but yeah a lil helper in `lb/logging.rs` sounds like a pretty solid addition

tvanderstad:
> Didn't get a chance to think too hard if your changes are breaking across ffi
I didn't change any of the ffi data structures but we should experience the app on all platforms before releasing as usual

> Can we get some screenshots
Will get to this soon!

> de-duplication?
There's room for creativity/investment in the core logic of the task manager. Right now the deduplication happens when tasks are launched, which works, but if they were deduplicated when queued then the queue would have a bounded size. If a sync is in progress when you queue one, we'll do that one after, but we'll effectively never queue up more than one. This is just so you know that if you call `queue_sync` and sync will actually happen soon

Parth:
Nice work, glad we're getting this situation under control

Parth (approved):

PR #3312, opened 2025-01-13
apple: rewrite

smailbarkouch:
This PR is the mass rewrite of all our Apple clients for improved readability and stability. On the way, I'm going to fix a host of other bugs. Check out this document for the initial proposal: `lb://17ac0970-535f-43b3-ab69-0e5f8db57b64`.

State is propagated through three different means, `EnvironmentObject`, `StateObject`, and `ObservedObject`. You can learn more about them [here](https://medium.com/@chaitali25/swiftui-exploring-observableobject-observedobject-stateobject-and-environmentobject-3b699cccdba4). There is a much larger emphasis on using `StateObject` in this rewrite so that more views are independent of one another.

I'm also leaning into the idea that views are a "function of state", and thus have tried to reduce the amount of business logic in views. Examples of this are my various sheet views. This has allowed me to get Previews for every view written thus far (with minimal effort).

What's left
+ MacOS: Most of the UI has already been made (except for file tree), it just needs to be layed out correctly.
    + File tree
    + Add relevant toolbar buttons
    + Add back search (just need to use the wrapper)
    + Add back bottom bar
+ iOS
    + Pull to open sidebar
    + Keep tree state when sidebar is closed
    + Fix UI glitches with opening search

I also need to test billing to make sure I didn't break anything.


Parth:
recording some thoughts for myself:
- I would expect a whole category of log noise ("Published UI change from background thread", and other similar log messages) to go away as a result of doing things more idiomatically.

Parth:
This PR is exciting stuff I'm really excited to get this merged. Some thoughts as I explore:

# macOS

image

I think this is a nice touch. We should probably discuss if where we want to be using this sort of symbols. Is this just for smarts? For ML? For LLMs? etc. I'm down to do for any smarts / predictions. This doesn't seem to be there on iOS tho. So that's a bit strange.

image

maybe this folder ought to be filled? Just seems a little inconsistent with the ✨ 

I also somewhat expect these to be hidable, but doesn't need to be a merge blocker. I'd also expect hiding it to be persisted, but that's a larger conversation we can have.

I still see a lot of these messages:

image

was hopping this refactor would be the last time we basically saw any such noise. Basically any interaction with opening and closing a file will show this on macOS.

image

this is a lot more aesthetic, and I particularly like the system level authentication you've added. But this debug view doesn't offer me any way to select and copy text. Ideally it would be a one click action to copy the whole debug view to the clipboard.

# iOS

pulling to refresh also causes those warnings. not sure why my landing page looks like this... 
image

ipad so far seems a lil buggy, I opened a doc and closed it and the file tree seems stuck on the same doc and I can't click on anything. I got these not sure if relevant:

image

steps to repro: open a doc, close it (kinda hard but maybe also the case on master), and then you're pretty much busted no other docs can be opened.

I also see multiple instance of what looks like a title to me, but only when there's only one tab, there's at-least one strange thing going on there.

interestingly as I cycle through the multi tasking states and my ipad "becomes an iphone" it resolves the issue and the doc is open so there's still some multitasking weirdness going on. Let me see what it feels like on just my iPhone:

I def want to be able to pull from the sidebar. 

iPhone seems much more stable than iPad. 

image

I do personally like this whole direction of actual (and same) file tree that iPad and other clients use. What level of control do we have over the proportion of the screen that is a file tree? 

I think even with the same amount of file tree width, a lot of space is wasted on the extreme right -- maybe that's the part that's easier to solve. But also I wonder if there's some alternate approach that we can explore that would not have as much empty space on the left. But I don't think that's an immediate concern.

# Closing thoughts
This feels close to being ready, I think these are the things we need to do before we can merge:
- ipad stability needs to be much higher.
- we setup maybe a 90 minute session with two goals: you walk me through the new pattern in great detail (45m?) and we spend 30minutes doing some collective QA maybe with some new accounts, and an emphasis on billing. And then we finally take 15m to make sure the issues are in a reasonable state.

Parth:
Search bug discovered: shared files seem to show up strangely (running.md)

Parth:
Path search found to be using 100% cpu

Parth:
https://github.com/lockbook/lockbook/pull/3312#issuecomment-2845399726

found to be all resolved

smailbarkouch:
Fix logout

smailbarkouch:
fix empty suggest docs state

Parth:
> I find myself needing to tap buttons a few times to get them to work (experienced with side bar toggle button and tab switch button) - do I just have skill issues?

me and adam felt similarly fyi

Parth:
@smailbarkouch if you rebase you'll have the fix for `ERROR` in search results

libs/lb/lb-rs/src/service/logging.rs
```diff
@@ -34,7 +34,7 @@ pub fn init(config: &Config) -> LbResult<()> {
                 .boxed(),
         );
 
-        if config.stdout_logs {
+        if false {
```
Parth, line 37:
```suggestion
        if config.stdout_logs {
```

PR #3333, opened 2025-01-22
lb-rs: only record activity to suggested docs for user initiated interactions

Parth:


libs/lb-fs/src/fs_impl.rs
```diff
@@ -56,7 +56,7 @@ impl NFSFileSystem for Drive {
         let entry = data.get_mut(&id).unwrap();
         let id = entry.file.id;
 
-        let mut doc = self.lb.read_document(id).await.unwrap();
+        let mut doc = self.lb.read_document(id, false).await.unwrap();
```
Parth, line 59: would consider changing these to true down the line, it's just a very noisy protocol

libs/content/workspace/src/task_manager.rs
```diff
@@ -649,7 +649,7 @@ impl TaskManager {
     #[instrument(level = "warn", skip(self), fields(thread = format!("{:?}", thread::current().id())))]
     fn background_load(&self, request: LoadRequest) {
         let id = request.id;
-        let content_result = self.core.read_document_with_hmac(id);
+        let content_result = self.core.read_document_with_hmac(id, true);
```
tvanderstad, line 652: I think this could also be triggered by a sync pulling a change to this file

Parth: thanks for having a look, i think that will still lead to good data (compared to what is in it's absence)

Issue #3380, opened 2025-01-31
Unable to type certain characters in german keyboard layout on macos


tvanderstad:
To reproduce:
* add german to keyboard layouts (I tried the first two of the five and both exhibited issues)
* select german layout
* type option+`]`: nothing happens (in other apps it inserts characters)

tvanderstad:
https://discord.com/channels/1014184997751619664/1334944546110570587/1334944552561152061

PR #3400, opened 2025-02-09
macOS: fix german keyboard input

smailbarkouch:
fixes https://github.com/lockbook/lockbook/issues/3380

### DIFF
--- a/libs/content/workspace-ffi/src/apple/macos/api.rs
+++ b/libs/content/workspace-ffi/src/apple/macos/api.rs
@@ -28,11 +28,15 @@ pub unsafe extern "C" fn key_event(
     let Some(key) = NSKeys::from(key_code) else { return };
 
     // Event::Text
-    let alt_only = modifiers.alt && !modifiers.ctrl && !modifiers.shift && !modifiers.mac_cmd  && !modifiers.command;
+    let alt_only = modifiers.alt
+        && !modifiers.ctrl
+        && !modifiers.shift
+        && !modifiers.mac_cmd
+        && !modifiers.command;
     let text_modifiers = modifiers.shift_only() || alt_only || modifiers.is_none();
     let text = CStr::from_ptr(characters).to_str().unwrap().to_string();
     let is_valid_text = text.chars().any(|c| !c.is_control());
-    
+
     if pressed && text_modifiers && is_valid_text {
         obj.raw_input.events.push(Event::Text(text));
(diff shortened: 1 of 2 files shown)


libs/content/workspace-ffi/src/apple/macos/api.rs
```diff
@@ -27,15 +27,15 @@ pub unsafe extern "C" fn key_event(
...
     let Some(key) = NSKeys::from(key_code) else { return };
 
-    // Event::Text
-    let text_modifiers = modifiers.shift_only() || modifiers.is_none();
-    if pressed && text_modifiers && key.valid_text() {
-        let text = CStr::from_ptr(characters).to_str().unwrap().to_string();
-        obj.raw_input.events.push(Event::Text(text));
-    }
+    let alt_without_ctrl_cmd =
+        modifiers.alt && !modifiers.ctrl && !modifiers.mac_cmd && !modifiers.command;
```
tvanderstad, line 31: Could this be `alt_only`?

tvanderstad: Also you don't need to check `mac_command` (like command but excludes ctrl on windows/linux)

libs/content/workspace-ffi/src/apple/macos/api.rs
```diff
@@ -27,15 +27,15 @@ pub unsafe extern "C" fn key_event(
...
+    let text_modifiers = modifiers.shift_only() || alt_without_ctrl_cmd || modifiers.is_none();
+    let text = CStr::from_ptr(characters).to_str().unwrap().to_string();
+    let is_valid_text = text.chars().any(|c| !c.is_control());
```
tvanderstad, line 34: `.to_string()` makes a copy that's unnecessary bc you're just iterating to check for control chars

tvanderstad (changes requested):
A few review comments but mainly: navigating a doc with arrow keys inserts a bunch of tofu

Parth:
no tofu

libs/content/workspace-ffi/src/apple/macos/api.rs
smailbarkouch: You can insert some characters using `alt` + `shift`. I didn't know that, I'll remove regular command in that check then.

smailbarkouch:
No more tofu 🙏

tvanderstad (approved):

PR #3438, opened 2025-02-18
lb-rs: search and status refinements

Parth:


libs/lb/lb-rs/src/subscribers/status.rs
```diff
@@ -0,0 +1,222 @@
...
+    pub update_required: bool,
+
+    /// metadata or content for this id is being sent to the server
+    pub pushing_files: Vec<Uuid>,
+
+    /// following files need to be pushed
+    pub dirty_locally: Vec<Uuid>,
+
+    /// metadata or content for this id is being from the server
+    pulling_files: Vec<Uuid>,
```
Parth, line 59: pub

PR #3451, opened 2025-02-25
infra: release automation

Kousay-Jebir:
So far you can choose either to automatically create a github release and thus the platform chosen for the workflow to be ran on doesn't matter so it default to macos or, if you choose the all option then the platform matters since different steps are executed based on the chosen platform


Kousay-Jebir:
edit: changed this to only perform github releases

Issue #3364, opened 2025-01-25
content search results sometimes don't emphasize whole match


tvanderstad:
<img width="385" alt="Image" src="https://github.com/user-attachments/assets/549e7503-2096-448e-a936-1acbd13c54af" />

Issue #3363, opened 2025-01-25
content search results sometimes contain completely empty entries


tvanderstad:
<img width="385" alt="Image" src="https://github.com/user-attachments/assets/8ac3c936-61ac-427c-ba0c-c4392c0371d8" />

Parth:
I believe you won't find this behavior in egui, please do try and lmk

I have a feeling search may report "fresher" results that a long lived swift client is not aware of and this is what that looks like.

Issue #3362, opened 2025-01-25
content search results sometimes stutter/continuously change for a few seconds


tvanderstad:
https://github.com/user-attachments/assets/25818ba1-1ac1-41fb-997c-139a7d42a71c

PR #3468, opened 2025-03-21
lb-rs: much better search

Parth:
current search is slow and the quality leaves much to be desired.

In this PR I experiment with [tantivy](https://github.com/lockbook/lockbook) designed to be the internals of a product like elastic search. Because it's a relatively low level library we can embed it in-process and search very quickly.

Currently, searching for a phrase like "2025" takes `40ms` on a powerful computer (and generally 2x that in most places).

The same search in tantivy takes `1ms` or so, making it something we can do on the main thread with ease. Even though it's doing significantly more work upon startup (indexing) the index-build time is only 1-10ms slower.

Tantivy also gives us access to advanced search features that we could use more widely the most interesting of which is it's advanced search capabilities. [Read more here](https://docs.rs/tantivy/latest/tantivy/query/struct.QueryParser.html).

I can see a world where we insert all sorts of facts about a document into this local search engine like extension, ownership information, last modified, etc. And users can search something like: "extension: .md AND owner: parth AND shared: false". This could be wrapped in some fancy UI features or we can just document the syntax and let power users do the work (like github does).

Tantivy is weak in the domains of fuzzy finding, it's part of the reason it's faster (but it's also faster due to the fact that it indexes documents before hand). As a user the only place I really want fuzzy matches is for path searches. We could just use the existing library, but despite it's name I don't think it faithfully implements search the way sublime fuzzy behaves. [This post](https://www.forrestthewoods.com/blog/reverse_engineering_sublime_texts_fuzzy_match/) seems like a credible description of how sublime works (it seems to have some input from the sublime team). A key difference (and a complaint we've received) is that it doesn't take total size of the [match](https://github.com/Schlechtwetterfront/fuzzy-rs/blob/master/src/matching.rs) into account (the negative scoring from the blog post). So I'm inclined not to use that anyways. 

So for a good path search experience I have one of three options that I have yet to explore (they're listed in the order I intend to explore them in):
1. Tantivy does have a regex style query, I could explore transforming path queries to a relatively basic regex that performs subsequence matching.
2. Just have a separate path search tree traversal matching algorithm that lives alongside tantivy, and just use tantivy for document and advanced searching.
3. Tantivy seems relatively extensible, and it's possible to add custom implementations at all parts of the pipeline: tokenization, indexing, searching, and scoring. I could look at bringing something like multi term prefix searching to tantivy.

That's the last unknown for this PR, it uses the same API as the existing search implementation so we'll be able to drop this in without having to change any platform specific code, however as a code cleanup I think there are some modifications I would like to make to contract:
- encode the idea of a query matching multiple criteria (path, docs, and more) rather than it being represented like a mutually exclusive enum
- don't model the highlighted region as a `Vec<usize>` that represents which characters to emphasize, but rather encode these as a `Vec<Range<usize>>` for more compactness.
- Explore representing the highlighted regions as markdown instead. SwiftUI can just render this as as, can other platforms? Would lighten implementation burden for the UI dev. HTML and arbitary delimiters are also an option.
- ideally capture any fields the UI would show as part of the search result itself to avoid having to reason about consistency / implementation issues, I believe the current Apple implementations behave poorly if you return an ID they're not aware of (like a pending share).

fixes: https://github.com/lockbook/lockbook/issues/3363
fixes: https://github.com/lockbook/lockbook/issues/3362
fixes: https://github.com/lockbook/lockbook/issues/3364
fixes: https://github.com/lockbook/lockbook/issues/3194


Parth:
+ [ ] handle delete
+ [ ] reduce code duplication
+ [ ] address clippy 
+ [ ] fix tests

Parth:
I ended up doing number 2 for the path match and the results are promising and at a surface level have been :heart:'d by @tvanderstad as part of QA.

Parth:
additional things that are unsatisfactory with this contract: don't need score. don't need 1-n relationship with a content match and it's snippets.

PR #3479, opened 2025-03-28
windows: new clippy warnings

Kousay-Jebir:
- Tried to annotate types for transmute()
- Removed an unused import
- added #[allow(dead_code)] for Message struct to suppress unused struct fields warnings


Parth:
Ideally you would paste the clippy messages somewhere here

Parth:
`#[allow(dead_code)]` I think we don't want these, any reason it's required?

Kousay-Jebir:
> Ideally you would paste the clippy messages somewhere here

This is the initial output before changes 
image

Kousay-Jebir:
> `#[allow(dead_code)]` I think we don't want these, any reason it's required?

Didn't know how i would get rid of unused struct field warnings without adding unnecessary code that's why i opted for it

PR #3494, opened 2025-04-02
workspace: pan and zoom in image viewer 

ad-tra:


Parth:
@ad-tra you didn't squash 😠

Issue #3054, opened 2024-10-31
restore android CI


Parth:
will be disrupted as a result of #2798

Parth:
I believe we are happy with the state of android CD

PR #3502, opened 2025-04-03
workspace-ffi: Remove unused dependencies

Kousay-Jebir:
`cargo check --all-features`
image


Parth (changes requested):
so `workspace-ffi` is used by our android app and our iOS app. And at a very minimum `apple` requires `pollster`. Can't tell about the others immediately.

I see that you did `--all-targets` and unfortunately the word `target` is an overused rust keyword:

From cargo check's perspective `--all-targets` means: `lib`, `bin`, `test`, `bench`, `docs`, etc.

And presumably both of us expect it to mean it in the sense that `rustup` means `target` (`aarch64-apple-ios`).

One thing you can do to sanity check results is to just seach for the library in the module. For instance, you'll find pollster. I think `libc` doesn't show up and may be unused or have a slightly more complicated story, I'll check on that in a bit.

Nonetheless, we have to think about what to do next:
- Can we run udeps with the additional context for android, for example. Can you say I want to check for unused dependences *from the perspective of android*? On Windows & Linux you should be able to test everything besides apple. Ideally you understand what command would break if you remove `pollster`. Installing android toolchain may be complicated and not worthwhile, but may also help you understand why #3054 is helpful for us (on the CI track).
- If we can't get udeps to think about our additional context, maybe we could narrow it's scope, and only run udeps in a handful of targeted locations where we can be confident of the results. Running in `lb-rs`, `egui`, `workspace`, `server`, `cli` etc on linux is a huge amount of the value of this work anyways.
- Another option that's growing on me is that the code should just reflect the situation. For example see line 26 of this file -- we only use `ndk-sys` on `android` we could just give these dependencies that treatment.
- The flipside of that is to change the code so that it's just always present on all platforms (so it doesn't show up as unused) but this is non ideal because it'll prob make our binaries larger.

So probably I lean towards using the cfg flags in the Cargo.toml.

Also I would rarely use screenshots for text content, harder to search it.

Lmk if any of the above needs clarification.

Kousay-Jebir:
Does it make sense to just ignoe these two dependencies in udeps's ignore file 

- i now realize that pollster is used for ffi which is a limitation by udeps where it can't capture it at all 
- Yep as you said i thought that all-targets meant the triplet target i now see that to use a specific target u need the --target option
- i suppose using udeps directly with --target for every specific target used in android / linux is not enough to fix this 
a good example for this is  ndk-sys  like u mentioned.

tried to use udeps with `aarch64-linux-android` to see what happens and i get a lot of errors 
I'm still stuck on figuring out if libc is used or not
To summarize, 
Should we just ignroe pollster and libc and keep using udeps normally ?

Parth:
fundamentally these dependencies *are unused* on a normal build, and so my preferred path forward is to just not include them on a normal build. Checkout how @smailbarkouch imported `ndk-sys` for just `android` [here](https://github.com/lockbook/lockbook/blob/master/libs/content/workspace-ffi/Cargo.toml#L26-L27).

`workspace-ffi` is only used by `vendor = apple` (iOS and macOS) and by `android`. So I would be down to make our `Cargo.toml` reflect that. The corresponding code is already "hidden" to only be invoked for those platforms:
- android: https://github.com/lockbook/lockbook/blob/master/libs/content/workspace-ffi/src/android/mod.rs
- apple: https://github.com/lockbook/lockbook/blob/master/libs/content/workspace-ffi/src/apple/mod.rs

so imo let's just carry that structure forward to the `Cargo.toml`.

Ideally then `udeps` won't bother us about these things, because the dependencies won't show up where it's not needed.

- ignoring these udeps in some specific file is an option, but what I described is probably cleaner and has other benefits
- I think `libc` is genuinely not used. If it's used at all, it'll be on `apple` and we'd catch this pretty quick on `master` so you can feel free to remove it for now.

Parth:
fyi the reason you probably can't just run android is because it requires the [native dev kit](https://developer.android.com/ndk). Which is not managed by cargo (but it probably could be to make people's lives easier. My guess is that no one got around to it or there's a licensing conflict that just needs to be worked through).

On apple you can just install the targets and the libraries are baked into Xcode or some other fundamental system dependency that you need for rust anyways, but obviously you need a Mac for that.

Soon we'll have a team wide Mac (thanks to @ad-tra) that you would just be able to ssh into to test things out (probably).

The description for how to do android dev is [here](https://github.com/lockbook/lockbook/blob/master/docs/guides/build/android.md) if you wanted to sort out the ndk errors

Kousay-Jebir:
`cargo +nightly udeps -p workspace-ffi' 's output is now
`All deps seem to have been used.`

PR #3512, opened 2025-04-11
infra: auto merging version bump commit into master

Kousay-Jebir:
Thought about also also adding the deletion of the bump branch, any thoughts on that ?

### DIFF
--- a/utils/releaser/src/version.rs
+++ b/utils/releaser/src/version.rs
@@ -168,7 +168,7 @@ fn push_to_git(version: &str) {
     Command::new("bash")
         .args([
             "-c",
-            &format!("git checkout master && git pull origin master && git add -A && git commit -m 'bump-{version}' && git push origin master")
+            &format!("git fetch origin && git checkout master && git pull origin master && git add -A && git commit -m 'bump-{version}' && git push origin master")
         ])
         .assert_success()
 }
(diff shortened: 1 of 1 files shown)


Kousay-Jebir:
Still testing this on my fork

Kousay-Jebir:
Progress so far: 
this gave the following error 
`! [rejected]        bump-0.9.23 -> bump-0.9.23 (fetch first)`
so i tested again by adding `git fetch origin` before checking out to the master branch 
Then it gave a new error 
`fatal: refusing to merge unrelated histories`
I'm not sure if this is normal and we should just merge by adding the `--allow-unrelated-histories` flag 
If it's not normal i'll try to dig more into seeing what caused it.
Or, should we instead create a pr instead of merging directly and enable auto merge on the pr via github instead

Parth:
All my thoughts so far:
- instead of making a branch and pushing that and then deleting it and all that, you can evolve this to just make sure we start on master, do the bump, try to do some checks to make sure things are still fine (cargo style checks), and then commit and push directly to master.
- `! [rejected]        bump-0.9.23 -> bump-0.9.23 (fetch first)` is probably because your repo has a bump commit that differs slightly from ours (different authors, resulting in a different hash?). And so it doesn't know what to do.
- I think approaching it via how I laid out in my first bullet point is likely to just solve that problem.

Kousay-Jebir:
Got this to work properly , this now performs some the following checks 
`cargo fmt --check`
`cargo check`
before pushing directly to master
master branch on my fork correctly contains the bump commit

PR #3515, opened 2025-04-16
update rust to 1.86.0

Kousay-Jebir:


Parth:
I would `git reset --hard HEAD~1` (deletes the [6f76e73](https://github.com/lockbook/lockbook/pull/3515/commits/6f76e73abbbc2405f034e781d458b6db6fcce2a5) commit)

do this, either manually or [by applying this patch](https://stackoverflow.com/a/15438863).

and then force push the new commit for linux

```patch
diff --git a/clients/linux/src/input/clipboard_paste.rs b/clients/linux/src/input/clipboard_paste.rs
index 3c013e5d..78639483 100644
--- a/clients/linux/src/input/clipboard_paste.rs
+++ b/clients/linux/src/input/clipboard_paste.rs
@@ -158,7 +158,7 @@ impl<'a> Context<'a> {
     }
 }
 
-impl<'a> Ctx<'a> {
+impl Ctx<'_> {
     fn get_targets(&self) -> Result<Vec<Atom>, Box<dyn std::error::Error>> {
         let Ctx { conn, atoms, window } = *self;
         let formats = conn
```

fortunately this is the last thing we'll have to do on this front

Parth:
looks like I do have push permissions, great

Parth (approved):

PR #3517, opened 2025-04-17
Mind Map: node names & multi touch

Kjets22:
Fully integrated touch support now enables seamless deployment on both iOS and Android platforms. In addition, URL parsing has been enhanced to extract and display descriptive title information instead of just showing a basic link.


Parth (changes requested):
This is a step in the right direction, touch support does feel great, titles do seem more meaningful. 

However when I close the mind map tab it still seems to be doing some incredibly heavy operations. Which in my mind is the sort of thing the requires it to stay in beta -- it's lack of isolation when it comes to impact for the overall application.

I would merge this if you removed the beta checks or implemented proper resource cleanup.

PR #3519, opened 2025-04-18
canvas: fix being stuck in drawing dots bug

ad-tra:


Parth:
I can break pan here by panning and switching tabs constantly. Eventually one of the pans will just cease to respond.

It does feel like an improvement because any stroke will rectify the situation.

It just doesn't feel like we've solved all category of whatever the defect is and achieved mastery over the situation.

Parth:
I can upload a video / talk through breaking if you're unable to reproduce.

Issue #3463, opened 2025-03-12
macOS stuck unable to sync with no error message (fixed by restarting)


tvanderstad:
observed on master (0.9.21), steps to repro unclear

tvanderstad:
Experienced this again this morning. It was a similar situation:
* This time using macOS 0.9.20 built from source
* Open side by side with egui on macos - I've been leaving macOS open to edit my dev log then previewing it using egui on my branch while I iterate on md++
* Eventually, edits stop syncing from macOS to egui
* macOS app reports last synced "just now" (concerning!)
* Edits do not appear on my iPhone (0.9.20)
* After restarting the macOS app, all edits are synced to all devices
* macOS debug info hung indefinitely (I waited a ~1 minute) until another restart (some kind of deadlock?)

```
{
  "time": "2025-03-16 10:21:47 -04:00",
  "name": "travis",
  "last_synced": "just now",
  "lb_version": "0.9.20",
  "rust_triple": "aarch64.unix.macos",
  "os_info": "15.1.0",
  "lb_dir": "/Users/travis/Library/Containers/app.lockbook/Data/.lockbook",
  "server_url": "https://api.prod.lockbook.net",
  "integrity": "Ok([EmptyFile(443bc589-d6fd-4425-b587-6d6130b0d033), EmptyFile(8d74adbd-a77b-42e8-85fb-3d6200f8528a), EmptyFile(0ca2e59e-44c3-490c-a6b7-04e21cf758e4), EmptyFile(8eb3dc21-7ee9-4637-86c8-3dd19369b50f)])",
[...]

PR #3554, opened 2025-05-24
lb-rs: don't hold read lock for full debug calc

Parth:


Parth:
in service of #3463 and #3539

PR #3560, opened 2025-05-28
workspace: Space Inspector Submenu

AravDesai:
Adds a context menu to Space Inspector. Files can now be right-clicked to be deleted. Deletion immediately updates the visual of Space Inspector (flamegraph and folder sizes).


libs/content/workspace/src/space_inspector/show.rs
```diff
@@ -445,7 +450,33 @@ impl SpaceInspector {
...
+                            ui.close_menu();
+                        }
+                    }
+
+                    if ui.button("Delete").clicked() {
+                        let lb = self.lb.clone();
+                        let id = item_filerow.file.id;
+                        deleted_id = Some(id);
+                        thread::spawn(move || {
+                            let _ = lb.delete_file(&id);
```
Parth, line 476: I'd like to see this error handled before merging -- you can unwrap, it's within a thread and the application won't die. Unwrapping will at-least generate log and panic files. Or you can handle it better.

AravDesai:
Portion sizes are now only recalculated when "Refresh Root" is pressed.

Parth:
Sorry if I was unclear, I'm saying that if you think any metadata has changed, you should just recalculate and re-draw everything. This is not going to happen very often, but it's easier to implement something robust this way.

Later you can use the subscription api to be notified when there are metadata changes and redraw if anything changes anywhere.

libs/content/workspace/src/space_inspector/show.rs
```diff
@@ -445,7 +450,31 @@ impl SpaceInspector {
...
+                    ui.spacing_mut().button_padding = egui::vec2(4.0, 4.0);
+
+                    ui.label(self.data.all_files[&item.id].file.name.to_string());
+                    ui.label(&display_size);
+                    ui.separator();
+
+                    if ui.ctx().input(|i| i.key_pressed(egui::Key::Escape)) {
+                        ui.close_menu();
+                    }
+                    if item_filerow.file.is_folder() && ui.button("Make Root").clicked() {
```
Parth:
```suggestion
                    if item_filerow.file.is_folder() && ui.button("Focus File").clicked() {
```

PR #3569, opened 2025-05-31
apple: add a recalculate button for debug information

Parth:


clients/apple/Shared/Screens/DebugView.swift
```diff
@@ -51,6 +53,15 @@ struct DebugView: View {
         }
         .navigationTitle("Debug Info")
     }
+    
+    func calculateDebugInfo() {
```
smailbarkouch, line 57: Can we use this for line 43

Parth: nice, done

smailbarkouch (approved):

PR #3580, opened 2025-06-05
workspace: use subscription and status api

Parth:


Parth:
Some icons as I force sync to return various errors: 
image

Parth:
Made some further investments to messages -- status can be summarized into a single message shown as a tooltip on the sync button, which will be gray and show this message if you're offline:

image

It's nuanced, for example if I didn't have local changes:

image

but if I did:

image

errors look like this:

image

It's only shown for update req & out of space.

the red is a lil aggressive, but down to tweak that as a broader color journey that I feel we are about to go on.

During normal operation you see either:

image

or more likely 

image

I had to add a 2s sleep to even be able to see this. 

In the past we had some distracting stuff imo because the geometry kept changing, iirc it went from "syncing..." to synced and all sorts of other words. I think this is the minimum viable communicative set of state changes.

In some future world where we show everything that's dirty as a lil filled in dot (think vscode) I'm down to change the spinner to a less distracting same dot. But for now seems like icons all around fit a little better.

clients/egui/src/account/mod.rs
```diff
@@ -38,11 +41,14 @@ pub struct AccountScreen {
...
     update_rx: mpsc::Receiver<AccountUpdate>,
 
+    lb_rx: events::Receiver<Event>,
+
     tree: FileTree,
     is_new_user: bool,
     full_search_doc: FullDocSearch,
     sync: SyncPanel,
-    usage: Result<Usage, String>,
+    // todo: this will prob move into sync panel
```
Parth: this is fine here

clients/egui/src/account/mod.rs
```diff
@@ -56,7 +62,7 @@ impl AccountScreen {
         let core = core.clone();
         let (update_tx, update_rx) = mpsc::channel();
 
-        let AccountScreenInitData { sync_status, files, usage } = acct_data;
+        let AccountScreenInitData { files } = acct_data;
```
Parth: this can probably be deleted

PR #3582, opened 2025-06-09
(chore) Update Cargo.toml add repository field

szabgab:
To make it easier to find it from the standard locations More explanation: https://github.com/szabgab/rust-digger/issues/89


Parth (changes requested):
Thanks!

Although if you're just going to add this to one library it should probably be libs/lb-rs.

Parth:
Oops, I see that it has that already, I'll just merge this.

szabgab:
Thanks for accepting it.

I am sending these PRs via the web interface that, as far as I can tell, only allows me to edit one file at a time. I'll send more PRs Rust-digger reports me more crates.

Parth:
Sounds good, looking forward to seeing what else turns up. Thanks!

Issue #159, opened 2020-07-25
optional passphrase


Parth:
+ Pin
+ ~Store pin and / or entire key in HSM / Yubikey / TPM / Enclave / Keystore~

Parth:
@tvanderstad I consider #2821 #2821 duplicates of this

Parth:
in the ideal case "biometrics" store this passphrase

PR #3590, opened 2025-06-16
lb-rs: enum approach for co-operative core processes

Kousay-Jebir:
The first commits do the following big changes : 

1.  Renamed `Lb` to `InnerLb`. This has the consequence of aliasing imports of Lb in all the `service/*.rs` files 
2. Moved `InnerLb` definition and its `init` definition to its own module.
3. `lib.rs` now only imports the enum as well as the two concrete structs `InnerLb` and `ProxyLb` 
to `use crate::InnerLb as Lb`.
4. the enum itslef is now taking the name `Lb`. Added  `pub use enum_lb::Lb;` so that `lb-rs `clients don't have to adjust their imports.
5. `rpc` module contains two structs to properly define what a request and a response looks like. for now these reside in the module file itself but i think it needs to be moved to the `model` folder in future commits
6. `error.rs`'s LbErr now implements Ser and Deser, **I did ask the AI to do these implementations just demo purposes for now since it's that relevant right now for what I've been working on will probably work on improving them in later commits**
7. For now i went with the approach where each call opens its own tcp connection. thus the definition of `ProxyLb` not containing a connection field used across the whole api.
8. For drafting purposes, I hardcoded a TCP port that `init` will try to connect to, Thinking about moving this in `Config`

The goal of these first commits is to prepare the overall structure.
If this gets approved, future commits will focus on implementing the other apis, further organize the files and at some point implement robust measures such as retrial on error on connections etc...

### DIFF
--- a/libs/lb/lb-rs/src/blocking.rs
+++ b/libs/lb/lb-rs/src/blocking.rs
@@ -33,14 +33,14 @@ use crate::{
 
 #[derive(Clone)]
 pub struct Lb {
-    lb: crate::Lb,
+    lb: crate::InnerLb,
     rt: Arc<Runtime>,
 }
 
 impl Lb {
     pub fn init(config: Config) -> LbResult<Self> {
         let rt = Arc::new(Runtime::new().unwrap());
-        let lb = rt.block_on(crate::Lb::init(config))?;
+        let lb = rt.block_on(crate::InnerLb::init(config))?;
         Ok(Self { rt, lb })
     }
 
--- a/libs/lb/lb-rs/src/enum_lb.rs
+++ b/libs/lb/lb-rs/src/enum_lb.rs
@@ -1,0 +1,27 @@
+pub enum Lb {
+    ActualLb(InnerLb),
+    ExposedLb(ProxyLb)
+}
+
+impl Lb {
+    pub async fn init(config: Config) -> LbResult<Self> {
+        let socket = SocketAddrV4::new(Ipv4Addr::new(127, 0, 0, 1), 8080);
+
+        match TcpListener::bind(socket) {
+            Ok(listener) => {
+                let inner_lb = InnerLb::init(config).await?;
+                Ok(Lb::ActualLb(inner_lb))
+            }
+            Err(error) if error.kind() == std::io::ErrorKind::AddrInUse => {
+                Ok(Lb::ExposedLb(ProxyLb{addr: socket}))
+            }
+            Err(error) => Err(LbErrKind::Unexpected(format!("Failed to bind: {error}")).into())
+        }
+    }
+}
+
+use std::net::{Ipv4Addr, SocketAddrV4, TcpListener};
+use crate::Config;
+use crate::{LbErrKind, LbResult};
+use crate::inner_lb::InnerLb;
+use crate::proxy_lb::ProxyLb;
--- a/libs/lb/lb-rs/src/enum_lb.rs
+++ b/libs/lb/lb-rs/src/enum_lb.rs
@@ -21,7 +21,21 @@ impl Lb {
 }
 
 impl Lb {
-
+    pub async fn create_account(
+        &mut self,
+        username: &str,
+        api_url: &str,
+        welcome_doc: bool,
+    ) -> LbResult<Account> {
+        match self {
+            Lb::ActualLb(inner) => {
+                inner.create_account(username, api_url, welcome_doc).await
+            }
+            Lb::ExposedLb(proxy) => {
+                proxy.create_account(username, api_url, welcome_doc).await
+            }
+        }
+    }
 }
 
 use std::net::{Ipv4Addr, SocketAddrV4};
@@ -30,3 +44,4 @@ use crate::model::core_config::Config;
 use crate::{LbErrKind, LbResult};
 use crate::inner_lb::InnerLb;
 use crate::proxy_lb::ProxyLb;
+use crate::model::account::Account;
--- a/libs/lb/lb-rs/src/enum_lb.rs
+++ b/libs/lb/lb-rs/src/enum_lb.rs
@@ -7,7 +7,7 @@ impl Lb {
     pub async fn init(config: Config) -> LbResult<Self> {
         let socket = SocketAddrV4::new(Ipv4Addr::new(127, 0, 0, 1), 8080);
 
-        match TcpListener::bind(socket) {
+        match TcpListener::bind(socket).await {
             Ok(listener) => {
                 let inner_lb = InnerLb::init(config).await?;
                 Ok(Lb::ActualLb(inner_lb))
@@ -20,8 +20,13 @@ impl Lb {
     }
 }
 
-use std::net::{Ipv4Addr, SocketAddrV4, TcpListener};
-use crate::Config;
+impl Lb {
+
+}
+
+use std::net::{Ipv4Addr, SocketAddrV4};
+use tokio::net::TcpListener;    
+use crate::model::core_config::Config;
 use crate::{LbErrKind, LbResult};
 use crate::inner_lb::InnerLb;
 use crate::proxy_lb::ProxyLb;
--- a/libs/lb/lb-rs/src/lib.rs
+++ b/libs/lb/lb-rs/src/lib.rs
@@ -26,6 +26,7 @@ pub mod subscribers;
 pub mod inner_lb;
 pub mod proxy_lb;
 pub mod enum_lb;
+pub mod rpc;
 
 pub fn get_code_version() -> &'static str {
     env!("CARGO_PKG_VERSION")
@@ -34,8 +35,8 @@ pub fn get_code_version() -> &'static str {
 pub static DEFAULT_API_LOCATION: &str = "https://api.prod.lockbook.net";
 pub static CORE_CODE_VERSION: &str = env!("CARGO_PKG_VERSION");
 
-pub use uuid::Uuid;
-pub use model::errors::{LbErrKind, LbResult};
-pub use model::core_config::Config;
+
 pub use enum_lb::Lb;
 pub use inner_lb::InnerLb;
+pub use uuid::Uuid;
+pub use model::errors::{LbErrKind, LbResult};
--- a/libs/lb/lb-rs/src/lib.rs
+++ b/libs/lb/lb-rs/src/lib.rs
@@ -23,44 +23,8 @@ pub mod io;
 pub mod model;
 pub mod service;
 pub mod subscribers;
-
-#[derive(Clone)]
-pub struct Lb {
-    pub config: Config,
-    pub keychain: Keychain,
-    pub db: LbDb,
-    pub docs: AsyncDocs,
-    pub search: SearchIndex,
-    pub client: Network,
-    pub events: EventSubs,
-    pub syncing: Arc<AtomicBool>,
-    pub status: StatusUpdater,
-}
-
-impl Lb {
-    #[instrument(level = "info", skip_all, err(Debug))]
-    pub async fn init(config: Config) -> LbResult<Self> {
-        logging::init(&config)?;
-
-        let db = CoreDb::init(db_rs::Config::in_folder(&config.writeable_path))
-            .map_err(|err| LbErrKind::Unexpected(format!("{:#?}", err)))?;
-        let keychain = Keychain::from(db.account.get());
-        let db = Arc::new(RwLock::new(db));
-        let docs = AsyncDocs::from(&config);
-        let client = Network::default();
-        let search = SearchIndex::default();
-        let status = StatusUpdater::default();
-        let syncing = Arc::default();
-        let events = EventSubs::default();
-
-        let result = Self { config, keychain, db, docs, client, search, syncing, events, status };
-
-        result.setup_search();
-        result.setup_status().await?;
-
-        Ok(result)
-    }
-}
+pub mod inner_lb;
+pub use inner_lb::InnerLb;
 
 pub fn get_code_version() -> &'static str {
     env!("CARGO_PKG_VERSION")
@@ -69,19 +33,6 @@ pub fn get_code_version() -> &'static str {
 pub static DEFAULT_API_LOCATION: &str = "https://api.prod.lockbook.net";
 pub static CORE_CODE_VERSION: &str = env!("CARGO_PKG_VERSION");
 
-use crate::io::CoreDb;
-use crate::service::logging;
-use db_rs::Db;
-use io::docs::AsyncDocs;
-use io::network::Network;
-use io::LbDb;
-use model::core_config::Config;
-pub use model::errors::{LbErrKind, LbResult};
-use service::events::EventSubs;
-use service::keychain::Keychain;
-use std::sync::atomic::AtomicBool;
-use std::sync::Arc;
-use subscribers::search::SearchIndex;
-use subscribers::status::StatusUpdater;
-use tokio::sync::RwLock;
 pub use uuid::Uuid;
+pub use model::errors::{LbErrKind, LbResult};
+pub use model::core_config::Config;
--- a/libs/lb/lb-rs/src/lib.rs
+++ b/libs/lb/lb-rs/src/lib.rs
@@ -24,7 +24,22 @@ pub mod model;
 pub mod service;
 pub mod subscribers;
 pub mod inner_lb;
-pub use inner_lb::InnerLb;
+pub mod proxy_lb;
+
+pub enum Lb {
+    ActualLb(InnerLb),
+    ExposedLb(ProxyLb)
+}
+
+impl Lb {
+    pub async fn init(config: Config) -> LbResult<Self> {
+        if true { 
+            let inner_lb = InnerLb::init(config).await?;
+            return Ok(Lb::ActualLb(inner_lb));
+        }
+        Ok(Lb::ExposedLb(ProxyLb{}))
+    }
+}
 
 pub fn get_code_version() -> &'static str {
     env!("CARGO_PKG_VERSION")
@@ -33,6 +48,8 @@ pub fn get_code_version() -> &'static str {
 pub static DEFAULT_API_LOCATION: &str = "https://api.prod.lockbook.net";
 pub static CORE_CODE_VERSION: &str = env!("CARGO_PKG_VERSION");
 
+use proxy_lb::ProxyLb;
 pub use uuid::Uuid;
 pub use model::errors::{LbErrKind, LbResult};
 pub use model::core_config::Config;
+pub use inner_lb::InnerLb;
--- a/libs/lb/lb-rs/src/lib.rs
+++ b/libs/lb/lb-rs/src/lib.rs
@@ -25,21 +25,7 @@ pub mod service;
 pub mod subscribers;
 pub mod inner_lb;
 pub mod proxy_lb;
-
-pub enum Lb {
-    ActualLb(InnerLb),
-    ExposedLb(ProxyLb)
-}
-
-impl Lb {
-    pub async fn init(config: Config) -> LbResult<Self> {
-        if true { 
-            let inner_lb = InnerLb::init(config).await?;
-            return Ok(Lb::ActualLb(inner_lb));
-        }
-        Ok(Lb::ExposedLb(ProxyLb{}))
-    }
-}
+pub mod enum_lb;
 
 pub fn get_code_version() -> &'static str {
     env!("CARGO_PKG_VERSION")
@@ -48,8 +34,8 @@ pub fn get_code_version() -> &'static str {
 pub static DEFAULT_API_LOCATION: &str = "https://api.prod.lockbook.net";
 pub static CORE_CODE_VERSION: &str = env!("CARGO_PKG_VERSION");
 
-use proxy_lb::ProxyLb;
 pub use uuid::Uuid;
 pub use model::errors::{LbErrKind, LbResult};
 pub use model::core_config::Config;
+pub use enum_lb::Lb;
 pub use inner_lb::InnerLb;
--- a/libs/lb/lb-rs/src/lib.rs
+++ b/libs/lb/lb-rs/src/lib.rs
@@ -40,3 +40,4 @@ pub use enum_lb::Lb;
 pub use inner_lb::InnerLb;
 pub use uuid::Uuid;
 pub use model::errors::{LbErrKind, LbResult};
+use model::core_config::Config;
--- a/libs/lb/lb-rs/src/model/errors.rs
+++ b/libs/lb/lb-rs/src/model/errors.rs
@@ -7,7 +7,7 @@ use std::sync::PoisonError;
 
 use hmac::crypto_mac::MacError;
 use serde::ser::SerializeStruct;
-use serde::{Serialize, Serializer};
+use serde::{Serialize, Serializer,Deserialize,Deserializer,de::{self,Visitor}};
 use tracing::error;
 use uuid::Uuid;
 
@@ -22,6 +22,32 @@ pub type LbResult<T> = Result<T, LbErr>;
 pub struct LbErr {
     pub kind: LbErrKind,
     pub backtrace: Option<Backtrace>,
+}
+
+impl Serialize for LbErr {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where S: Serializer {
+        let s = format!("{:?}", self);
+        serializer.serialize_str(&s)
+    }
+}
+
+impl<'de> Deserialize<'de> for LbErr {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where D: Deserializer<'de> {
+        struct LbErrVisitor;
+        impl<'de> Visitor<'de> for LbErrVisitor {
+            type Value = LbErr;
+            fn expecting(&self, f: &mut fmt::Formatter) -> fmt::Result {
+                f.write_str("a debug‐formatted LbErr string")
+            }
+            fn visit_str<E>(self, s: &str) -> Result<LbErr, E>
+            where E: de::Error {
+                Ok(LbErrKind::Unexpected(s.to_string()).into())
+            }
+        }
+        deserializer.deserialize_str(LbErrVisitor)
+    }
 }
 
 /// Using this within core has limited meaning as the unexpected / expected error
--- a/libs/lb/lb-rs/src/rpc.rs
+++ b/libs/lb/lb-rs/src/rpc.rs
@@ -1,0 +1,55 @@
+#[derive(Serialize, Deserialize)]
+pub struct RpcRequest {
+    pub method: String,
+    pub args: Vec<u8>,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct RpcResponse<T> {
+    pub result: LbResult<T>,
+}
+
+impl RpcRequest {
+    pub fn new(method: impl Into<String>, args: Vec<u8>) -> Self {
+        Self {
+            method: method.into(),
+            args,
+        }
+    }
+}
+
+pub async fn call_rpc<T>(
+    stream: &mut TcpStream,
+    method: &str,
+    args: Vec<u8>,
+) -> LbResult<T>
+where
+    T: for<'de> Deserialize<'de>,
+{
+    let req = RpcRequest::new(method, args);
+    let encoded = bincode::serialize(&req).map_err(core_err_unexpected)?;
+    let len = encoded.len() as u32;
+
+    let mut full_msg = Vec::with_capacity(4 + encoded.len());
+    full_msg.extend_from_slice(&len.to_be_bytes());
+    full_msg.extend_from_slice(&encoded);
+
+    stream.write_all(&full_msg).await.map_err(core_err_unexpected)?;
+
+    let mut len_buf = [0u8; 4];
+    stream.read_exact(&mut len_buf).await.map_err(core_err_unexpected)?;
+    let resp_len = u32::from_be_bytes(len_buf);
+
+    let mut resp_buf = vec![0u8; resp_len as usize];
+    stream.read_exact(&mut resp_buf).await.map_err(core_err_unexpected)?;
+
+    let resp: RpcResponse<T> = bincode::deserialize(&resp_buf).map_err(core_err_unexpected)?;
+    resp.result
+}
+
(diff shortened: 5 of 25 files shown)


Kousay-Jebir:
I still haven't started working on the Leader's side of things just wanted to commit this first

libs/lb/lb-rs/src/enum_lb.rs
```diff
@@ -0,0 +1,47 @@
+pub enum Lb {
+    ActualLb(InnerLb),
+    ExposedLb(ProxyLb)
+}
```
Parth: Some thoughts on naming

```suggestion
pub enum LbConn {
    Direct(Lb),
    Network(LbClient)
}
```

Parth: This reduces the scope of your diff

Kousay-Jebir: Does this imply leaving the direct Lb as the default export instead of default exporting the enum ?

Parth:
am I correct in the following:
- the port is bound to, but nothing right now is listening for requests to do responses. This is probably important to see as far as POC goes.
- The idea of using a string as the rpc identifier and then using tuples for arguments is great, choice of bincode is also good. 
- consider adding some things that measures the overhead of forming a connection so we can determine how prohibitive it is although I think you'll just need to add an int id to correlate request and responses. probably wouldn't be hard to do.

Parth:
Doesn't need to happen now, but a handful of this behavior will ideally be configurable.

libs/lb/lb-rs/src/enum_lb.rs
Parth: I think `LbConn` is what most people will store a reference to, maybe it should be called `Lb` for that reason, I think it's useful that the word `Conn` hints at what's going on, but could be ugly. I'm also open to this:

```rust
pub enum Lb {
    Direct(LbServer),
    Network(LbClient)
}
```

Users just hang on to an `Lb`, you can have direct access to the server and field requests "Directly" or you can talk to the server over the LbClient.

That's growing on me

Kousay-Jebir:
> the port is bound to, but nothing right now is listening for requests to do responses. This is probably important to see as far as POC goes.

Yes you're right, my plan was to include that too in this commit but i spent too much time about the rest of the things that i forgot about it right when i decided to push but yeah that's my priority for the next thing right now

Parth:
```rust
pub enum Lb {
    Direct(LbServer),
    Network(LbClient)
}
```

fleshing out this idea more, the equivalent is:

```rust
pub trait Lb {
  ...
}

// this would effectively be a router, calling the actual implementations inside `LbServer`. 
impl Lb for LbServer {
  ...
}

// this is just calling directly
impl Lb for LbClient {
  ...
}
```

What people would hang on to, as a result is `Box<dyn Lb>`. If we can make `init` not-async, does the trait approach work out? 

I think this is going to be an annoying object to give someone, even in a non-async context. Possibly making `Clone`, `Debug` etc annoying to implement. 

I think we also don't need the ability to have other people implement these, so I'm down to just commit to the enum approach. And I'm down to use the naming scheme specified in this comment

Parth:
Looking forward to seeing next steps here

Kousay-Jebir:
RPC communication did work when i tested it with an api call that just returns whatever integer is passed to it in my `lb-rs-prototype repo`.

I still haven't found a better way to implement `dispatch` to cover all `Lb` methods other than a match statement on the method string name. I did ask AI about it and it proposed two approaches , one with a macro that generates the rest of the methods and the other some sort of hashmap registry that stores pointers to all lb functions.

I also got rid of RpcResponse because it served me no purpose when i completed the leader's listening and handling requests.

I think next steps is probably start thinking about the mechanism that lets a follower become a leader.

and also this
>consider adding some things that measures the overhead of forming a connection so we can determine how prohibitive it is although I think you'll just need to add an int id to correlate request and responses. probably wouldn't be hard to do.

Kousay-Jebir:
My opinion on the dispatch thing i talked about in my last comment : 
I don't like how its coupled with `lb` methods. I like to view `rpc` module to just be responsible with communication and not having to know about the methods.  for this reason i like the registry approach where you would register what methods this rpc module can handle

Parth:
code is looking great, some thoughts:
- if you explore router_service.rs in server you can see how I attempted a macro in a very similar setting. The macro I think was ultimately not worth the code it saved because of how much complexity it adds. I think macros should be used carefully and not in this circumstance.
- you can see a file like `blocking.rs` where we have to wrap basically every single core call, and it's totally fine, and it makes it clear what's going on. It may even be a good starting list for you to implement
- with regards to the registry method, think about the number of tokens or ideas you're expressing if it's roughly the same amount then your "registry" can just look like blocking.rs in the sense that, to register a new rpc call you just add it to this hardcoded list.
- similar to macros you want to be prtety hesitant to introduce layers of indirection until absolutely needed. For instance there's never going to be a situation where a RPC list is changed during runtime. It'll always require a codechange. So no point introducing a dynamic portion of this pipeline that's more complicated and possibly slower.
- imo the string based rpc addressing, having a large list of functions that we're matching on and routing is also totally fine. We'd want to keep it isolated to a file and make it clear what's going on and make it easy for anyone to add more. But we don't have too many more requirements than that.

Kousay-Jebir:
I focused on implementing most of the rest of core methods.
`event` `import_export` and `sync` contains calls with callbacks in the params. so currently im doing some research on how to tackle this. my goal is to find a solution for this, implement the rest of the api and get the whole thing to compile before moving on to other things

libs/lb/lb-rs/src/blocking.rs
```diff
@@ -33,14 +33,14 @@ use crate::{
 
 #[derive(Clone)]
 pub struct Lb {
-    lb: crate::Lb,
+    lb: crate::LbServer,
```
Parth: I think you want the `Lb` enum on here, ffi clients and others using blocking may be connecting to either.

Kousay-Jebir: ye definitely, i have that in mind

libs/lb/lb-rs/src/lb_enum/account.rs
```diff
@@ -0,0 +1,136 @@
+impl Lb {
```
Parth: imo everything that takes the shape of blocking.rs (basically a wrapper of all the lb functions) we should just stick in one file. I think the file structure here probably causes more problems than it's worth

libs/lb/lb-rs/src/model/errors.rs
```diff
@@ -24,6 +24,32 @@ pub struct LbErr {
...
 
+impl Serialize for LbErr {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where S: Serializer {
+        let s = format!("{:?}", self);
+        serializer.serialize_str(&s)
+    }
+}
+
+impl<'de> Deserialize<'de> for LbErr {
```
Parth, line 38: we probably want a `todo!()` here so we don't forget to implement this for real

libs/lb/lb-rs/src/service/admin.rs
```diff
@@ -2,7 +2,7 @@ use crate::io::network::ApiError;
 use crate::model::account::Username;
 use crate::model::api::*;
 use crate::model::errors::{core_err_unexpected, LbErrKind, LbResult};
-use crate::Lb;
+use crate::LbServer as Lb;
```
Parth: I think all of these will remain unchanged

libs/lb/lb-rs/src/dispatch.rs
```diff
@@ -0,0 +1,358 @@
+pub async fn dispatch(lb: Arc<LbServer>, req: RpcRequest) -> LbResult<Vec<u8>> {
+
+    let raw = req.args.unwrap_or_default();
+    let payload = match req.method.as_str() {
+        "create_account" => {
+            let (username, api_url, welcome): (String, String, bool) =
+                bincode::deserialize(&raw).map_err(core_err_unexpected)?;
+            let res = lb.create_account(&username, &api_url, welcome).await?;
+            bincode::serialize(&res).map_err(core_err_unexpected)?
+        }
```
Parth: while I did say don't do a macro, I think there's a fair amount of repeat that could be minimized by just using functions

Parth (changes requested):
some comments which may reduce the overall size of the diff, but looks like you're largely on the right track

libs/lb/lb-rs/src/service/admin.rs
Kousay-Jebir: Are you saying to change the enum to be named `LbServer` instead of `Lb`  and let `Lb` (the struct) named Lb as it is currently ? (i'm asking this because i don't  see how else this can remain unchanged)

Parth: Oops, sorry I reviewed too quickly, I thought this file was in the clients

Parth: I am down for this to be `impl LbServer` though

Parth: instead of the import changing, I think `impl LbServer` is very likely to be something someone searches for while trying to understand these better, finding these impl'd on `Lb` may be confusing.

libs/lb/lb-rs/src/lb_enum/account.rs
Kousay-Jebir: Does this also concern the  `lb_client` folder ?

Parth: yeah

libs/lb/lb-rs/src/service/admin.rs
Kousay-Jebir: makes sense

libs/lb/lb-rs/src/blocking.rs
Kousay-Jebir: Is it okey if i made all of `lb_client` wrappers `async` even when their corresponding `lb_server` are sync (some are sync) ?

Kousay-Jebir: (Also asking this here ) `get_account` returns `&Account`. in my wrapper i couldn't stick to this due to rpc so i made the wrapper return `Account`.  (`blocking.rs` is now giving a mismatch type because of this so i wanted to ask if its okey to also change `blocking.rs` to return `Account`

Parth: for account I would do the following:

```
    pub fn get_account(&self) -> LbResult<&Account> {
        self.keychain.get_account()
    }
```

every caller **inside** lb-rs who calls this should just call `self.keychain.get_account()` instead (`&Account`). Once this is the case you can make `get_account(&self)` return `LbResult<Account>` instead (for callers outside of lb-rs.

Parth: in most cases you can turn `sync` things `async`, but maybe do that in it's own commit, just so I can see all of them

Kousay-Jebir:
I've been playing around with expanding the rpc module to take into consideration core functions that take a callback as a param , thought a lot about how not to introduce a lot of extra complexity and ended up with this approach for now
In dispatch i'm calling the real lb and giving a callback whose job is to collect the arguments received and store them in a buffer, in rpc then it will call the actual callback passed by the client with this buffer.

Kousay-Jebir:
Also introduced some functions to help refactor dispatch and get rid of some duplication (still didn't rewrite all of the match arms waiting for a review on this) .

Parth:
I haven't looked at the recent code yet, but you can basically skip the functions with the callbacks, I view those callbacks as essentially tech debt that needs to be thought through again.

So for the important stuff like `sync` you can just pass `None`.

libs/lb/lb-rs/src/lb_client.rs
```diff
@@ -0,0 +1,620 @@
...
+
+    pub async fn get_pending_shares(&self) -> LbResult<Vec<File>>{
+        let mut stream = TcpStream::connect(&self.addr)
+            .await
+            .map_err(core_err_unexpected)?;
+        call_rpc(&mut stream, "get_pending_shares", None).await
+    }
+
+    pub async fn reject_share(&self, id: &Uuid) -> Result<(), LbErr>{
+        let mut stream = TcpStream::connect(&self.addr)
```
Parth: This identical connection can be moved to `call_rpc`, which will further reduce the amount of duplication and make it easier to do something more sophisticated later.

libs/lb/lb-rs/src/rpc.rs
```diff
@@ -0,0 +1,130 @@
...
+    let mut out = Vec::with_capacity(4 + payload.len());
+    out.extend_from_slice(&(payload.len() as u32).to_be_bytes());
+    out.extend_from_slice(&payload);
+    stream.write_all(&out).await?;
+
+    Ok(())
+}
+
+
+pub async fn listen_for_connections(lb: Arc<LbServer>, listener: TcpListener) -> LbResult<()> {
```
Parth: I think this can prob be on `impl LbServer`, use `&self` and be called upon `init` if there is some `cfg` that says it should do this.

Parth:
I think all this is heading in the right direction, feels like we're close to being able to call `list_for_connections` and try it out.

I think we'll be able to make it pretty far without worrying about impl'ing Deserialize, but looking forward to seeing what you come up with.

libs/lb/lb-rs/Cargo.toml
```diff
@@ -16,6 +16,7 @@ bench = false
 [features]
 default = []
 no-network = ["db-rs/clone"]
+serve = []
```
Parth: imo this is better of being configured by `core_config.rs` rather than `Cargo.toml`

Kousay-Jebir: i thought about that at the start then said to myself `core_config.rs`  probably should be kept to only the actual logic for core

Kousay-Jebir:
```
pub struct LbClient {
    pub addr: SocketAddrV4,
    pub events: EventSubs
}
```
do you think this is valid ? , upon inspecting i see that there is no reason for `subscribe` to be called via rpc and let the leader handle it, i think followers can just call their own without interacting with the leader here ?

libs/lb/lb-rs/src/blocking.rs
```diff
@@ -56,23 +56,38 @@ impl Lb {
...
     }
 
     pub fn export_account_qr(&self) -> LbResult<Vec<u8>> {
-        self.lb.export_account_qr()
+        self.rt.block_on(self.lb.export_account_qr())
     }
 
-    pub fn get_account(&self) -> LbResult<&Account> {
-        self.lb.get_account()
+    pub fn get_account(&self) -> LbResult<Account> {
```
Parth, line 64: All of these `match` statements should probably live in `lb_enum.rs`

libs/lb/lb-rs/src/dispatch.rs
```diff
@@ -0,0 +1,419 @@
+pub async fn dispatch(lb: Arc<LbServer>, req: RpcRequest) -> LbResult<Vec<u8>> {
+
+    let raw = req.args.unwrap_or_default();
```
Parth: I will have some comments in the client side of this that should render the `unwrap_or_default()` un-needed.

libs/lb/lb-rs/src/dispatch.rs
```diff
@@ -0,0 +1,419 @@
...
+        }
+
+        "import_account_phrase" => {
+            let (phrase_vec, api_url): (Vec<String>, String) = deserialize_args(&raw)?;
+            let slice: Vec<&str> = phrase_vec.iter().map(|s| s.as_str()).collect();
+            let phrase_arr: [&str; 24] = slice
+                .try_into()
+                .map_err(core_err_unexpected)?;
+            call_async(|| lb.import_account_phrase(phrase_arr, &api_url)).await?
+        }
```
Parth: not sure why these have been lifted out to here

libs/lb/lb-rs/src/dispatch.rs
```diff
@@ -0,0 +1,419 @@
...
+    bincode::serialize(&res).map_err(|e| core_err_unexpected(e).into())
+}
+
+fn call_sync<R>(f: impl FnOnce() -> LbResult<R>) -> LbResult<Vec<u8>>
+where
+    R: Serialize,
+{
+    let res: R = f()?;
+    bincode::serialize(&res).map_err(|e| core_err_unexpected(e).into())
+}
```
Parth, line 326: is this needed? I thought all the lb-rs fns became `async` because of the network activity

libs/lb/lb-rs/src/dispatch.rs
```diff
@@ -0,0 +1,419 @@
...
+use crate::model::errors::{LbErrKind};
+use crate::model::errors::{core_err_unexpected};
+use crate::model::file::ShareMode;
+use crate::model::file_metadata::{DocumentHmac, FileType};
+use crate::model::path_ops::Filter;
+use crate::rpc::{CallbackMessage, RpcRequest};
+use crate::service::activity::RankingWeights;
+use crate::service::import_export::{ExportFileInfo, ImportStatus};
+use crate::subscribers::search::SearchConfig;
+use crate::{LbServer,LbResult};
```
Parth: Personally I'm sympathetic to all use statements being at the bottom of all files, I like to order things in some sort of significance, either alpha or importance and use statements usually are just noise to me.

But it's too un-idiomatic and foreign of a thing to do. If `rustfmt` supported it as a config and it happened automatically I could be down to be weird. But in the absence of that I think it's better off not happening on a project we want many people to contribute to.

libs/lb/lb-rs/src/lb_enum.rs
```diff
@@ -0,0 +1,864 @@
+#[derive(Clone)]
+pub enum Lb {
+    Direct(LbServer),
+    Network(LbClient)
+}
+
+impl Lb {
+    pub async fn init(config: Config) -> LbResult<Self> {
+        if let Some(port) = config.rpc_port {
```
Parth: `clippy` (prob) and myself would like to see this sort of thing as a `match` statement.

libs/lb/lb-rs/src/lb_enum.rs
```diff
@@ -0,0 +1,864 @@
...
+            Lb::Direct(inner) => {
+                inner.status().await
+            }
+            Lb::Network(proxy) => {
+                proxy.status().await
+            }
+        }
+    }
+
+    //The follwing methods are not implemented by LbServer but exist in blocking.rs and needed elsewhere
```
Parth: down to just impl them in `LbServer`

libs/lb/lb-rs/src/lib.rs
```diff
@@ -69,19 +36,9 @@ pub fn get_code_version() -> &'static str {
 pub static DEFAULT_API_LOCATION: &str = "https://api.prod.lockbook.net";
 pub static CORE_CODE_VERSION: &str = env!("CARGO_PKG_VERSION");
 
-use crate::io::CoreDb;
```
Parth: I didn't realize I did this here, but in hindsight should not have, to not imply that the pattern should be repeated

libs/lb/lb-rs/src/rpc.rs
```diff
@@ -0,0 +1,139 @@
+#[derive(Serialize, Deserialize)]
+pub struct RpcRequest {
+    pub method: String,
+    pub args: Option<Vec<u8>>,
```
Parth: I believe you can serialize `(())` as `Vec<u8>`, I would hope it serializes to `[]` (Vec's empty representation). `Option<Vec<u8>>` here seems redundant.

libs/lb/lb-rs/src/rpc.rs
```diff
@@ -0,0 +1,139 @@
...
+    args: Option<Vec<u8>>,
+) -> LbResult<T>
+where
+    T: for<'de> Deserialize<'de>,
+{
+    let mut stream = TcpStream::connect(socket_address)
+            .await
+            .map_err(core_err_unexpected)?;
+    let req = RpcRequest::new(method, args);
+    let encoded = bincode::serialize(&req).map_err(core_err_unexpected)?;
```
Parth: Here all calls get serialized for the second time. Serialization is generally a pretty expensive operation. It's possible rust / bincode realize that they're serializing a vec (that's already serialized) and do something efficient, but I don't want to have to reason about that.

So I think `RpcRequest` should take a `<T>` of some general form which all the tuples will be pushed into.

This would also remove all the `bincode::serialize` statements from dispatch, further reducing the scope of repeat.

A lot of these wrapper files which list basically every API operation, should be as maintainable as possible. So if we can push complexity that's repeated into a function like this it's a great outcome.

libs/lb/lb-rs/src/rpc.rs
```diff
@@ -0,0 +1,139 @@
...
+    let len = encoded.len() as u32;
```
Parth: This limits the maximum request size to about 4gb, which to be fair many parts of our platform would struggle with at the moment. But I think you should not impose an additional limitation here and shoot for the full size of a `usize`.

libs/lb/lb-rs/src/rpc.rs
```diff
@@ -0,0 +1,139 @@
...
+        let msg: CallbackMessage<S, T> =
+            bincode::deserialize(&payload).map_err(core_err_unexpected)?;
+        match msg {
+            CallbackMessage::Status(s) => on_status(s),
+            CallbackMessage::Done(res) => return res,
+        }
+    }
+}
+
+pub async fn handle_connection(stream: TcpStream, lb: Arc<LbServer>) -> LbResult<()> {
```
Parth: consider impl'ing this on `LbServer` itself (`&self.handle_connection`).

libs/lb/lb-rs/src/rpc.rs
```diff
@@ -0,0 +1,139 @@
...
+
+    let mut len_buf = [0u8; 4];
+    stream.read_exact(&mut len_buf).await?;
+    let msg_len = u32::from_be_bytes(len_buf) as usize;
+
+    let mut buf = vec![0u8; msg_len];
+    stream.read_exact(&mut buf).await?;
+
+    let req: RpcRequest = bincode::deserialize(&buf).map_err(core_err_unexpected)?;
+    let payload = dispatch(lb, req).await?;
```
Parth:
```suggestion
    let resp = dispatch(lb, req).await?;
```

libs/lb/lb-rs/src/service/keychain.rs
Parth, line 1: Not sure why serialization stuff was added here, this file is prob only for LbServer internals.

libs/lb/lb-rs/src/subscribers/search.rs
```diff
@@ -27,20 +27,38 @@ pub struct SearchIndex {
     pub tantivy_reader: IndexReader,
 }
 
-#[derive(Copy, Clone, Debug)]
+impl Serialize for SearchIndex {
```
Parth, line 30: Not sure why serialization stuff was added here, this file is prob only for LbServer internals.

libs/lb/lb-rs/src/lb_enum.rs
Parth, line 1: I think you should just call this file `lb.rs`

libs/lb/lb-rs/src/lb_client.rs
```diff
@@ -0,0 +1,461 @@
+#[derive(Clone)]
+pub struct LbClient {
+    pub addr: SocketAddrV4,
+    pub events: EventSubs
```
Parth: I don't think this makes sense:
>  i think followers can just call their own without interacting with the leader here ?

let me explain the point of the subscription api:
lots of people will ask questions from `lb-rs` (like: "what are all the files?"), and then they'll show that to the user.

The subscription api tells them when that data is stale and it's time to get fresh data.

Followers simply won't have access to the information unless it's pushed from the server to them.

Followers are interested in subscribing to a `tokio` [broadcast channel](https://docs.rs/tokio/latest/tokio/sync/broadcast/index.html). I would read generally about channels and then read about broadcast if you're unfamiliar with these synchronization primitives. So you're close regarding what you have, but I don't think this field belongs in `LbClient`, I think it's going to be something like this:

With your existing stateless protocol you'd have to do something like this:

when someone calls subscribe the client implementation do the following:
- create a broadcast channel of it's own
- create a new thread, this thread will `call_rpc`, and the `dispatch` version of `subscribe` will do the actual subscription and continuously respond with updates (the channel can't be sent over the network so this is the only sensible reply here). This helper thread on the client will recv those messages form them back into rust structs and broadcast them to whoever is listening.
- the client subscribe impl will return the recv end of the broadcast channel as soon as the thread is created
- the thread will populate the channel beyond that point.

When I was advising you to not really worry about callbacks it's to reserve your resources for mastery over this one method -- subscribe. I am interested in pretty soon deprecating all the callbacks we have already in favor of this new api.

Pushing data from within lb to clients is annoying, you're experiencing it's annoyance over the network, smail experiences in ffi, etc. So we just want to do it once and then just rely on our rich typesystem to carry us beyond that point once the pipe is setup.

There are three reasons I want your connection to be stateful, and you can work around all of them:
- compatibility, if one client is updated before another one, then the connection may be incompatible, we'll need to explore conventions here, maybe one broadcasts it's min supported version (like our actual server does). Ideally this would just happen at connection time, but it's prob fast enough for it to happen over every connection. You can reserve room in the protocol like you did for the size of the transmission. Or it's part of the connection sequence (that's what I would do).
- Performance overhead -- I assume it's small but worth addressing, but you can measure and if it's insignificant then you don't need to do it.
- Security, right now there's no benefit to guarding access to localhost (to be clear it's not 0.0.0.0) but one day (#159) there likely will be and then the connection will probably take a passphrase or some other sort of API key to start. It's possible that you'd still need to *log in to* the client regardless, and that client would have a bb core with just an account or something that's used to authenticate with `LbServer`.

Parth (changes requested):
good stuff, continuing to head in the right direction, you should fire off a `cargo fmt` soon

libs/lb/lb-rs/src/dispatch.rs
Kousay-Jebir: all of lb-enum (in turn lb-client) are async due to the network activity, but i haven't turned sync methods of lb-server to async cuz no need to

libs/lb/lb-rs/src/service/keychain.rs
Kousay-Jebir: had to do this cuz now theres get_keychain

libs/lb/lb-rs/src/subscribers/search.rs
Kousay-Jebir: same here, for get_search

Kousay-Jebir:
Should i work on fixing resulting errors in `lb-c` `lb-java` and `test_utils`

libs/lb/lb-rs/src/service/keychain.rs
Parth: I am pretty sure no one outside `lb` is using keychain tho. So there prob shouldn't be a get_keychain

libs/lb/lb-rs/src/subscribers/search.rs
Parth: still I may be mis-understanding but I'm fairly sure that no one will be serializing / deserializing the whole search index -- which is basically every document stored in mem

Parth:
You can probably fix those errors last -- we can get very far with just CLI & egui as a proof of concept. Happy to see this moving again!

libs/lb/lb-rs/src/rpc.rs
```diff
@@ -0,0 +1,200 @@
...
+    send_rpc_request(&mut stream, method, &args).await?;
+    recv_rpc_response(&mut stream).await
+}
+
+pub async fn send_rpc_request<T: Serialize>(
+    stream: &mut TcpStream, method: Method, args: &T,
+) -> LbResult<()> {
+    let method_id = method as u16;
+    let body = bincode::serialize(args).map_err(core_err_unexpected)?;
+    let len = body.len();
```
Parth, line 102: I think sending a len prefix is fine and pretty common

you should just know about another approach, one that fits this 1 connection per request a bit better:

after finishing your writes you can execute `stream.Shutdown(write)` and then on the reading side you can do `read_all`. This means you don't have to negotiate the types.

This lets you do a single, probably more optimal syscall, signalling to the OS, I want to put all available bytes into memory, I assume this is the fastest way to express the idea if the end goal is in-fact to get all the bytes into memory.

Parth: however if we find that we're trying to head in the direction of re-using the connection (which is likely) then you'll need to negotiate the length. So I think keeping it in this half way point is fine until we know what direction we want to head and then produce the optimal implementation for that strategy.

libs/lb/lb-rs/src/rpc.rs
```diff
@@ -0,0 +1,200 @@
...
+
+pub async fn recv_rpc_response<R: for<'de> Deserialize<'de>>(
+    stream: &mut TcpStream,
+) -> LbResult<R> {
+    let mut len_buf = [0u8; std::mem::size_of::<usize>()];
+    stream
+        .read_exact(&mut len_buf)
+        .await
+        .map_err(core_err_unexpected)?;
+    let resp_len = usize::from_be_bytes(len_buf);
```
Parth, line 121: there is a `stream.read_u64` and a `stream.write_u64` which could simplify these and related calls

Parth: why do you use `_be_` in some places and `_le_` in others? this makes me upgrade the above to a "requested change" from a suggestion.

Kousay-Jebir:
Regarding the `SearchIndex` serialization, in `clients/cli/src/main.rs` line 241 used to have 
`lb.search.tantivy_reader.reload().unwrap();` i had to add a `get_search` method so that this can retreive the index from the leader when lb is a follower. i think this was my only reason that made me impl ser and deser on it.

Parth:
> Regarding the `SearchIndex` serialization, in `clients/cli/src/main.rs` line 241 used to have `lb.search.tantivy_reader.reload().unwrap();` i had to add a `get_search` method so that this can retreive the index from the leader when lb is a follower. i think this was my only reason that made me impl ser and deser on it.

OKay that makes sense, I think the preferred way to solve this would be to expose a new endpoint for the CLI to refresh the index.

Parth:
There are good ideas here, but this probably will need to be done from scratch due to abandonment

PR #3593, opened 2025-06-24
workspace: new tab look

ad-tra:


Parth:
rebased

Parth:
so I was able to change code to properly center these elements vertically without any hardcoded constants

image

however because of the use of different types of fonts I think this still looks strange to me:

image

perhaps this is acceptable but there are symbols that look strange:

image

and perhaps monospace was selected because it was assumed that the `*` wouldn't behave this way

Parth:
image

this is without monospace

Parth:
@tvanderstad requests that I use the jetbrains mono for vertically centered `*` characters and so it'll eventually match the editor

Parth:
indeed these bad boys are much nicer: 
image

a lil more tweaking and testing and this should be ready

Parth:
image

I removed another hack -- a manual line height setting to get the `x` to be vertically aligned and created a general process to align and render symbols (as egui doesn't seem to be able to calculate line height and vertically align them correctly):

Issue #2337, opened 2024-01-20
Remove doc from suggestions


Parth:
+ press X on a suggestion to purge it from history

Need to think through how to express "don't show again".

Three approaches come to mind:
1. try to be smart about it: if a previously dismissed notification comes up again, this time offer to dismiss forever
2. have a modifier which impacts the x. how do you find out about this feature? Maybe when you press x it changes color to advertise itself
3. have a little gear icon by suggested that has a menu where you can choose what should not show up. Possibly this is also where you could interact with pinned documents or something like that

Requested by @steverusso

Parth:
> I'd be fine even if it just resets it

tvanderstad:
Possibly unneeded after https://github.com/lockbook/lockbook/pull/2675 (should ask requestors)

Parth:
From the original top 3 request: 
> Ideally both, but at least one of: a) manually drop files from the suggested docs or b) hide suggested docs altogether.

tvanderstad:
If it's fine to do a one-time clear (similar to clearing browser history - visiting the page can make it come up again in suggestions) then this would be easy to implement. Just add an 'x' that reveals itself when hovering the suggestion and deletes all activity related to the document so that it won't make sense to suggest it anymore.

Parth:
Possibly related to workspace persistence overall

Parth:
This is done in lb-rs, lb-c, lb-java, and egui, coming to other clients soon

Parth:
will be released in 0.9.26 on android

Parth:
We haven't done this on apple yet

PR #3631, opened 2025-07-21
android: remove entries from suggested docs

ad-tra:
does the android portion of #2337

https://github.com/user-attachments/assets/701c37f7-292b-4bd1-bf1d-4180d977cc46

### DIFF
--- a/clients/android/app/src/main/java/app/lockbook/screen/FilesListFragment.kt
+++ b/clients/android/app/src/main/java/app/lockbook/screen/FilesListFragment.kt
@@ -9,6 +9,7 @@ import android.text.method.LinkMovementMethod
 import android.view.*
 import android.widget.ImageView
 import android.widget.LinearLayout
+import androidx.appcompat.widget.PopupMenu
 import androidx.core.content.ContextCompat
 import androidx.core.content.res.ResourcesCompat
 import androidx.core.view.isVisible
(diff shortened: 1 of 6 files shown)


smailbarkouch:
Can we have a quick demo 😭

clients/android/app/src/main/java/app/lockbook/screen/FilesListFragment.kt
```diff
@@ -424,16 +435,32 @@ class FilesListFragment : Fragment(), FilesFragment {
...
                     icon.setImageResource(iconResource)
+
+                    itemView.setOnLongClickListener { view ->
+                        val popup = PopupMenu(view.context, view)
+
+                        popup.menu.add(0,1,0,"Remove")
+
+                        popup.setOnMenuItemClickListener { menuItem ->
+                            Lb.clearSuggestedId(item.fileMetadata.id)
+                            model.suggestedDocs.removeAt(i)
```
smailbarkouch, line 446: rather than removing the item at the index, i would refetch suggested docs as a whole

libs/lb/lb-java/lib/src/main/java/net/lockbook/Lb.java
```diff
@@ -49,6 +49,10 @@ public class Lb {
     public static native SearchResult[] search(String input) throws LbError;
     public static native String[] suggestedDocs() throws LbError;
 
+
+    public static native void clearSuggestedId(String id) throws LbError;
```
smailbarkouch, line 53: assume the jni rust functions have already been made?

smailbarkouch (approved):
looks good! nice job

clients/android/app/src/main/java/app/lockbook/screen/FilesListFragment.kt
ad-tra: why? will there be state mismatch between the model's suggested docs and `lb`'s ?

PR #3632, opened 2025-07-21
android: refresh status with lb status every 1s

ad-tra:


smailbarkouch:
What was wrong before? (message flashing?)

ad-tra:
@smailbarkouch we were using workspace status instead of core status which makes the app appear like it's stuck in `calculating`

PR #3639, opened 2025-07-23
workspace: open relative links

### LINKED ISSUE #2700: navigate relative links to other lockbook documents
optionally make images/drawings into links to their destinations

### LINKED ISSUE #3565: relative markdown links always open browser
regardless of variation of link (with or without `../`) this link will always open a non-sensical URL in my browser. Experienced on macOS

Image

Parth:
fixes #3565 fixes #2700

### DIFF
--- a/libs/content/workspace/src/tab/mod.rs
+++ b/libs/content/workspace/src/tab/mod.rs
@@ -535,7 +535,10 @@ pub fn canonicalize_path(path: &str) -> String {
     result.to_string_lossy().to_string()
 }
 
-pub fn core_get_by_relative_path(core: &Lb, from: Uuid, path: &Path) -> Result<File, String> {
+pub fn core_get_by_relative_path<P: AsRef<Path>>(
+    core: &Lb, from: Uuid, path: P,
+) -> Result<File, String> {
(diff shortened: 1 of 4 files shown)


libs/content/workspace/src/tab/mod.rs
```diff
@@ -535,7 +535,10 @@ pub fn canonicalize_path(path: &str) -> String {
     result.to_string_lossy().to_string()
 }
 
-pub fn core_get_by_relative_path(core: &Lb, from: Uuid, path: &Path) -> Result<File, String> {
+pub fn core_get_by_relative_path<P: AsRef<Path>>(
+    core: &Lb, from: Uuid, path: P,
+) -> Result<File, String> {
+    let path = path.as_ref();
```
tvanderstad, line 541: is this validated? this is how we decide if the url is relative or not, ultimately

Parth: lb-rs will be the thing that does the final check on the resolved url, if the thing they click, after processing `../` and `/` is a valid path in lockbook, we'll open that UUID, otherwise we don't do anything and let the OS take it from there.

Issue #870, opened 2021-09-22
Be able to navigate forward and backwards through the documents you've edited


Parth:

ad-tra:
android now supports backwards and forwards through the workspace api. Workspace itself is a bit inconsistent though: when you click on an image, it will open in a new tab but when you click on the link itself, it will replace the current tab

PR #3650, opened 2025-07-24
workspace: open files in place and add tab history navigation

tvanderstad:
fixes https://github.com/lockbook/lockbook/issues/2104, on some platforms https://github.com/lockbook/lockbook/issues/870

Introduces tab history; navigate forwards and back with buttons or keyboard shortcuts:

https://github.com/user-attachments/assets/93205cfe-0e45-4931-87ce-81e0861b0733

On egui, the file tree and content search results open in the current tab. Path search results and new files open in a new tab. On native clients, the current behavior is preserved.

https://github.com/user-attachments/assets/4ef53593-7531-48ff-bee1-bada62932507

### DIFF
--- a/clients/egui/src/account/mod.rs
+++ b/clients/egui/src/account/mod.rs
@@ -24,6 +24,7 @@ use workspace_rs::theme::icons::Icon;
 use workspace_rs::widgets::Button;
 use workspace_rs::workspace::Workspace;
 
+use crate::account::tree::OpenRequest;
 use crate::settings::Settings;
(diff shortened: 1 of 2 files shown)


Parth:
I like this more than I thought i would, I would like idiomatic keybinds for non-apple prior to merging. I believe @ad-tra has some visual notes. But generally on board.

I think, also in this PR maybe we want to add nuance in more places like context menus and what not for controling this behavior more finely.

PR #3708, opened 2025-08-03
nfs3: preload fattr3 in readdirplus

Vaiz:
In the initial implementation, the iterator locks the `data` mutex on every call to `next`, which is inefficient for such a hot path.

To address this, the iterator has been split into two distinct variants:
- readdir version: works without `fattr3`
- readdirplus version: preloads `fattr3` at construction time to avoid locking during iteration.

I also got rid of manual iteration, and switched to Iterator trait instead.

### DIFF
--- a/libs/lb-fs/src/cache.rs
+++ b/libs/lb-fs/src/cache.rs
@@ -1,4 +1,4 @@
-use crate::fs_impl::Drive;
+use crate::{fs_impl::Drive, utils::file_id};
 use lb_rs::model::file::File;
 use lb_rs::model::work_unit::WorkUnit;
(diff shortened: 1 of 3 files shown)


libs/lb-fs/src/cache.rs
```diff
@@ -17,7 +17,7 @@ impl FileEntry {
         // todo this deserves some scrutiny and cross platform testing
         let mode = if file.is_folder() { 0o755 } else { 0o644 };
 
-        let fileid = file.id.as_u64_pair().0;
+        let fileid = file_id(&file);
         // intereREADDIR3resstingly a number of key read operations rely on this being correct
```
Parth:
```suggestion
        // interestingly a number of key read operations rely on this being correct
```

Issue #190, opened 2020-08-02
Represent the content of files as an event log


Parth:
Representing the contents of a file as an event log allows conflict resolution to happen in a nicer way.

This could be a breaking change and we would need infrastructure in place for modeling versions

Parth:
Must have: @steverusso 

Nice to have: @mahakanakala

Parth:
wip

<details>

# proposal - minimal & open format

in true lockbook fashion let's interoperate with some open standard around diffs / patches. 

our current pipeline is something like:
1. decrypted snapshot
2. compressed
3. encrypted

and that's what gets propagated everywhere.

and I think in the ideal world a document will now be described as some combination of snapshots & diffs.

```rust
struct Change {
  content: Vec<Diff>,
  is_snapshot: bool,
  author: PubKey,
  time: Timestamp,
} 

enum Diff {
  Add(usize, Vec<u8>),
  Rem(usize, Vec<u8>),
}

pub struct DecryptedDoc {
  changes: Vec<Change>,
}
```

this data model gives us some flexibility that we can explore:
- for every snapshot we still record who changed what to achieve a git-blame style look into the document
- a `Vec<Diff>` can describe both a current snapshot as well as an incremental evolution to the document
- based on a user's values (compactness vs history fidelity) we can convert between these formats

let's see what the encrypted versions of these documents look like.

# open questions
- can we operate on non-binary data?
- when do we know it doesn't make sense to think in terms of diffs?

</details>

Parth:
@tvanderstad's thoughts:
- canvas: it's fine
- non-binary: it's fine
- random-meta: it's fine
- binary: maybe when change is as large as snapshot?
- data model: consider [this](https://github.com/lockbook/lockbook/blob/master/libs/lb/lb-rs/src/text/operation_types.rs#L19-L23)
- metadata history should probably be independently tracked
- api "scrub through history of both"?
- extensibility thoughts (streaming, append only inside lockbook)
- generally negative thoughts about bincode's in-ability to evolve fields

Parth:
# Phase 1 - Server Migration Support
# Phase 2 - Lb Migration Support
# Phase 3 - Migration Testing roll out
- Phase 1 is tested against Phase 2 in PR form
- Phase 1 is merged and deployed 
- Phase 2 is merged and beta users test migrations on master
# Phase 4 
- Meta::V2 data model is solidified and merged
- Migration Support is generally available, and eventually updates are forced
[...]

PR #3737, opened 2025-08-14
lb-rs: LbV4 evolvable metadata

Parth:
Final step before we can make a data model for #190


libs/lb/lb-rs/src/io/mod.rs
```diff
@@ -39,6 +44,96 @@ pub struct CoreV3 {
...
+        if let Some(root) = old.root.get().copied() {
+            db.root.insert(root)?;
+        }
+        for (id, file) in old.base_metadata.get() {
+            let mut meta: SignedMeta = file.clone().into();
+            if meta.is_document() {
+                if let Some(doc) = docs.maybe_get(*id, file.document_hmac().copied()).await? {
+                    meta.timestamped_value
+                        .value
+                        .set_hmac_and_size(file.document_hmac().copied(), Some(doc.value.len()));
```
tvanderstad, line 92: 🙌🏻

PR #3764, opened 2025-08-22
android: sync file tree on all file opens

ad-tra:


clients/android/app/src/main/java/app/lockbook/model/FileModel.kt
```diff
@@ -52,15 +53,28 @@ class FileModel(
...
+
+                while (curr.id != parent.id && !curr.isRoot) {
+                    temp.add(curr)
+                    curr = Lb.getFileById(curr.parent)
+                }
+                temp.reverse()
+                fileDir.addAll(temp)
+            }catch (err: LbError){
+                println(err)
+            }
```
smailbarkouch: Some thoughts
+ We should move this into helper function
+ Try to minimize `getFileById` calls

Parth:
What problem is this solving? 

I ask because the ideal thing to do here would be to keep a list of files accessible to most parts of your codebase, and keep it up to date via the subscription api.

ad-tra:
> What problem is this solving?
> 
> I ask because the ideal thing to do here would be to keep a list of files accessible to most parts of your codebase, and keep it up to date via the subscription api.

this makes it such that the file tree corresponds to the open tab in workspace.

PR #3784, opened 2025-09-06
android: import files

ad-tra:
## loading flow
https://github.com/user-attachments/assets/258fd3a2-80aa-44b9-8cb9-bf1b5e16d818


clients/android/app/src/main/java/app/lockbook/screen/ShareReceiverActivity.kt
```diff
@@ -0,0 +1,198 @@
...
+            fragmentFinishedCallback,
+            false
+        )
+
+        binding.toolbar.setOnMenuItemClickListener {
+            activityModel.launchTransientScreen(TransientScreen.Create(model.currentParent.id))
+            CreateFileDialogFragment().show(
+                supportFragmentManager,
+                CreateFileDialogFragment.TAG
+            )
```
smailbarkouch, line 126: Might be forgetting something, but what's the `CreateFileDialogFragment` supposed to be doing here? I thought we're just trying to show the create folder modal here?

clients/android/app/src/main/java/app/lockbook/screen/ShareReceiverActivity.kt
```diff
@@ -0,0 +1,198 @@
...
+            binding.importProgress.visibility = View.INVISIBLE
+            binding.importButton.isEnabled = true
+
+            if (id !== null) {
+                startActivity(
+                    Intent(this, MainScreenActivity::class.java).apply {
+                        putExtra(IMPORTED_FILE_KEY, id)
+                    }
+                )
+            }
```
smailbarkouch, line 140: Add a comment what's going on here - hard to tell

clients/android/app/src/main/java/app/lockbook/screen/ShareReceiverActivity.kt
```diff
@@ -0,0 +1,198 @@
...
+            if (cursor.moveToFirst()) {
+                val displayNameIndex = cursor.getColumnIndex(OpenableColumns.DISPLAY_NAME)
+
+                if (displayNameIndex != -1) {
+                    fileName = cursor.getString(displayNameIndex)
+                }
+            }
+        }
+        return fileName
+    }
```
smailbarkouch, line 160: Surprised there isn't a simpler way to get an absolute file path, can we check again (I believe I do something like this in other places, exporting a file)

clients/android/app/src/main/java/app/lockbook/screen/ShareReceiverActivity.kt
```diff
@@ -0,0 +1,198 @@
...
+            } catch (err: LbError) {
+                alertModel.notifyError(err)
+            } catch (err: Exception) {
+                err.message?.let { alertModel.notify(it) }
+            }
+        }
+
+        // todo: when android supports multiple tabs. open all the imported files?
+        return newFileId
+    }
```
smailbarkouch, line 181: Lockbook api has a function for importing, we should be using that instead of creating and writing our own files

clients/android/app/src/main/java/app/lockbook/screen/ShareReceiverActivity.kt
```diff
@@ -0,0 +1,198 @@
...
+                Lb.writeDocumentBytes(lbFile.id, data)
+            } catch (err: LbError) {
+                alertModel.notifyError(err)
+            } catch (err: Exception) {
+                err.message?.let { alertModel.notify(it) }
+            }
+        }
+
+        // todo: when android supports multiple tabs. open all the imported files?
+        return newFileId
```
smailbarkouch, line 180: Android does support tabs for tablet

clients/android/app/src/main/java/app/lockbook/screen/ShareReceiverActivity.kt
```diff
@@ -0,0 +1,198 @@
...
+            if (f is CreateFileDialogFragment) {
+                model.refreshOverFolder()
+            }
+        }
+    }
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        _binding = ActivityShareReceiverBinding.inflate(layoutInflater)
+        setUpView()
```
smailbarkouch: Kinda confusing since we're also setting up the views after this lol

clients/android/app/src/main/java/app/lockbook/screen/ShareReceiverActivity.kt
ad-tra: Yeah that was my instinct too but the lb import file function accepts file paths. But android share api provides URIs which don't always map to file paths

clients/android/app/src/main/java/app/lockbook/screen/ShareReceiverActivity.kt
ad-tra: Sure but what's happening is that the file that's imported is being opened in workspace.

clients/android/app/src/main/java/app/lockbook/screen/ShareReceiverActivity.kt
smailbarkouch: Good point, on other platforms i've opted to only support URLS until lb supports it

clients/android/app/src/main/java/app/lockbook/screen/ShareReceiverActivity.kt
ad-tra: to create a folder, `createFileDialog` needs to be used. the main activity is responsible for [creating file dialogs based on the transient screen value](https://github.com/lockbook/lockbook/blob/master/clients/android/app/src/main/java/app/lockbook/screen/MainScreenActivity.kt#L137-L141). But because `ShareRecieverActivity` is a different activity, I create the file dialog manually. The feature wouldn't work otherwise.

clients/android/app/src/main/java/app/lockbook/screen/ShareReceiverActivity.kt
ad-tra: it's `content` URI which doesn't always resolve to an on disk file path. You're right that the export flow is easier, but the conversion from URI to file path is not bidirectional.

clients/android/app/src/main/java/app/lockbook/screen/ShareReceiverActivity.kt
ad-tra: I'm not sure we can do the uri resolving in rust tbh

clients/android/app/src/main/java/app/lockbook/screen/ShareReceiverActivity.kt
ad-tra: hmm that's right. I refactored this to open multiple files at once but that was a bit funky on mobile because mobile doesn't handle multiple tabs that well. For one the back button was redirecting back to the last tab instead of home screen. 
I think i'll revisit this in a separate pr where android handles tabs in the same way as ios

clients/android/app/src/main/java/app/lockbook/screen/ShareReceiverActivity.kt
ad-tra: oh you're right, ty

ad-tra: wait actually this is a pretty common pattern through the code base

clients/android/app/src/main/java/app/lockbook/screen/ShareReceiverActivity.kt
smailbarkouch: I see, why are we calling `activityModel.launchTransientScreen(TransientScreen.Create(model.currentParent.id))` then? (if its launching it on a different activity you don't care about)

clients/android/app/src/main/java/app/lockbook/screen/ShareReceiverActivity.kt
smailbarkouch: I see - yeah a quick comment would be nice (I know I don't comment a lot in the codebase, should be doing that more myself)

clients/android/app/src/main/java/app/lockbook/screen/ShareReceiverActivity.kt
smailbarkouch: Oh, makes sense

clients/android/app/src/main/java/app/lockbook/screen/ShareReceiverActivity.kt
smailbarkouch: Yeah, not by directly passing the URI, but envisioning we can handle URI conversions, and pass in the raw info to rust. That doesn't cover all cases, but seems possible with something like a remote server. All in all, my philosophy is to use lb whenever I can, and not come up with too many bespoke platform solutions to stuff like this (because we're going to have to duplicate it across platforms). But aligned with this - not too concerned

clients/android/app/src/main/java/app/lockbook/screen/ShareReceiverActivity.kt
smailbarkouch: I call stuff `setUpView` adjacent and then set up more views?

ad-tra: yeah there's an example in [MoveFileDialog](https://github.com/lockbook/lockbook/blob/9572df2ec5e81d294475d546e796b45294ecd11b/clients/android/app/src/main/java/app/lockbook/ui/MoveFileDialogFragment.kt#L56-L60)

Issue #3795, opened 2025-09-10
multitasking during sheet causes it to be lost forever


Parth:
https://github.com/user-attachments/assets/62f0fc62-d4e3-4cea-8ef2-b1ff65ed4765

Issue #3794, opened 2025-09-10
multitasking causes edit region computation to be wrong


Parth:
this happens prior to my changes as well. Starting lockbook in either mode is fine, it just doesn't survive the transition.

https://github.com/user-attachments/assets/a61c31ff-3922-4494-ae78-8abe755c9cd5

Issue #3792, opened 2025-09-10
rename bool doesn't seem to get set back to false


Parth:
https://github.com/user-attachments/assets/a31d3f71-effe-4916-98b7-a73d5f87d19a

for whatever reason changing between multitasking modes will keep re-triggering rename. This is on the released app store version, so I have to go down to compact to see the rename thing. But when the tab strip is visible this will just keep happening.

Parth:
It's not clear in the video but this behavior happens even if I swipe the rename window away, and even if I do it each time

in this video It seems like if you swipe it down, it specifically returns the next multitasking change

https://github.com/user-attachments/assets/c6eb74d3-4b50-4004-ac37-3a330c91015b

Issue #3791, opened 2025-09-10
iOS landscape looks busted


Parth:
there's the obvious visual artifact captured here:

https://github.com/user-attachments/assets/45df6153-cd99-475e-9fce-4eb5ac2ac21c

the weirdest part is that you can't even drag from the left most part of the screen. there's probably some drag sone that needs to support negative values or something.

PR #3796, opened 2025-09-10
apple: cleanup some swift

Parth:
as a follow up to the cleanup in #3793 this PR does some more cleanup:

First it eliminates the idea of `isConstrainedLayout` replacing it with it's source of truth `horizontalSizeClass`. Each instance is roughly the same amount of tokens and has a similar level of complexity, but is more generally idiomatic and easy to understand.

Each of these instances were tested and a handful of bugs were found that were not introduced by this cleanup (or the prior one):
- https://github.com/lockbook/lockbook/issues/3791
- https://github.com/lockbook/lockbook/issues/3792
- https://github.com/lockbook/lockbook/issues/3794
- https://github.com/lockbook/lockbook/issues/3795
- https://github.com/lockbook/lockbook/issues/3797

But each instance maintained prior functionality in as many situations as I could test. 

I also renamed other instances of constrained, like `constrainedSheetHeight` to `compactSheetHeight`. In my mind the word constrained, especially in the realm of design and UI indicates an arbitrary system of constraints (where a constraint may make something really large). I don't think most people will think of the more normie use of the word which means limited (like I'm time constrained). The word `compact` is smaller, less strange, and also idiomatic as it's what `horizontalSizeClass` is used to compare. These ideas are very related and anyone googling either term + swiftui will have a good time trying to understand what's going on.

I was not able to find any situation where `horizontalSizeClass` behaved strangely.

### DIFF
--- a/clients/apple/Shared/Widgets/SuggestedDocsView.swift
+++ b/clients/apple/Shared/Widgets/SuggestedDocsView.swift
@@ -24,7 +24,7 @@ struct SuggestedDocsView: View {
                         ForEach(suggestedDocs) { info in
                             Button(action: {
                                 if horizontalSizeClass == .compact {
-                                    homeState.constrainedSidebarState = .closed
+                                    homeState.compactSidebarState = .closed
                                 }
                                 
                                 AppState.workspaceState.requestOpenDoc(info.id)
--- a/clients/apple/iOS/Screens/TabsSheet.swift
+++ b/clients/apple/iOS/Screens/TabsSheet.swift
@@ -95,7 +95,7 @@ struct TabsSheet: View {
     Color.accentColor
         .optimizedSheet(
             item: $sheetInfo,
-            constrainedSheetHeight: .constant(100),
+            compactSheetHeight: .constant(100),
             presentedContent: { item in
                 TabsSheet(
                     info: item.info
--- a/clients/apple/iOS/Widgets/FileTreeView.swift
+++ b/clients/apple/iOS/Widgets/FileTreeView.swift
@@ -189,7 +189,7 @@ struct FileRowView: View {
             AppState.workspaceState.requestOpenDoc(file.id)
             
             if horizontalSizeClass == .compact {
-                homeState.constrainedSidebarState = .closed
+                homeState.compactSidebarState = .closed
             }
         }
     }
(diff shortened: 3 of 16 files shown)


clients/apple/Shared/Widgets/SuggestedDocsView.swift
```diff
@@ -23,7 +23,7 @@ struct SuggestedDocsView: View {
                     if let suggestedDocs = model.suggestedDocs {
                         ForEach(suggestedDocs) { info in
                             Button(action: {
-                                if isConstrainedLayout {
+                                if horizontalSizeClass == .compact {
```
Parth, line 26: Although this maintains existing functionality, the existing functionality is slightly wrong.

On an ipad pro, during multitasking, I have states that are non-compact and don't have the file tree fully locked out. There's no easy way at the moment to detect these states, or close the sidebar when a suggested doc is selected.

I don't think this is a huge deal, but is a code smell. I think in the situations this works in it's a nice touch, and I also added this behavior for the tab sheet (fyi).

clients/apple/iOS/Screens/TabsSheet.swift
```diff
@@ -29,6 +29,7 @@ struct TabsSheet: View {
             ForEach(info, id: \.id) { info in
                 Button(action: {
                     AppState.workspaceState.requestOpenDoc(info.id)
+                    dismiss()
```
Parth, line 32: note: new feature?

clients/apple/iOS/Widgets/FileTreeView.swift
```diff
@@ -190,7 +188,7 @@ struct FileRowView: View {
         } else {
             AppState.workspaceState.requestOpenDoc(file.id)
             
-            if isConstrainedLayout {
+            if horizontalSizeClass == .compact {
```
Parth, line 191: same comment as above applies here, although it's much more annoying here.

clients/apple/Shared/Widgets/SuggestedDocsView.swift
smailbarkouch: Not sure what you mean by "fully locked out." The reason I can't detect iPad states open and close is because they are fundamentally different components. I am using my custom drawer for the compact view, and Apple's built in one for iPad. All of that to say, I wouldn't say its a code smell lol

clients/apple/iOS/Screens/TabsSheet.swift
smailbarkouch: nice!

clients/apple/Shared/Widgets/SuggestedDocsView.swift
smailbarkouch: Agreed they should be unified in some way - I can explore that

PR #3847, opened 2025-09-25
apple: fix search crash

### LINKED ISSUE #3842: searching for any file crashes on iOS

Parth:
fixes #3842


Parth:
fwiw there was a purple warning at this place 
image

`Publishing changes from within view updates is not allowed, this will cause undefined behavior.`

PR #3853, opened 2025-09-27
apple: remodel workspace relationship and multiple tabs

smailbarkouch:


libs/content/workspace-ffi/SwiftWorkspace/Sources/Workspace/iOSMTK.swift
tvanderstad, line 1: Can you explain the changes in this file? I'm just trying to learn how it works

Parth: I don't know about `headerSize` but I can fill in some other info info (@smailbarkouch feel free to correct / augment when you have time):

we're changing swiftui's relationship with workspace. In swiftui when you subscribe to a `@Published` field your view is automatically re-drawn. Presently we have some problems:
- things that should be normal function calls are modeled as state transitions (which contributes to the next)
- views are getting redrawn erroneously (recall our double open file discovery @tvanderstad)

we're making the inputs more accessible everywhere and callable as normal ass function calls. That should explain the deletion (relocation) of `openFile()` for instance. The inputs to workspace and the outputs are separated into their own entities, which is also shown here as the deletion of `WorkspaceState` and the introduction of distinct Inputs and Outputs. Now that inputs are normal function calls, only one view will subscribe to the Input and be re-drawn.

Other discoveries worth noting:

During our pairing session we also dove deep into when we're allowed to mutate swift state and when we're not, under what circumstances we need something like `DispatchQueue.main.async` and when we don't. We found that there are two situations we need them:

+ output from ws
+ subscription api reloading data

and many other instances (most of them) are going to be removed.

The delegate business we've been talking about @tvanderstad I haven't discussed with Smail yet, just an FYI. I'm working through a Swift queue in my head.

Issue #3836, opened 2025-09-20
UI refinement for liquid glass


smailbarkouch:

Parth:
#3865

smailbarkouch:
In a good state

Parth:
agreed

PR #3865, opened 2025-10-05
apple: liquid glass

smailbarkouch:


Parth:
Does new Xcode have a formatter or something?

smailbarkouch:
I found the xcode formatter shortcut lol. Doing it selectively, but will see if I can auto format and add format checks in workflows

Parth:
Is it control p?

Parth:
I think pending shares is too important to not surface more prominently: 
image

going to see if I can fix that

Parth:
I think I'm experiencing a framework bug -- if you toggle sidebar visibility on macOS when the sidebar re-appears pending shares stays stuck in the overflow menu. `ToolbarItem(placement: .navigation)` is the only setting that doesn't behave this way, but it is ugly, so I'll leave it as it is for now. QA'ing iOS now.

Parth:
image
okay but I do think the tabs indicator is more important, The other two operations can be done from the file tree so I'm going to look at removing them from the iOS view

Parth:
image

@smailbarkouch I brought back the old tab indicator appearance, and got rid of the other two buttons like we've talked about doing in the past. Open to your thoughts about the following commit:

Parth:
starts #3836 but does not finish it

smailbarkouch:
I do think we should include the other two, even if they're hidden in an overflow menu. But aligned with keeping it like this for now.

Parth:
yeah we should do some wireframing or something soon together, the concept of sharing / exporting and the iconography is confusing to even our savvy people. Down to brainstorm together.

PR #3881, opened 2025-10-08
canvas: kinetic paning through direct touch  

ad-tra:


Parth:
this pr prob needs to be renamed

PR #3908, opened 2025-10-19
android: initial sync in foreground

ad-tra:
There's a new foreground service `InitialSync` which on start imports lb account using `lb.import()` and performs an `lb.sync()`. 
However the sync callback specified in [lib.rs](https://github.com/lockbook/lockbook/blob/5a1b5b3a0258c0668c4c2ef22ae69abafc121e6b/libs/lb/lb-java/src/lib.rs#L547-L551) which updates the sync repository with the progress can only be read in the foreground service. When the main activtiy observes the livedata of the sync progress, it receives no updates.


Parth:
I am assuming this is no longer something you're going to fix, and will likely do something new to address lazy file fixes, feel free to re-open if I'm wrong.

Issue #2078, opened 2023-09-16
lockbook import tool


Parth:
import from:
+ apple notes
+ bear

requested by:
+ brandon
+ @Luca409

Parth:
interesting project: https://github.com/microsoft/markitdown

Parth:
related #1928

paulhovey:
I would add obsidian to the list, they're a big player in the "markdown editor" world.  I would assume a number of people have vaults there (it's where I came from!) that would look to migrate.

Parth:
@paulhovey the only reason obsidian isn't here is because our format feels largely compatible with theres. All of those tools listed have sorta a proprietary format that perhaps we would want to migrate from. Were there features in obsidian you felt didn't translate well to lockbook? Those would be good criteria for an import tool.

paulhovey:
I found out during the import that obsidian has a slightly unique flavor of markdown, largely compatible with others, but some differences.  It's why I have so many metadata deleted files because I thought I fixed my markdown, only to find a set of files not rendering properly so I just wiped and re-imported until it was mostly right.  I (okay, GLM-5 with coaxing) made a little script that I used to translate my markdown files:  https://github.com/paulhovey/format-obsidian-markdown 

I don't think it's unreasonable to let people figure it out, especially since AI exists that can easily create tools, but if your markdown parser/linter already has some capability to do this it might be low-cost.

Parth:
yeah glad @tvanderstad saw that message, I think a few of those things, perhaps I would view as missing features in our product, some of which may already be addressed, rather than needs from an import tool. Highlight & Wiki style links may already be in! But understandably may not have been in when you first came to lockbook.

But probably he's the right person to forecast what we would want to migrate at time of insertion vs. support with regards to obsidian.

PR #3929, opened 2025-11-04
cli: handle spaces in filenames better

### LINKED ISSUE #2606: CLI doesn't handle spaces well
Probably the way we're invoking vim is not right and ignoring quoted arguments or something

requested by @alexwasserman

Parth:
fixes #2606


Parth:
(in service of #2078)

PR #3930, opened 2025-11-04
ws: handle spaces in paths

Parth:
I'm working on #2078 (the bear part) for @Luca409 and bear normalizes spaces. The export format has links to images with url-encoded-spaces. It looks something like this: 
image

This PR uses a tiny crate which I've already audited (though feel free to explore yourself) to decode these. It falls back on the original string if it reports any failures.

@ad-tra / @tvanderstad the ask is to QA the various image handling flows that use this link resolution logic. I checked a bunch of things but it would be good for you guys to check as well and lmk if there are any failures. Ideally you'll do this in the next couple days, but for sure before the next release.

Please & Thanks.


Parth:
I'm going to merge this to continue working on the import functionality, but I will be checking for 2 approvals here before I release

PR #3939, opened 2025-11-08
canvas: background colors

ad-tra:
https://github.com/user-attachments/assets/ac488562-852a-4132-90a7-8b82b7108177

### DIFF
--- a/clients/android/app/src/main/java/app/lockbook/screen/MainScreenActivity.kt
+++ b/clients/android/app/src/main/java/app/lockbook/screen/MainScreenActivity.kt
@@ -1,22 +1,14 @@
 package app.lockbook.screen
 
 import android.content.ClipData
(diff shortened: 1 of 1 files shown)


clients/android/app/src/main/java/app/lockbook/screen/MainScreenActivity.kt
```diff
@@ -313,3 +326,56 @@ class MainScreenActivity : AppCompatActivity() {
         finishAffinity()
     }
 }
+
+class GlobalExceptionHandler(
```
Parth: can you stick this in its own PR so we can discuss

ad-tra: yeah that's what i'm doing

PR #3940, opened 2025-11-09
ios: fix sidebar and landscape issues

### LINKED ISSUE #3791: iOS landscape looks busted

smailbarkouch:
This PR fixes the sidebar offset and landscape issues. 

fixes #3791


clients/apple/Shared/State/HomeState.swift
```diff
@@ -51,10 +51,6 @@ class HomeState: ObservableObject {
         self.workspaceOutput = workspaceOutput
         self.filesModel = filesModel
         
-        #if os(iOS)
-        expandSidebarIfNoDocs()
-        #endif
```
smailbarkouch, line 56: Logic not needed now that sidebarState is a direct function of state

PR #3961, opened 2025-11-15
apple: liquid glass icons

smailbarkouch:
Add liquid glass icons to iOS and macOS.

Removed a lot of icon files since now all I need is a `.icon` file (which uses a single SVG).

iPad:
image

macOS:
image


Parth:
looks good

Issue #3934, opened 2025-11-06
foldable sections


Parth:
requested by @Luca409 

does need to be persisted, it is in bear.

PR #3962, opened 2025-11-16
editor: heading & list folding

tvanderstad:
This PR implements Bear-style heading & list folding using invisible HTML tags that match Bear's. The state of folding in the document is persisted in the document itself, with the following pros & cons:

pros:
* fold state is persisted for no additional complexity
* fold state is invisible to markdown renderers
* copy & paste preserves state even from lockbook to bear and back (2nd video demo below)
* undo applies to fold changes in case something unexpected happens

cons:
* fold state is **not** invisible to markdown editors, shows up as inline HTML in Obsidian (screenshot below)
* affects suggested docs & activity
* incurs network overhead
* fold state is synced across users
* requires write access to a document to use

https://github.com/user-attachments/assets/bc2ef9ae-6fac-4b07-9575-2e86abd8e97e

https://github.com/user-attachments/assets/ff91d8ae-97b9-4c91-9836-8f5416965041

image

fixes: #3934

There's some kind of funkiness with the selection - when folding an element, if the cursor is below the element, the editor will scroll to the cursor.

### DIFF
--- a/libs/content/workspace/src/tab/markdown_editor/widget/mod.rs
+++ b/libs/content/workspace/src/tab/markdown_editor/widget/mod.rs
@@ -11,7 +11,7 @@ pub(crate) mod inline;
 pub(crate) mod toolbar;
 pub(crate) mod utils;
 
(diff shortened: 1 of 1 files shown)


Parth:
for your convenience `fixes: #3934`

Issue #3952, opened 2025-11-14
canvas floating islands overlap due to minimap on iPhone


tvanderstad:
Perhaps minimap should not appear here at all?

<img width="603" height="1311" alt="Image" src="https://github.com/user-attachments/assets/05f5b973-7d0c-46f4-9f1c-abd1b6da0505" />

Issue #3951, opened 2025-11-14
taps always open 'paste' menu on iPhone


tvanderstad:
https://github.com/user-attachments/assets/5f65b588-0a55-4ed8-ae3f-2b22f00ca803

tvanderstad:
seems to happen when I tap kind of anywhere, including when I tap with the pen tool, which both opens a menu and draws a small mark

tvanderstad:
fixed by #4101; issue in comment split to https://github.com/lockbook/lockbook/issues/4105

PR #3969, opened 2025-11-19
ios: input refinements

tvanderstad:
This PR addresses mainly the dead zone created on the left side of the screen by the drawer view, which intercepts all input to check if you want to drag open the file tree drawer.
- moves drawer view gesture handlers to the ZStack itself where they can be negotiated with subview gesture handlers (SwiftUI has no mechanism for negotiating gestures among ZStack siblings)
- replaces the drag handler in the drawer view with one that doesn't cancel touches
  - updates the min iOS version because this was accomplished using an API only available in iOS 18+
- ceases the delay of input events in the markdown wrapper (checkbox responsiveness)
- removes a duplicate gesture handler from our MTKView
- makes the grab zone for the drawer bigger (now that it isn't creating a dead zone)
- makes editor checkboxes slightly larger (now that this is what makes them hard to tap)
- reduces editor padding slightly (almost back to pre-fold-button level)

todo:
- [ ] opening or closing the file tree drawer duplicates input events to iOSMTK (e.g. simultaneously draws on an open canvas if the draw tool is selected)
- [x] alternating taps between a checkbox and a fold button show indications that the button is interacted with but do not register as clicks
  - this issue was a wild goose chase; took 10 hours and 10kb of notes

### DIFF
--- a/clients/apple/iOS/Widgets/DrawerView.swift
+++ b/clients/apple/iOS/Widgets/DrawerView.swift
@@ -10,6 +10,7 @@ struct DrawerView<Main: View, Side: View>: View {
     @ViewBuilder let sideView: Side
 
     @State private var sidebarOffset: CGFloat = Constants.sidebarOffsetClosed
+    @State private var mtkView: iOSMTK?
 
     var body: some View {
         GeometryReader { geometry in
@@ -237,6 +238,7 @@ struct NonCancellingDragGesture: UIGestureRecognizerRepresentable {
         let recognizer = UIPanGestureRecognizer()
         recognizer.cancelsTouchesInView = false // <- the only meaningful line in this struct
         recognizer.delegate = context.coordinator
+        recognizer.name = "DRAWER RECOGNIZER"
         return recognizer
     }
 
@@ -265,6 +267,23 @@ struct NonCancellingDragGesture: UIGestureRecognizerRepresentable {
         }
 
         @objc func handle(_ gesture: UIPanGestureRecognizer) {
+            switch gesture.state {
+            case .possible:
+                print("drawer: state -> possible")
+            case .began:
+                print("drawer: state -> began")
+            case .changed:
+                print("drawer: state -> changed")
+            case .ended:
+                print("drawer: state -> ended")
+            case .cancelled:
+                print("drawer: state -> cancelled")
+            case .failed:
+                print("drawer: state -> failed")
+            default:
+                break
+            }
+            
             guard let view = gesture.view else { return }
 
             let location = gesture.location(in: view)
@@ -300,6 +319,23 @@ struct NonCancellingDragGesture: UIGestureRecognizerRepresentable {
             default:
                 break
             }
+        }
+
+        func gestureRecognizer(
+            _ gestureRecognizer: UIGestureRecognizer,
+            shouldBeRequiredToFailBy otherGestureRecognizer: UIGestureRecognizer) -> Bool {
+            guard let view = gestureRecognizer.view,
+                  let otherView = otherGestureRecognizer.view else { return false }
+
+//            let result = false
+            let result = otherView.isDescendant(of: view)
+//            let result = gestureRecognizer.name == "DRAWER RECOGNIZER" && otherGestureRecognizer.name == "WORKSPACE GESTURE RECOGNIZER"
+                
+            if (gestureRecognizer.name ?? "").contains("RECOGNIZER") && (otherGestureRecognizer.name ?? "").contains("RECOGNIZER") {
+                print("DRAWER: \(gestureRecognizer.name ?? "nil") shouldBeRequiredToFailBy \(otherGestureRecognizer.name ?? "nil"): \(result)")
+            }
+                
+            return result
         }
     }
 }
--- a/libs/content/workspace-ffi/SwiftWorkspace/Sources/Workspace/iOSMTK.swift
+++ b/libs/content/workspace-ffi/SwiftWorkspace/Sources/Workspace/iOSMTK.swift
@@ -46,9 +46,6 @@ public class MdView: UIView, UITextInput {
     var floatingCursorNewStartY = 0.0
     var floatingCursorNewEndY = 0.0
     var autoScroll: Timer? = nil
-    
-    // touch
-    var touchRecognizer: WsTouchGestureRecognizer?
 
     init(mtkView: iOSMTK, headerSize: Double) {
         self.mtkView = mtkView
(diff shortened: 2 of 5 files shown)


clients/apple/iOS/Widgets/DrawerView.swift
```diff
@@ -216,10 +208,105 @@ struct DrawerView<Main: View, Side: View>: View {
...
+        Coordinator(
+            minimumDistance: minimumDistance,
+            onChanged: onChanged,
+            onEnded: onEnded
+        )
+    }
+
+    @MainActor func makeUIGestureRecognizer(context: Context) -> UIGestureRecognizerType {
+        let recognizer = UIPanGestureRecognizer()
+        recognizer.cancelsTouchesInView = false // <- the only meaningful line in this struct
```
smailbarkouch, line 239: I would argue we want this gesture to cancel touches in the view.

Wondering whether we are experiencing this now - someone goes to drag the sidebar open, then move vertically a bit, and the entire editor scrolls up. That is undesirable imo

libs/content/workspace-ffi/SwiftWorkspace/Sources/Workspace/iOSMTK.swift
```diff
@@ -61,14 +61,11 @@ public class MdView: UIView, UITextInput {
...
             // this supports checkboxes and other interactive markdown elements in the text area (crudely)
-            if gestureRecognizer.name == "UITextInteractionNameSingleTap"
-                || gestureRecognizer.name == "UITextInteractionNameTapAndAHalf"
-                || gestureRecognizer.name == "UITextInteractionNameLinkTap"
-            {
-                gestureRecognizer.cancelsTouchesInView = false
-            }
+            gestureRecognizer.cancelsTouchesInView = false
+            gestureRecognizer.delaysTouchesBegan = false
+            gestureRecognizer.delaysTouchesEnded = false
```
smailbarkouch, line 68: Awesome - happy to get rid of this jank

libs/content/workspace/src/tab/markdown_editor/widget/mod.rs
```diff
@@ -11,7 +11,7 @@ pub(crate) mod inline;
 pub(crate) mod toolbar;
 pub(crate) mod utils;
 
-pub const MARGIN: f32 = 45.0; // space between the editor and window border; must be large enough to accommodate bordered elements e.g. code blocks
+pub const MARGIN: f32 = 35.0; // space between the editor and window border; must be large enough to accommodate bordered elements e.g. code blocks
```
smailbarkouch, line 14: Does this affect top and bottom margin? If so - we want to adjust the swift constants accordingly (or maybe we should be deriving them from rust directly now?)

Parth:
- Things def feel better here. Measuring notes in kilobytes is wild. Also you shot up to be our most active user by a large margin.
- It feels like the dead zone related clicking is a lot better, checkboxes, folded sections, and other things feel better.
- it feels like cursor location with regards to keyboard summoning is better (did you work on that or is it placebo?)
- Something I was curious about was flicking a scroll view and tapping it to stop it still summons a keyboard, not a release blocker but something I'm still looking for. It is however less jarring because it seems like the cursor is in the right place, and I can actually dismiss the keyboard.
- The canvas stroke problem is prob a release blocker as it's going to muck up a bunch of my docs. Update: it just pans my docs, so prob not a release blocker, but still annoying.
- speaking of which let me now try on iPad
- seems like the gestures are still messed up regarding pan + zoom combinations, but it looks like that's the case on master as well so not a regression.
- unsure if it's a regression, but on iPad mini, in portrait, with the sidebar hidden, I'm not able to re-summon by pulling from the edge of the screen over the canvas portion of the screen. 
- pretty sure a regression but squeeze controls in canvas seem to be gone. Update: a regression, but not caused by this PR, caused by the prior cleanup.

tvanderstad:
> * it feels like cursor location with regards to keyboard summoning is better (did you work on that or is it placebo?)
> 
> * Something I was curious about was flicking a scroll view and tapping it to stop it still summons a keyboard, not a release blocker but something I'm still looking for. It is however less jarring because it seems like the cursor is in the right place, and I can actually dismiss the keyboard.

Any keyboard summoning related improvements are placebo, sorry!

tvanderstad:
> * pretty sure a regression but squeeze controls in canvas seem to be gone. Update: a regression, but not caused by this PR, caused by the prior cleanup.

Just fixed these in #3976

tvanderstad:
> * The canvas stroke problem is prob a release blocker as it's going to muck up a bunch of my docs. Update: it just pans my docs, so prob not a release blocker, but still annoying.

I did experience it actually drawing in my docs on iPhone and have a recording to prove it. Open to it still not being a release blocker since [some](https://github.com/lockbook/lockbook/issues/3952) [issues](https://github.com/lockbook/lockbook/issues/3951) make canvas a bad time on iPhone right now anyway and I understand it's not highly prioritized. Resolving this is my next step either way.

tvanderstad:
> * seems like the gestures are still messed up regarding pan + zoom combinations, but it looks like that's the case on master as well so not a regression.

That's right, aiming to fix soon, possibly before keyboard summoning excellence if it looks like an easy fix

Parth:
I think the only thing that would compete with that is the documents that for some reason you can't interact with. It could also be something @smailbarkouch takes a look at, if you determine that that's more appropriate.

tvanderstad:
Just discovered that this PR also fixes an issue where the dead zone created on the left side of the screen by the drawer view actually covers the left half of the button that opens the drawer, making the button unreliable

tvanderstad:
update: simultaneous pan and pinch fixed in https://github.com/lockbook/lockbook/pull/3981

clients/apple/iOS/Widgets/DrawerView.swift
tvanderstad: It's true, we are duplicating the inputs now to the drawer gesture and to workspace, and you can scroll the editor (or draw on a canvas) while opening the drawer. I'll fix this before merging and solicit another review from you

libs/content/workspace/src/tab/markdown_editor/widget/mod.rs
tvanderstad: Good catch!

tvanderstad:
closing in favor of https://github.com/lockbook/lockbook/pull/4016, https://github.com/lockbook/lockbook/pull/4017, https://github.com/lockbook/lockbook/pull/4018, https://github.com/lockbook/lockbook/pull/4019, https://github.com/lockbook/lockbook/pull/4020

PR #3971, opened 2025-11-23
apple: fixed toolbar access for zoomed in displays

smailbarkouch:
Fixes inaccessible overflow menu in sidebar when in zoomed mode.

This was caused by a padding view for the toolbar items - we should leave padding up to the system. Additionally, removing this padding control got rid of the badge on pending shares. Labels and toolbars are generally avoidant of complex icons. So they flattened the icon into just the system image.

Side-note, we have a lot of toolbar items. We probably should put everything into an overflow menu by default (except "edit" and sidebar close button). 

On normal displays:

On zoomed displays (overflow now accessible):


Parth:
good simplification, pending shares will be sicker soon

PR #3980, opened 2025-11-25
egui: continue visual refinements

Parth:
@ad-tra refined the UI in https://github.com/lockbook/lockbook/pull/3948 which was well received by all. 

However due to all the copy and pasting we're doing in the `filetree.rs` implementation, the visual changes didn't apply to suggested docs. If you look in that screenshot it looks like adam's folder is a parent of suggested docs. Suggested docs' icon also doesn't open and close. Also the documents inside suggested docs aren't aligned nicely, and they have extensions when they shouldn't.

This PR consolidates the code that draws a file cell so that it can be drawn with a uniform appearance.

I need a reliable visual abstraction for showing file tree like things for #3511 

before: 
image
image

after: 
image
image

### DIFF
--- a/clients/egui/src/account/tree.rs
+++ b/clients/egui/src/account/tree.rs
@@ -893,7 +893,7 @@ impl FileTree {
 
         resp
             // show suggested docs
-            .union(ui.vertical(|ui| self.show_suggested(ui, toasts)).inner)
+            .union(ui.vertical(|ui| self.show_suggested(ui)).inner)
             // show file tree
             .union({
                 ui.vertical(|ui| {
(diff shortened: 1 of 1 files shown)


libs/content/workspace/Cargo.toml
```diff
@@ -18,7 +18,7 @@ epaint = "0.28.1"
 glam = "0.22.0"
 image = "0.24"
 lb-fonts = { git = "https://github.com/lockbook/lb-fonts" }
-hayro = "0.4.0"
+hayro = "0.3.0"
```
tvanderstad, line 21: @Parth was this intentional?

Parth: I fixed in a subsequent commit a couple minutes later

PR #3983, opened 2025-11-26
ws: add context menu to the tab strip

tvanderstad:
https://github.com/user-attachments/assets/4b208ef1-475e-47ef-87a8-d0d1ef1a9e33


Parth:
I like it, but this doesn't fix that issue, maybe that one should be clarified that she was using the renaming part of the tabstrip and struggling to right click to paste.

But I do think this is good, other context menu items that I see here are generally `Close Other Tabs`, `Close Tabs to the Right`, etc

PR #3986, opened 2025-11-27
editor: PageUp/PageDown advance by 50 lines

raayan:
Using editor on linux noticed there was no PageUp/Down support. 50 is just a random number, I'm sure it could be chosen better.


tvanderstad (approved):
Looks like a value-add to me!

PR #4006, opened 2025-12-05
canvas: read only mode

### LINKED ISSUE #3997: read only mode

ad-tra:
This pr adds a flag `read_only` in the `Tab` struct so that different content types can handle read mode differently.  

@tvanderstad feel free to ad a MD read only mode based on that. 

on the canvas front: read only mode will: 
1. hide the tools island so that the user can't select a pen, eraser, or any other mutation tools
2. never saves the buffer even on a pan 
3. skips processing mutation events like new strokes, clipboard handling, etc. 

image

fixes #3997

### DIFF
--- a/libs/content/workspace/src/tab/svg_editor/mod.rs
+++ b/libs/content/workspace/src/tab/svg_editor/mod.rs
@@ -307,19 +307,19 @@ impl SVGEditor {
             return;
         }
 
-        // if !self.read_only {
-        if ui.input_mut(|r| {
-            r.consume_key(egui::Modifiers::COMMAND.plus(egui::Modifiers::SHIFT), egui::Key::Z)
-        }) {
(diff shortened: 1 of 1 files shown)


tvanderstad (approved):
lgtm

libs/content/workspace/src/workspace.rs
```diff
@@ -404,13 +408,27 @@ impl Workspace {
...
                         };
 
-                        let ext = match self.core.get_file_by_id(id) {
-                            Ok(file) => file
-                                .name
-                                .split('.')
-                                .next_back()
-                                .unwrap_or_default()
-                                .to_owned(),
+                        let (ext, read_only) = match self.core.get_file_by_id(id) {
```
Parth, line 411: this should use the file_cache

PR #4008, opened 2025-12-07
editor: remove new file suggested rename

tvanderstad:
fixes CI but breaks android build (removes now-unused FFI fn parameter)


tvanderstad:
@ad-tra I'm merging this to fix CI, can you un-break android by removing the unused params?

PR #4021, opened 2025-12-11
editor: add ==highlights== and :shortcodes:

tvanderstad:
* updates comrak to 0.49
* adds <mark>==highlights==</mark>
* adds `:shortcodes:` (though we're missing glyphs for many emojis - they show up as tofu for now)

https://github.com/user-attachments/assets/4a297c82-69e8-48c9-a66d-06f62e901e3f


Parth:
can we merge this

PR #4023, opened 2025-12-12
workspace: new landing page

tvanderstad:
https://github.com/user-attachments/assets/f506d3e7-6d18-4b4d-805e-601d3da95cc9

merge blockers:
* [x] create button
* [x] app performance / usage as billing usage
* [x] macos not changing ws focused parent

ideas generated:
* show folder-scoped mind map (and space inspector?)
* github-style heatmap? stocks style [heatmap](https://x.com/StockMKTNewz/status/1988268029126295557/photo/1)?
* show folders toggle
* varied greetings
* what's new
* make it a tab
* filter presets for recents, etc

### DIFF
--- a/libs/content/workspace/src/landing.rs
+++ b/libs/content/workspace/src/landing.rs
@@ -37,7 +37,7 @@ impl Default for LandingPage {
             collaborators: Default::default(),
             only_me: Default::default(),
             sort: Default::default(),
-            sort_asc: Default::default(),
+            sort_asc: true,
             flatten_tree: true,
         }
     }


libs/content/workspace/src/show.rs
```diff
@@ -124,344 +147,1171 @@ impl Workspace {
...
+        let mut create_note = false;
+        let mut create_drawing = false;
+        let mut create_folder = false;
+
+        ui.vertical_centered_justified(|ui| {
+            Frame::canvas(ui.style())
+                .inner_margin(MARGIN)
+                .stroke(Stroke::NONE)
+                .fill(Color32::TRANSPARENT)
+                .show(ui, |ui| {
```
Parth: Some basic code organization would go a long way here

Parth:
1. First of all this looks so beautiful, I plan to do a post to `/r/linux` like we did with `/r/macapps` and I'd def like this available everywhere before we do this. Also Amumu's work around flatpak would be a nice addition before that post.
2. the default sort gives me the least relelvant documents 
image should probably sort the other way for modified
3. looks like kwe're interacting with lb-rs for root & get file by id every frame, leading to an explosion in log volume
4. would be nice to sort by collaborators as well seems to kinda just not work but the other columns are sortable
5. this is sick

Parth:
number three is the only merge blocker, the rest I would consider release blockers

libs/content/workspace/src/landing.rs
```diff
@@ -0,0 +1,1010 @@
...
+
+        response
+    }
+
+    /// Create button, filter text box, show folders toggle, file types selector, collaborators selector
+    fn show_filters(&mut self, ui: &mut egui::Ui) -> Response {
+        let mut response = Response::default();
+
+        let (Some(files), Some(account)) = (&self.files, &self.account) else {
+            ui.ctx().request_repaint_after(Duration::from_millis(8));
```
Parth, line 194: 8ms?

libs/content/workspace/src/landing.rs
```diff
@@ -0,0 +1,1010 @@
...
+                    },
+                );
+            });
+        });
+
+        response
+    }
+
+    /// Files table with columns that you click to select a sort key
+    fn show_files(&mut self, ui: &mut egui::Ui) -> Response {
```
Parth: is there no further structure that can be imposed upon this function?

libs/content/workspace/src/landing.rs
tvanderstad: arbitrarily poll interval for when files become available

PR #4059, opened 2025-12-26
egui: make pdfs faster

Parth:


Parth:
nothing of value on this branch but will re-do this right now

PR #4062, opened 2025-12-27
lb-rs: fix duplicate search results

### LINKED ISSUE #4061: duplicate search results after file updates or deletions
Likely the weirdness that some people (@raayan ?) has been experiencing

smailbarkouch:
Did a few things
1. Added `id_str_raw` with `STRING` type - will be used to remove deleted docs/ids
2. Updated the `Event::MetadataChanged` callback to pass the deleted ids to the `delete` field of `update_tantivy`
3. Updated `build_index` to compute deleted

fixes #4061

### DIFF
--- a/libs/lb/lb-rs/src/subscribers/search.rs
+++ b/libs/lb/lb-rs/src/subscribers/search.rs
@@ -9,7 +9,7 @@ use std::sync::Arc;
 use std::sync::atomic::AtomicBool;
 use tantivy::collector::TopDocs;
 use tantivy::query::QueryParser;
-use tantivy::schema::{STORED, STRING, Schema, TEXT, Value};
+use tantivy::schema::{INDEXED, STORED, Schema, TEXT, Value};
 use tantivy::snippet::SnippetGenerator;
 use tantivy::{Index, IndexReader, IndexWriter, ReloadPolicy, TantivyDocument, Term, doc};
 use tokio::sync::RwLock;
@@ -214,11 +214,10 @@ impl Lb {
         let schema = self.search.tantivy_index.schema();
         let id_field = schema.get_field("id").unwrap();
         let id_str = schema.get_field("id_str").unwrap();
-        let id_str_raw = schema.get_field("id_str_raw").unwrap();
         let content = schema.get_field("content").unwrap();
 
         for id in delete {
-            let term = Term::from_field_text(id_str_raw, &id.to_string());
+            let term = Term::from_field_bytes(id_field, id.as_bytes());
             index_writer.delete_term(term);
         }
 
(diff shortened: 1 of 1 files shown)


Parth:
> Updated id_str from TEXT to STRING - did this because we don't want to tokenize it

I do want to be able to search via ID

libs/lb/lb-rs/src/subscribers/search.rs
```diff
@@ -189,7 +196,7 @@ impl Lb {
                                 let current_index = lb.search.metadata_index.read().await.clone();
                                 let deleted_ids = replacement_index.compute_deleted(&current_index);
                                 *lb.search.metadata_index.write().await = replacement_index;
-                                lb.update_tantivy(vec![], deleted_ids).await;
+                                lb.update_tantivy(deleted_ids, vec![]).await;
```
Parth, line 199: this is a crazy mistake for me to have made, nice catch

Parth:
> Updated build_index to compute deleted
I don't know if this change is totally needed just because this only happens on startup. There are some obscure situations where we call `build_index` like the CLI, but given the nature of how that gets called it's not needed there either. 

I would remove, but I would not mind if you do not remove, could go either way.

Parth (changes requested):
nice, https://github.com/lockbook/lockbook/pull/4062#issuecomment-3694884411 is the only thing I care about

smailbarkouch:
> > Updated build_index to compute deleted
> > I don't know if this change is totally needed just because this only happens on startup. There are some obscure situations where we call `build_index` like the CLI, but given the nature of how that gets called it's not needed there either.
> 
> I would remove, but I would not mind if you do not remove, could go either way.

Agreed not totally needed - leaning towards leaving it since I would assume `build_index` removes deleted ids

smailbarkouch:
> > Updated id_str from TEXT to STRING - did this because we don't want to tokenize it
> 
> I do want to be able to search via ID

Using `TEXT` makes it impossible (?) to remove the term by ID, since `delete_term` performs an ID lookup and no ID exists as a standalone entry

Parth:
The old code has this:

```rust
        for id in delete {
            let term = Term::from_field_text(id_str, &id.to_string());
            index_writer.delete_term(term);
        }
```

so does this not delete the whole entry? It just clears out that field?

I think having both `id` as text & string is prob fine, it's small. But yeah checkout the docs.

smailbarkouch:
> so does this not delete the whole entry? It just clears out that field?

It doesn't delete anything since id_str is tokenized and it fails to find it

Parth:
I see, unfortunate API design, sounds good best of luck

PR #4081, opened 2026-01-03
android: fix flicker caused by soft keyboard inset 

ad-tra:


Parth:
we did this elsewhere

Issue #3977, opened 2025-11-25
add support for hindi


Parth:
requested by Chetna

related to #2293

Parth:
@tvanderstad hindi is normal left -> right language. As far as I can tell all that would be required is for us to add a font that supports the glyphs. Is this correct? Would probably earn us an immediate premium subscription. Also a lot of people from india try our app generally and is also the largest population on earth lol.

Parth:
may be blocked on this: https://github.com/emilk/egui/issues/2517

tvanderstad:
[egui-glyphon.zip](https://github.com/user-attachments/files/23755395/egui-glyphon.zip)

<img width="2056" height="1290" alt="Image" src="https://github.com/user-attachments/assets/1cd9c0d2-c6f2-4666-914d-3babc6fb9734" />

I wrote a test app using egui + glyphon (cosmic text wrapper) that renders emojis and also hindi (with some mistakes)

Parth:
I think it does a bit better than egui is doing by just adding a devanagari font in: 

<img width="1854" height="977" alt="Image" src="https://github.com/user-attachments/assets/27e9a81b-98f4-4062-8248-4f78a1023b2d" />

but looks like this is going to be a bit sophisticated to support. I'll also see if text rendering in general can get some sort of ffmpeg style support. Maybe there's some based C lib that we could be using that's the standard and does a better job.

Parth:
Looking at cosmic's readme and specifically showing accurate and sophisticated hindi, arabic and chinese means maybe something is just wrong with our code. I'll take another look next time I'm down to invest.

Parth:
gathering more data, I took a look at [glyphon directly](https://github.com/grovesNL/glyphon/blob/main/examples/text-sizes.rs) and things seem slightly better, but still a little bit off 

<img width="1512" height="982" alt="Image" src="https://github.com/user-attachments/assets/32e554f3-a3d7-498b-9bb9-864a6164cb75" />

I'll start reading through some code now

Parth:
[glyphon.zip](https://github.com/user-attachments/files/24019027/glyphon.zip)

@tvanderstad try this -- it hard codes a font for macOS

Parth:
<img width="1051" height="891" alt="Image" src="https://github.com/user-attachments/assets/31cb06d5-29cd-4511-a07c-acbd5dc059a7" />

[...]

Issue #1788, opened 2023-06-04
emoji and other characters support


Parth:
[⌥ + ⌦](https://stackoverflow.com/questions/76386013/behaves-strangely-in-a-custom-uitextinput-implementation)

and emojis don't work

MacOS probably needs to support the emoji picker as well.

ad-tra:
requested by @mahakanakala

Parth:
also requested by @KrishmaBawa

tvanderstad:
From egui's [docs](https://docs.rs/egui/latest/egui/special_emojis/index.html):
> NOTE: In egui all emojis are monochrome!

egui has an [issue](https://github.com/emilk/egui/issues/2551) open since Jan 2023 for color emoji support. Some discussion participants have tried and failed to get it working. Others are indicating they're willing to fund the effort.

In either case I don't see this being a quick win

tvanderstad:
fixed by https://github.com/lockbook/lockbook/pull/4251

PR #4083, opened 2026-01-03
egui: 0.28 -> 0.30

Parth:
last pre-req for cosmic text:
- #3977 
- #1788


Parth:
Okay I did a bunch of the grunt work involved and the last couple things involve changes to canvas and editor that I want to delegate to @tvanderstad & @ad-tra respectively. You guys just have to get your directories to compile, then I can have a look at QA.

- I recommend glancing at the diff to get a sense of some of the things deprecated -- there's a bunch of old UI functions that have been replaced by a single ui function that uses an instance of a `UiBuilder::new()` api.
- and then just make `cargo check` succeed on your area of the codebase

I imagine once that's done I'll have to fix the rest of the errors in the egui app -- which won't be surfaced until workspace is taken care of. I can do these things, but if you're feeling charitable feel free to do these as well.

libs/content/workspace/src/show.rs
```diff
@@ -1067,12 +1072,13 @@ impl Workspace {
...
-                                    LayerId::new(
-                                        Order::PanelResizeLine,
+                                ui.scope_builder(
+                                    UiBuilder::new().layer_id(LayerId::new(
+                                        // This used to be PanelResizeLine, but mr. milk says that
+                                        // is an unused hack: https://github.com/emilk/egui/pull/5455 this is just a guess at a new value
+                                        Order::Middle,
                                         Id::from("tab_reorder_drop_indicator"),
-                                    ),
+                                    )),
```
tvanderstad, line 1081: if the tab reorder drop indicator still shows then I'd say we're good - I don't think there's even a way to get something to be drawn over it in our app

Parth:
(master) egui 0.28

<details> 

image

image

image

</details> 

egui 0.30

<details>
image

image

image

</details>

Parth:
unfortunately egui seems to have changed the colors defined in styles, it may be time for our colors & settings excellence journey now

tvanderstad:
closing in favor of https://github.com/lockbook/lockbook/pull/4216

PR #4089, opened 2026-01-08
android, iOS: elegant keyboard re-sizing

Parth:


libs/content/egui_wgpu_renderer/src/lib.rs
```diff
@@ -12,6 +12,7 @@ pub struct RendererState<'w> {
     pub context: egui::Context,
     pub raw_input: egui::RawInput,
     pub screen: ScreenDescriptor,
+    pub bottom_inset: Option<u32>,
```
Parth, line 15: @ad-tra this would ideally live in `WgpuWorkspace`. I think it doesn't need to be an `Option`, it's default value is probably fine for it's non-existent state. Probably leave a lil comment explaining the overall strategy right here.

Not sure if it should be before or after the screen scaling math.

libs/content/workspace-ffi/src/lib.rs
```diff
@@ -22,8 +22,19 @@ impl WgpuWorkspace<'_> {
...
         }
+
+        let workspace_frame = egui::Frame::default()
+            .fill(self.renderer.context.style().visuals.panel_fill)
+            .inner_margin(egui::Margin {
+                left: 0.0,
+                right: 0.0,
+                top: 0.0,
+                bottom: self.renderer.bottom_inset.unwrap_or(0) as f32
+                    / self.renderer.screen.pixels_per_point,
```
Parth, line 33: I don't think this should happen here -- i think this should happen at integration time setting the inset. Apple reports these already in the right coordinate space, so I have to scale them up - only for this guy to scale them back down feels wrong.

ad-tra: Fyi Android doesn't report them in the right scale

Parth: yeah I figured looking at the code, so Ideally that adjustment happens in an android specific location

Issue #3069, opened 2024-11-02
Remember cursor position


BuriedInTheGround:
I propose to add a feature that remembers the cursor position, so the next time you open a document, you'll find the cursor in the last position you left it. This may also be an optional feature, with a toggle in settings.

Parth:
Another one for workspace "settings": #2936

PR #4090, opened 2026-01-09
md: persist scroll & cursor

tvanderstad:
fixes https://github.com/lockbook/lockbook/issues/3069

works for the active tab only

### DIFF
--- a/libs/content/workspace/src/tab/markdown_editor/mod.rs
+++ b/libs/content/workspace/src/tab/markdown_editor/mod.rs
@@ -453,9 +453,19 @@ impl Editor {
                             d.insert_temp(scroll_area_id, state);
                         }
                     });
-                    self.event
-                        .internal_events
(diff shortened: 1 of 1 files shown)


Parth:
Just give this a quick test with multiple windows on macOS

tvanderstad:
> Just give this a quick test with multiple windows on macOS

It is a little weird because the scroll offset will put you in a different place depending on the window size. With one window that's fine because the size is remembered. With two windows the second one can persist a value that doesn't mean the same thing, then when you close and reopen the app, it opens in a different size and you are scrolled to a seemingly-random place.

Think this is value-add enough to merge?

Parth:
yeah I was mostly concerned about multiple windows interfering with one another making editing impossible or something like that. sounds like that's not the case

tvanderstad:
yeah the cursor + scroll position are set on load and saved on change

PR #4108, opened 2026-01-15
lb-rs: size-based request timeouts

### LINKED ISSUE #3463: macOS stuck unable to sync with no error message (fixed by restarting)

tvanderstad:
Fixes #3463 (or at least tries to)
* Adds a generous minimum timeout for all network requests (15s)
* Adds a generous size-based timeout for document push/pull (10KB/s)
* Populates missing fields on `File` so that `lockbook debug info` reports file size

### DIFF
--- a/libs/lb/lb-rs/src/io/network.rs
+++ b/libs/lb/lb-rs/src/io/network.rs
@@ -10,6 +10,14 @@ use crate::model::clock::{Timestamp, get_time};
 use crate::model::errors::LbErr;
 use crate::model::pubkey;
 
+// intended for small requests like metadata transfers and small file uploads
+const MINIMUM_TIMEOUT: Duration = Duration::from_secs(15);
+
+// network below this many bits per second is considered offline
+// expressed in time per byte (strange but useful for units/type safety)
+// this is on a per-request basis; consider parallelism when adjusting
+const MINIMUM_BITRATE: Duration = Duration::from_micros(100); // 100 micros per byte = 10KB per second
+
 impl<E> From<ErrorWrapper<E>> for ApiError<E> {
     fn from(err: ErrorWrapper<E>) -> Self {
         match err {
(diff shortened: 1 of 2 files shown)


libs/lb/lb-rs/src/io/network.rs
```diff
@@ -10,6 +10,14 @@ use crate::model::clock::{Timestamp, get_time};
 use crate::model::errors::LbErr;
 use crate::model::pubkey;
 
+// intended for small requests like metadata transfers and small file uploads
```
Parth, line 13:
```suggestion
/// intended for small requests like metadata transfers and small file uploads
```

and to the one below

libs/lb/lb-rs/src/io/network.rs
```diff
@@ -54,9 +62,30 @@ impl Default for Network {
...
+    ) -> Result<T::Response, ApiError<T::Error>> {
+        self.request_with_timeout(account, request, Duration::ZERO)
+            .await
+    }
+
+    /// Makes a network request with a timeout based on an expected payload
+    /// size. Timeout will never be less than the default/minimum timeout; it is
+    /// safe to pass a zero size.
+    pub async fn request_with_size<T: Request>(
+        &self, account: &Account, request: T, size: usize,
```
Parth, line 77: any reason we're not just using the serialized size?

libs/lb/lb-rs/src/model/file.rs
```diff
@@ -44,9 +44,10 @@ pub struct File {
     pub name: String,
     pub file_type: FileType,
     pub last_modified: u64,
-    /// Note: not yet populated
     pub last_modified_by: Username,
+    pub owner: Username,
     pub shares: Vec<Share>,
+    pub size_bytes: u64,
```
Parth, line 50: good additions

libs/lb/lb-rs/src/service/sync.rs
```diff
@@ -306,7 +306,13 @@ impl Lb {
...
 
             if let Some(remote_hmac) = remote_hmac {
-                docs_to_pull.push((id, remote_hmac));
+                let size = remote
+                    .find(&id)?
+                    .timestamped_value
+                    .value
+                    .doc_size()
+                    .unwrap_or_default();
+                docs_to_pull.push((id, remote_hmac, size));
```
Parth, line 315: I see why it's not based on the serialized size now

libs/lb/lb-rs/src/model/file.rs
Parth: I think `size_bytes` calculation needs to match the one in server though, it needs to take into account the metadata fee as well. There's a tiny function somewhere that should be trivial to move around

libs/lb/lb-rs/src/io/network.rs
Parth: never mind, I see why below

Parth:
I think this is a good move, I would be down to let you just dog food this and tell us if it actually fixes it before merging though

libs/lb/lb-rs/src/io/network.rs
tvanderstad: this approach also works for pulling files

libs/lb/lb-rs/src/service/sync.rs
tvanderstad: nice

Parth:
I'm curious to hear how this is been working out for you, I don't experience this as much as you, but I just experienced this recently while using my iPad, and I toggled on airplane mode which I imagine would kill requests but the app never recovered despite airplane mode.

Parth:
Our investigation detailed in the above issue determined that this is likely not a network timeout related issue

Issue #3867, opened 2025-10-06
receive crash reports from beta users


Parth:
when we sync, if we're syncing a beta user, send the crash reports to the server. 

expose a way for an admin to fetch the latest ones and also clear them.

Parth:
<img width="632" height="307" alt="Image" src="https://github.com/user-attachments/assets/006f5357-de6f-4737-962a-8adaa7c3f4f5" />

PR #4114, opened 2026-01-18
all: add crash reporting for beta users

ad-tra:
fixes https://github.com/lockbook/lockbook/issues/3867

### DIFF
--- a/libs/lb/lb-rs/src/service/debug.rs
+++ b/libs/lb/lb-rs/src/service/debug.rs
@@ -58,7 +58,7 @@ impl Lb {
         }
     }
 
-    async fn lb_id(&self) -> LbID {
+    async fn lb_id(&self) -> LbResult<LbID> {
         let mut tx = self.begin_tx().await;
         let db = tx.db();
 
@@ -66,13 +66,13 @@ impl Lb {
             id
         } else {
             let new_id = LbID::generate();
-            db.id.insert(new_id);
+            db.id.insert(new_id)?;
             new_id
         };
 
         tx.end();
 
-        lb_id
+        Ok(lb_id)
     }
 
     fn now(&self) -> String {
@@ -180,7 +180,7 @@ impl Lb {
             time: self.now(),
             name: account.username.clone(),
             lb_version: get_code_version().into(),
-            lb_id,
+            lb_id: lb_id?,
             rust_triple: format!("{arch}.{family}.{os}"),
             server_url: account.api_url.clone(),
             integrity: format!("{integrity:?}"),


libs/lb/lb-rs/Cargo.toml
```diff
@@ -34,8 +37,8 @@ tracing-subscriber = "0.3.9"
 tracing-appender = "0.2"
 crossbeam = "0.8.1"
 qrcode-generator = "4.1.6"
-db-rs = "0.3.3"
-db-rs-derive = "0.3.3"
+db-rs = { path = "../../../../db-rs/db" }
+db-rs-derive = { path = "../../../../db-rs/derive" }
```
Parth:
```suggestion
db-rs = "0.3.4"
db-rs-derive = "0.3.4"
```

libs/lb/lb-rs/src/blocking.rs
```diff
@@ -325,12 +326,16 @@ impl Lb {
...
         self.rt
-            .block_on(self.lb.debug_info(os_info))
+            .block_on(self.lb.get_debug_info_string(os_info))
             .unwrap_or_else(|e| format!("failed to produce debug info: {:?}", e.to_string()))
     }
 
+    pub fn get_debug_info(&self, os_info: String) -> LbResult<DebugInfo> {
+        self.rt.block_on(self.lb.get_debug_info(os_info))
+    }
+
```
Parth: I understand why this is a needed change in many places, but I think you should just impl a `to_string()` method on `DebugInfo` and the places that need the string can just do `get_debug_info().to_string()` rather than exposing multiple endpoints like this.

libs/lb/lb-rs/src/lib.rs
```diff
@@ -35,6 +35,7 @@ pub struct Lb {
     pub events: EventSubs,
     pub syncing: Arc<AtomicBool>,
     pub status: StatusUpdater,
+    pub id: LbID,
```
Parth: I would stick this at the top

libs/lb/lb-rs/src/service/debug.rs
```diff
@@ -1,22 +1,24 @@
...
 use tokio::io::AsyncWriteExt;
 
-#[derive(Serialize)]
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone)]
 pub struct DebugInfo {
     pub time: String,
     pub name: String,
     pub last_synced: String,
     pub lb_version: String,
+    pub lb_id: LbID,
```
Parth: I would stick this at the top

libs/lb/lb-rs/src/service/lb_id.rs
```diff
@@ -0,0 +1,43 @@
...
+use std::path::Path;
+use std::path::PathBuf;
+use uuid::Uuid;
+
+/// Persistant for a device across sessions, would only be destoreyd on re-download
+/// or delation of the lb_id.bin file.
+#[derive(Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Debug, Hash)]
+pub struct LbID(Option<Uuid>);
+
+pub const FILE_NAME: &str = "lb_id.bin";
```
Parth: I don't think this deserves it's own IO. You can add a field to db-rs of type `Single` for this.

libs/lb/lb-rs/src/service/sync.rs
```diff
@@ -148,6 +149,13 @@ impl Lb {
             }
         }
 
+        let debug_info = self.get_debug_info("".into()).await?;
+
+        let _ = self
```
Parth: you rarely want to `let _` to avoid handling an error. I think at a minimum this should end up in the logs

libs/lb/lb-rs/src/service/sync.rs
```diff
@@ -148,6 +149,13 @@ impl Lb {
...
+            .client
+            .request(self.get_account()?, UpsertDebugInfoRequest { debug_info })
```
Parth: As it stands currently, won't this send for all users? Don't you need an `self.is_beta()` check here?

server/local.env
```diff
@@ -26,6 +26,8 @@ APPLE_ISSUER_ID=${APPLE_ISSUER_ID:=1234}
 APPLE_ASC_SHARED_SECRET=${APPLE_ASC_SHARED_SECRET:=shared_secret}
 APPLE_MONTHLY_SUB_GROUP_ID=${APPLE_MONTHLY_SUB_GROUP_ID:=public_key}
 
+DISCORD_WEBHOOK_PANIC_CAPTURE="https://discord.com/api/webhooks/{}{}"
```
Parth: is this right? `{}{}` seems sus

server/src/debug_info.rs
```diff
@@ -0,0 +1,135 @@
...
+        let maybe_old_debug_info = db.debug_info.insert(
+            owner,
+            context.request.debug_info.lb_id,
+            context.request.debug_info,
+        )?;
+
+        let old_panics_count =
+            if let Some(debug_info) = maybe_old_debug_info { debug_info.panics.len() } else { 0 };
+
+        let maybe_webhook_url = env::var("DISCORD_WEBHOOK_PANIC_CAPTURE");
```
Parth: this should probably have some presence here: https://github.com/lockbook/lockbook/blob/master/server/src/config.rs instead of being checked at this point.

Mostly code organization, and operational clarity reasons, not a performance one, though it does likely perform marginally better.

server/Cargo.toml
```diff
@@ -44,11 +49,15 @@ tracing-subscriber = "0.3.9"
...
+    "rustls-tls",
+    "multipart",
+] }
 jsonwebtoken = "8.2.0"
-x509-parser = { version = "0.15.0", features = ["verify", "validate"]}
-db-rs = "0.3.1"
-db-rs-derive = "0.3.1"
+x509-parser = { version = "0.15.0", features = ["verify", "validate"] }
+db-rs = { path = "../../db-rs/db" }
+db-rs-derive = { path = "../../db-rs/derive" }
```
Parth:
```suggestion
db-rs = "0.3.4"
db-rs-derive = "0.3.4"
```

server/src/debug_info.rs
```diff
@@ -0,0 +1,135 @@
...
+            let webhook_url = match maybe_webhook_url {
+                Ok(ref val) => val,
+                Err(_) => break,
+            };
+
+            if let Some(panic) = debug_info.panics.get(i) {
+                self.send_panic_to_discord(&debug_info, webhook_url, panic)
+                    .await?;
+            }
+        }
```
Parth: something feels wrong about this. I think using panic count as a heuristic for whether or not to send is valid. But interacting with discord with potentially heavy requests in a loop is prob not the move. The latest debug info whatever that is will contain all the panics.

server/src/debug_info.rs
Parth: I think we want the panics to make it to the lolg regardless. Presently only the send to discord function will do that.

You probably can probably setup the discord client to have the webhook url "inside" it, feel like that's the pattern we apply for pagerduty and stripe. Could be wrong.

server/src/debug_info.rs
```diff
@@ -0,0 +1,135 @@
...
+            "username": "Panic Reporter",
+            "embeds": [{
+                "color": 14622784,
+                "author": { "name": debug_info.name },
+                "title": panic_title.unwrap_or("".to_string()),
+            }]
+        });
+
+        let debug_info_part = multipart::Part::bytes(serde_json::to_vec_pretty(&debug_info)?)
+            .file_name("debug_info.json")
```
Parth: if you feel like it -- consider calling this `{username}_debug.json` or something like that, so if we download these files and end up with a lot of them it's more clear. Up to you.

server/src/debug_info.rs
```diff
@@ -0,0 +1,135 @@
...
+        let mut panic_title = None;
+
+        if let Some(new_line_index) = maybe_new_line_index {
+            panic_title = Some(panic[0..new_line_index].to_string());
+        }
+
+        let payload = json!({
+            "username": "Panic Reporter",
+            "embeds": [{
+                "color": 14622784,
```
Parth, line 91: This is cool can, is this a hex color? if so consider this format: `0xFFFFFF`

server/src/debug_info.rs
```diff
@@ -0,0 +1,135 @@
...
+                response.status(),
+                response.text().await?
+            );
+        }
+        Ok(())
+    }
+
+    pub fn is_beta_user<E: Debug>(
+        db: &ServerDb, public_key: &PublicKey,
+    ) -> Result<bool, ServerError<E>> {
```
Parth: I don't see any `Err` branches or `?` here -- maybe this could just return a bool instead?

libs/lb/lb-rs/src/service/lb_id.rs
```diff
@@ -0,0 +1,43 @@
...
+use std::fs;
+use std::io;
+use std::path::Path;
+use std::path::PathBuf;
+use uuid::Uuid;
+
+/// Persistant for a device across sessions, would only be destoreyd on re-download
+/// or delation of the lb_id.bin file.
+#[derive(Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Debug, Hash)]
+pub struct LbID(Option<Uuid>);
```
Parth: I think ideally LbId would ideally be == `Uuid`. I don't think there's a situation where we would want none here. 

When you move this into lb-rs db-rs, the function can be aggro about making sure there's a value and all that.

Parth:
rebase with care: https://github.com/lockbook/lockbook/pull/4121/files

server/local.env
ad-tra: it's wrong, but the actual link is a secret

server/src/debug_info.rs
ad-tra: webhook url inside of the client is nice ergonomics, but unfortunately Reqwest [doesn't support the base url pattern](https://github.com/seanmonstar/reqwest/issues/2055)

server/src/debug_info.rs
ad-tra: yeah it made more sense when each panic had it's own discord message instead of the current approach of sending the entire debug_info in the message

server/src/debug_info.rs
ad-tra: no they don't support hex

libs/lb/lb-rs/src/io/mod.rs
```diff
@@ -66,6 +68,10 @@ pub async fn migrate_and_init(cfg: &Config, docs: &AsyncDocs) -> LbResult<CoreV4
         CoreDb::init(cfg.clone()).map_err(|err| LbErrKind::Unexpected(format!("{err:#?}")))?;
     let mut old = CoreV3::init(cfg).map_err(|err| LbErrKind::Unexpected(format!("{err:#?}")))?;
 
+    if db.id.get().is_none() {
+        db.id.insert(LbID::generate())?;
+    }
```
Parth: I would add this to whoever calls this function, only reason I say this is because this function is going to be deleted soon and it's possible this will be missed

libs/lb/lb-rs/src/model/account.rs
```diff
@@ -151,10 +152,7 @@ impl Account {
     /// their name to this list. Certainly telemetry in lockbook will always be opt in but the
     /// mechanism of consent may evolve over time.
```
Parth: Can you stick these up by the const now

libs/lb/lb-rs/src/service/debug.rs
```diff
@@ -27,6 +29,19 @@ pub struct DebugInfo {
...
+}
+
+impl DebugInfoDisplay for LbResult<DebugInfo> {
+    fn to_string(&self) -> String {
+        match self {
+            Ok(debug_info) => serde_json::to_string_pretty(debug_info).unwrap_or_default(),
+            Err(err) => format!("Error retrieving debug info: {:?}", err),
+        }
+    }
+}
```
Parth, line 43: I meant just `impl DebugInfo` but this is fine as well

libs/lb/lb-rs/src/service/debug.rs
```diff
@@ -43,6 +58,16 @@ impl Lb {
...
 
+    async fn lb_id(&self) -> LbResult<LbID> {
+        let tx = self.ro_tx().await;
+        let db = tx.db();
+
+        db.id
+            .get()
+            .copied()
+            .ok_or_else(|| LbErr::from(LbErrKind::Unexpected("Couldn't get Lb ID".to_string())))
+    }
```
Parth: I think this is the thing that should just generate and set it actually.

server/local.env
Parth: Ideally you would stick a literal message here then

PR #4116, opened 2026-01-19
linux: add a flatpak package folder

Amumug:


utils/dev/flatpak-package/logo.svg
Parth, line 1: This is the only thing that probably doesn't belong here. This file probably exists somewhere else in the repo already. It's not really a big space issue because svg's take up such little space, it's more of a -- it'll be annoying to update all the icons if we change icons (we've already changed once before). So try to re-use one of the existing locations / icons. And let me know if this style of icon is actually not present in the repository.

Amumug: @Parth i think i will delete it and use the one in snap-package folder

Parth:
@Amumug you should use the one at this location: https://github.com/lockbook/lockbook/blob/master/docs/graphics/logo.svg

utils/dev/flatpak-package/lockbook.appdata.xml
Parth, line 1: This should be in the other repo

Parth:
image
follow this structure

find more details here: https://github.com/lockbook/lockbook/blob/master/docs/CONTRIBUTING.md

Parth:
https://github.com/lockbook/lockbook/pull/4116#issuecomment-3786150733

you didn't do this

PR #4137, opened 2026-01-26
md: show images proper rotation

Parth:
process exif data and rotate the image accordingly, for markdown only for now


Parth:
previously we baked the rotation into the image in https://github.com/lockbook/lockbook/pull/4129:
image
Without re-processing the images we just have it the right way now:
image

Issue #3729, opened 2025-08-11
auto inserted - bullet character is often a bad time


Parth:
while thinking about: https://discord.com/channels/1014184997751619664/1403798488613130250/1403798488613130250

another situation where this annoying-ness is encountered is when someone enters a `-` as their bullet character. When they press enter and a bullet is included it turns it into a heading momentarily, a pretty confusing situation for first time users and/or users unfamiliar with markdown.

Users who don't know about markdown and are trying to use bullets in a non-bullet env often reach for `-` so it's overall a very unforgiving situation for them.

Some solutions come to mind:
- aggressive and compliant approach: always replace bullet characters with `+` / `*`
- non-compliant approach: require more than one `--` for a h2 to get created
- complex approach: model the ideal of provisional input: when the editor places a bullet for you, the editor is not trying to create an h2. And if non-bullet behavior happens, like they press newline, or mouse away, or something like that. The text is not "committed". The bullet is only committed if they enter input.

tvanderstad:
`+` isn't recognized by Slack last I checked, making `*` the ideal unordered list item character. That's why cmd+shift+8 and the toolbar insert with `*`, making the `-` situation only affect users who are typing their markdown.

Parth:
@tvanderstad i agree * is the reigning bullet character king. (so I would be down to replace the characters with stars in the aggressive course of action.)

I've observed two things that seem to make it an issue:
Normies seem to try dashes as bullets first.
Experienced people sometimes use dashes, but rarely are they exposed to the heading syntax so even experienced people are quite confused when their new bullet makes their first bullet gigantic.

tvanderstad:
Does it make an earlier bullet gigantic? I thought it would just make the paragraph above it gigantic. A test:

```md
- test
-
```
- test
-

another with a nested bullet:

```md
- test
  -
```
- test
  -

...I see

tvanderstad:
Perhaps I could change the behavior of setext (underlined) headings so they're revealed when the cursor is in the underline. The text on the above line would become grey/monospace, so you'd know something was up and if your intention was to create a setext heading you'd get confirmation that your syntax is correct, but it wouldn't be so jarring

PR #4139, opened 2026-01-26
md: disable setext headings

tvanderstad:
fixes https://github.com/lockbook/lockbook/issues/3729

note that:
```md
h
-
```
is now a single-cell table instead. Perhaps this is not the solution we want.


Parth:
I think this is a net benefit

I don't have a single one of these so again I'm good with this

I wonder if we could just disable them for single `-`. I've never seen a settext that didn't have like `---` or more. Can `=` still be used?

tvanderstad:
None can be used; had a real bad time with that kind of hack in previous md impl

PR #4149, opened 2026-01-28
iOS: stop onboarding text from being cut off

### LINKED ISSUE #4084: Onboarding private key explanation text cut off
The private key explanation is cut off with "..."

Reported by @tvanderstad

smailbarkouch:
If the font size is too large on the device - then the onboarding text will be cut off. This is primarily an issue for the account phrase onboarding screen - so made the view scrollable if all the text can't fit.

fixes #4084


Parth:
This is being reverted as it fails to build for macOS @smailbarkouch

Parth:
It can & will be released to iOS, try to fix ASAP to avoid further weirdness

PR #4171, opened 2026-02-02
cli: bear import bugfixes

### LINKED ISSUE #4118: bear migration interprets urls as tags

sergigp:
image

* skip non markdown files. I made this decision bc import doc points that user needs to export bear notes as markdown.
* use `from_utf8_lossy` to replace non utf8 chars
* improved reporting messages. Just reporting notes imported instead files. User is importing his bear notes, so I guess he doesn't care about lockbook internal implementation (folders, images counted as documents, etc.)

fixes #4118


clients/cli/src/migrate.rs
```diff
@@ -57,8 +74,10 @@ pub async fn bear(path: PathBuf) -> CliResult<()> {
         };
 
         let f = |status: ImportStatus| {
-            if let ImportStatus::FinishedItem(_) = status {
-                count.fetch_add(1, Ordering::Relaxed);
+            if let ImportStatus::FinishedItem(file) = status {
+                if file.name.ends_with(".md") {
```
Parth, line 77: I think we do not need this to be checked again, ideally we do just report any and all files imported at this stage

Parth:
Additionally, you can stick `fixes #issue-number` in your PR description, merging this PR will close that issue.

Also I think this is a great first choice for an issue!

clients/cli/src/migrate.rs
sergigp: In old impl folders were reported as imported files too. Another think we can do is to increment the counter based on `file_type` so it will count both notes and images... or if you think the old implementationmade more sense I can revert 👌

Parth: I see, for a bear user perhaps they aren't thinking in terms of individual folders, images, and files. They're just thinking in terms of the individual note files that were imported. That makes sense.

clients/cli/src/migrate.rs
```diff
@@ -68,49 +86,113 @@ pub async fn bear(path: PathBuf) -> CliResult<()> {
...
+                }
             }
-        }
-
-        if found_hash {
-            if path.is_empty() && (char == ' ' || char == '#') {
-                found_hash = false;
-                continue;
+        } else if char == '/' {
+            // Nested tag: reset to keep only last segment (eg #work/project -> project)
```
Parth: I don't think we only want to keep the last segment, `work/project` is the location we want the file to end

clients/cli/src/migrate.rs
sergigp: ok, so if I understood correctly no changes needed here as we are already counting notes with changes implemented in this pr. Feel free to reopen if this is not what you want 👌

clients/cli/src/migrate.rs
sergigp: done!

clients/cli/src/migrate.rs
```diff
@@ -68,49 +86,113 @@ pub async fn bear(path: PathBuf) -> CliResult<()> {
...
 
-    if !path.is_empty() {
+    if in_tag && !path.is_empty() {
         paths.push(path);
     }
 
     paths
 }
+
+#[cfg(test)]
```
Parth, line 143: great addition

Parth:
Congrats on your first contribution!

clients/cli/src/migrate.rs
sergigp: <3

PR #4211, opened 2026-02-11
md: simplify left/right arrow navigation

tvanderstad:
* removes a concept of paragraphs from the previous editor implementation
* fixes an issue where the cursor wouldn't respond to left/right arrows, such as at the end of:
  ```md
  * x
    |
  ```

This change has the usual high risk of a foundational editor cleanup. It was QA'd on macOS and iOS but internal dogfooding is also recommended before release.

This change causes triple-tapping a text location to select the whole source line, which can include/reveal bulleted list items, which could be considered a feature or a bug.

### DIFF
--- a/libs/content/workspace/src/tab/markdown_editor/widget/block/leaf/html_block.rs
+++ b/libs/content/workspace/src/tab/markdown_editor/widget/block/leaf/html_block.rs
@@ -25,6 +25,4 @@ impl<'ast> Editor {
             true,
         );
     }
-
-
 }


libs/content/workspace/src/tab/markdown_editor/input/advance.rs
```diff
@@ -10,14 +10,26 @@ use super::cursor;
...
+                if backwards {
+                    if result.0 > 0 {
+                        result -= 1;
+                    }
+                } else {
+                    result += 1;
+                    result = result.min(self.buffer.current.segs.last_cursor_position());
+                }
+                result
+            }
```
tvanderstad, line 30: The simplification, in a nutshell

PR #4216, opened 2026-02-12
egui: 0.28 -> 0.30

tvanderstad:


tvanderstad:
- [x] file tree icon alignment appears to be off in egui clients; landing page and markdown toolbar are fine; possibly settings and side panel buttons also misaligned
  
  image

PR #4230, opened 2026-02-19
docs: grammar fixes

tvanderstad:


tvanderstad:
I need to update in lb docs in lockbook and then run the script

PR #4240, opened 2026-02-26
lb-rs: lazy file support

### LINKED ISSUE #1918: core shouldn't retrieve all files

### LINKED ISSUE #4178: fs auto-syncer not resilient
- if the syncer crashes it doesn't resume
- then ctrl-c doesn't work

reported by @steverusso

Parth:
This PR performs a sweeping cleanup of legacy ideas, and fixes #1918 fixes #4178. 

# why now?

- people will paste a large doc in, taking sync offline. We improved the experience of telling people what's going on, but we shouldn't stop people from syncing. Lazy files will attempt to make progress on any documents that it can make progress for.
- initial imports have been a bad time for people with bad internet connections or people with lots of files. Very bad for people with both. We want to encourage people ingressing lots of files. Separately interrupting an long initial sync leaves clients in a broken state. This is skill issue from the client implementors perspective, but this PR will make this a non issue regardless.
- right now people like steve are annoyed when they get access to a folder like lb-maintainers folder. Because it degrades their experience regardless of whether or not they accepted the files. We want people to be sharing
- egress is one of our biggest costs, and we can be smarter about the default egress behavior that most people experience.

# old ideas? 

we implemented a few newer systems, and left the old legacy systems in place while the rest of the platform caught up. It is time to eliminate a number of these old ideas:
- `SyncProgress` closure was deprecated via lb-rs events and the status api. A caller is not guaranteed to be the only entity calling sync, and can't rely on the return value / progress closure for an accurate view into what's going on. Most clients already implement the sub and status api, CLI and lbfs were the only ones left. They were updated and the closure was eliminated. The closure was also particularly problematic because it would execute the progress code mid-sync, sometimes causing problems. The events api doesn't have this problem.
- `SyncProgress` was the old idea for events, and `WsStatus` was the old idea for status, `WsStatus` and a few other pieces of tech debt in `WsOutput` have also been eliminated.
- V1 endpoints, and other artifacts prior to #3737. All traffic has been forced to use the newer code for well over 6 months, now time to cleanup.
- sync is now triggered directly from within lb-rs. Clients can request a sync whenever, and will openly subscribe to any syncs and represent their status regardless of who the caller is. This is a mild simplification to `task_manager` which meant sync's were tied to the ui-frame-lifecycle. Now `task_manager` is only responsible for loads and saves of files.

# new ideas?

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -11,7 +11,8 @@ on:
         required: false
 jobs:
   DeployAndroid:
-    if: github.repository == 'lockbook/lockbook'
+    # disabled until android lazy-files are implemented
+    if: github.repository != 'lockbook/lockbook'
     runs-on: [self-hosted, ci]
     env:
       ANDROID_RELEASE_STORE_PASSWORD: ${{ secrets.ANDROID_RELEASE_STORE_PASSWORD }}


Parth:
some notes for myself
- Also would like to get rid of the annoying atomic bool in sync
- maybe the move is to model sync more closely after search -- just a thing that subscribes to events and does whatever needs to be done
- will probably need to expand the scope of the events -- everything will probably need to have an actor so that sync can filter out it's own events. Is there some overlap with #3853? Should an actor be sort of a session id? Should the session ids be per-core? Is there ever a situation you would want to differentiate two versions of workspace? Yes, probably in the #3853 situation. For metadata who needs to know? Well probably just sync needs to know that it, itself did not update that metadata and then it can sync it pretty safely. Everyone else perhaps can tolerate duplicate metadata updates. Maybe more annoying with documents, but documents have the hmac thing going on. Perhaps there can be a more general use of sequence numbers, and if you trigger an event you don't need to filter out "your" event, you can just filter out events that are from sequence numbers that you created. 
- can evaluate this a bit later, not critical to first steps

Parth:
Some doc service lifecycle thoughts:

# read side

- if someone asks for a file and it's not available locally, go get it. If it's not available on the server, trigger a metadata sync. If that results in an offline, then the file cannot be fetched because you're offline.
  - So we should never commit a new version of base metadata document without having first retrieved the file
  - what if we don't want to pull a new version of the file based on some **rules**? We would only run into problems if there are local content updates to a document. So whatever the **rules** are, we should be syncing contents of files that have updates locally.

# write side

- if a local update to a document exists, try to push it. If the push fails due to old version being incorrect, trigger a metadata sync.

# rules

eventually these will be configurable, but we need a starting configuration with the following goals:
- no perceived impact to anyone
- initial syncs are instant
- significant reduction in useless network activity client & serverside

so I'm thinking that initially we'll get:
- all `.md` and `.svg` that are in our file tree (not pulling things with un-accepted shares)
- anything we have a `local` copy of
- anything we have already pulled, if there's an update we must pull

## future rules
(in order of expected value)

directory based:
- don't pull anything in this directory
- pull everything in this directory / document (can also be applied to shared with me)
- don't push anything in this directory

device based:
- utilize this much space as effectively as possible
- keep this device fully offline
- keep this device fully online

Parth:
## levels of importance to a file we need to pull

- tier 1 importance - there are going to be problems if we don't pull this file. This is something like 
- tier 2 importance - if we can pull this file we probably should based on the user's preferences.

I think tier one exists because if we edit a file and there are changes from the server we will orphan some critical piece of data for a 3 way merge if we update that base's hmac without having access to the corresponding file. If we don't pull it in time, it may not exist at some point in the future and I believe the situation can become unsolvable. So at a minimum tier 1 is any file that we have a local modified hmac for. 

What about files for which there is about to be a local version for? Maybe these are files that have been fetched and workspace is viewing them, maybe they were tier 2 files that are about to become tier 1 files. Maybe the best bet is to not have a distinction.

My motivation to have a distinction is to minimize the amount of network activity required before someone can start using the app. If someone has 10k markdown files, we'll grab them, but I don't want them to have to wait to start using the app while that sync happens. 

Basically the question is, can we make logging in an O(1) experience?

Well, in the initial sync situation, this is less of a problem, they don't have any files yet. 

Maybe tier 1 is: the set of files that have any local changes + the set of files that have prior versions. During initial sync this will be no files, and during ongoing syncs we're less sensitive to this anyways. 

Tier 2 files are any files that match the rules they've specified and we can fetch those in the background, continuously while the app is in use. Hopefully providing nice updates along the way.

Parth:
# cleanup

the last open question in my mind is around cleanup. Presently, `safe_write` is surgical with it's cleanup, and sync takes it's time with a garbage collection like sweep. There's an atomic bool in the mix that deserves some scrutiny, it probably isn't fully resistant to race conditions, but maybe that's just fine. 

 Is there a way we can just generally be surgical everywhere and not do this garbage collection style step? 

we would probably just need to mark the provisional files as such. Or maybe more generally the surgical approach is just: know the id-hmacs before and after your operation. If you're operating on a single file (safe write) you can just eliminate that single, older hmac. If you're operating on a set (sync) then you should just get rid of what is no longer in the old set. Both parties hold a lock while doing this, and that's the only time the files are deleted. 

I don't think this would work either actually. I think there is still the risk of a sync starting, pulling a file B that safe_writes is trying to delete. We're not going to hold a lock while doing a network operation, and we don't want to hold the file contents in memory. So there's always a chance of something slipping through the cracks.

I guess we should just make sure that atomic bool is behaving properly.

Parth:
pretty on-track for this to be merged once the current `master` is qa'd and released.

Things I still need to do:
+ [x] sync to have something that triggers an autosync when the user does things. It will just drain all the events on a loop (solution to a problem)
+ [x] there needs to be some way for lb to keep track of if a user is active or not and modulate aggression accordingly (formerly done in ws)
+ [x] uncompressed usage needs to call lb's read_document instead of the helper, that needs to get moved into syncer and marked as private
+ [x] syncer will have it's own queue of maintainence tasks that it will do after each sync

good candidate for follow up would be to make status and file tree sicker while all the sync stuff is still fresh in my head

Parth:
updated remaining items checklist:
- [x] test_utils needs a calculate_work replacement
- [x] sync needs to issue a `pending_shares_changed` if pending shares have changed
- [x] make doc pushes resilient 
- [x] restore debug info probably
- [x] sync button `todo!`
- [x] apple needs to build (delete usages of uncompressed usage, sync status, etc)
- [x] apple needs to delete it's initial sync flow
- [ ] android needs to build (delete usages of uncompressed usage, sync status, calculate work, etc) @ad-tra 
- [ ] android needs to delete it's initial sync flow @ad-tra

Parth:
fixed #4178

Parth:
Things to QA:
- [ ] initial sync pretty much everywhere
- [ ] collaboration
- [ ] random access of files
- [ ] lb-fs linux
- [ ] lb-fs macOS

Parth:
Apart from android && the release this is now ready to be merged

PR #4241, opened 2026-02-28
infra: add new lbdev command

Amumug:


utils/flatpak-cargo-generator.py
```diff
@@ -0,0 +1,511 @@
+#!/usr/bin/env python3
```
Parth: We probably don't want to vendor this file ourselves, we probably want to download it as part of the job.

Parth: I deleted it just so I don't have to look at it before I hit the "run workflow" button

utils/lbdev/src/releaser/linux/desktop.rs
```diff
@@ -280,3 +281,139 @@ pub fn push_aur() -> CliResult<()> {
...
+        .create(true)
+        .truncate(true)
+        .open("../net.lockbook.Lockbook/net.lockbook.Lockbook.json")
+        .unwrap();
+    file.write_all(manifest.as_bytes()).unwrap();
+    Ok(())
+}
+
+pub fn update_flatpak() -> CliResult<()> {
+    let version = "26.2.11".to_string();
```
Parth: This needs to be calculated dynamically, not hard-coded, see how the other workflows are doing it

Parth: version is about to be bumped on master: https://github.com/lockbook/lockbook/actions/runs/22970475342

utils/lbdev/src/releaser/linux/desktop.rs
```diff
@@ -280,3 +281,139 @@ pub fn push_aur() -> CliResult<()> {
...
+        }}
+    ]
+}}"#
+    );
+
+    let mut file = OpenOptions::new()
+        .write(true)
+        .create(true)
+        .truncate(true)
+        .open("../net.lockbook.Lockbook/net.lockbook.Lockbook.json")
```
Parth: This needs to be cloned before it can be used, this only works on your local machine right now

utils/lbdev/src/releaser/linux/desktop.rs
```diff
@@ -280,3 +281,139 @@ pub fn push_aur() -> CliResult<()> {
...
+    Command::new("git")
+        .args([
+            "clone",
+            "--depth=1",
+            "--branch",
+            &version,
+            "https://github.com/lockbook/lockbook.git",
+            &clone_dir,
+        ])
+        .assert_success()?;
```
Parth, line 370: this is for a workflow that will run in this repo, you do not need to clone yourself

utils/lbdev/src/releaser/linux/desktop.rs
```diff
@@ -280,3 +281,139 @@ pub fn push_aur() -> CliResult<()> {
...
+
+    Command::new("curl")
+        .args([
+            "-fL",
+            &format!("https://github.com/lockbook/lockbook/archive/refs/tags/{version}.tar.gz"),
+            "-o",
+            &tarball,
+        ])
+        .assert_success()?;
```
Parth, line 384: for local testing consider hard coding the url for the latest release, we'll change this code back to what you have here before we merge

Amumug: do i just need to comment it?

utils/lbdev/src/releaser/linux/desktop.rs
Amumug:
```
Command::new("git")
        .args(["clone", "https://github.com/lockbook/net.lockbook.Lockbook.git"])
        .assert_success()?;
```

utils/lbdev/src/releaser/linux/desktop.rs
Parth: just hardcode this for now https://github.com/lockbook/lockbook/archive/refs/tags/26.2.11.tar.gz

utils/lbdev/src/releaser/linux/desktop.rs
Amumug: or i just comment it?

Issue #3888, opened 2025-10-10
Zoom support


Parth:
Requested by @arnoldoree

Parth:
related #1894

Parth:
requested by Chetna

PR #4259, opened 2026-03-12
ws: zoom

tvanderstad:
Fixes zoom not working on FFI clients and persists the zoom factor.

fixes https://github.com/lockbook/lockbook/issues/3888

### DIFF
--- a/libs/content/workspace/src/show.rs
+++ b/libs/content/workspace/src/show.rs
@@ -46,6 +46,11 @@ impl Workspace {
             self.cfg.set_tabs(&self.tabs, self.current_tab);
         }
 
+        let zoom = self.ctx.zoom_factor();
+        if zoom != self.cfg.get_zoom_factor() {
+            self.cfg.set_zoom_factor(zoom);
+        }
+
         mem::take(&mut self.out)
     }
 
--- a/libs/content/workspace/src/workspace.rs
+++ b/libs/content/workspace/src/workspace.rs
@@ -83,6 +83,7 @@ impl Workspace {
         let files = FileCache::new(core).log_and_ignore();
 
         let cfg = WsPersistentStore::new(core.recent_panic().unwrap_or(true), writeable_path);
+        ctx.set_zoom_factor(cfg.get_zoom_factor());
 
         let mut ws = Self {
             tabs: Default::default(),
@@ -878,6 +879,7 @@ pub struct WsPresistentData {
     auto_save: bool,
     auto_sync: bool,
     landing_page: LandingPage,
+    zoom_factor: f32,
 }
 
 impl Default for WsPresistentData {
@@ -890,6 +892,7 @@ impl Default for WsPresistentData {
             canvas: CanvasSettings::default(),
             markdown: MdPersistence::default(),
             landing_page: LandingPage::default(),
+            zoom_factor: 1.,
         }
     }
 }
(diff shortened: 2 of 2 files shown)


clients/linux/src/input/pointer.rs
```diff
@@ -50,7 +50,7 @@ fn handle(
         }
     } else {
         // button event
-        let pos = egui::Pos2::new(event_x as f32 / scale, event_y as f32 / scale);
+        let pos = app.renderer.pos_from_pixels(event_x as f32, event_y as f32);
```
tvanderstad, line 53: the renderer exposes new functions for converting positions, accounting for the scale internally

clients/windows/src/window.rs
```diff
@@ -300,8 +299,7 @@ fn handle_message(hwnd: HWND, message: Message) -> bool {
             if let Some(ref mut window) = maybe_window {
                 if let Some(ref mut app) = window.maybe_app {
                     // events sent to app every frame
-                    app.renderer.context.set_pixels_per_point(window.dpi_scale);
```
tvanderstad, line 303: egui's `set_pixels_per_point()` actually just sets the zoom factor, accounting for the native pixels per point to choose a zoom factor that gives you what you want. Here, we were effectively using the zoom factor to set the native pixels per point, which was fine until we also wanted to use the zoom factor to zoom.

libs/content/egui_wgpu_renderer/src/lib.rs
```diff
@@ -76,6 +76,22 @@ impl<'w> RendererState<'w> {
         }
     }
 
+    /// Call to update the screen ppp based on an up-to-date native ppp. This is
+    /// how the app responds to native ppp changes, such as when the app is
+    /// moved to a display with a different pixel density.
+    pub fn set_native_pixels_per_point(&mut self, native: f32) {
+        self.screen.pixels_per_point = native * self.context.zoom_factor();
```
tvanderstad, line 83: As a rule of thumb, egui uses "pixels_per_point" to refer to the product of the native pixels per point * the zoom factor. The screen descriptor is expected to use this product, not the native pixels per point.

Native clients are now expected to call this function instead of setting self.screen.pixels_per_point directly so that the zoom factor can be accounted for automatically.

libs/content/egui_wgpu_renderer/src/lib.rs
```diff
@@ -237,8 +261,14 @@ impl<'w> RendererState<'w> {
                 self.screen.size_in_pixels[1] as f32 / self.screen.pixels_per_point,
             ),
         });
-        self.context
```
tvanderstad, line 240: Before, we assigned the pixels per point, leaving the native pixels per point at the default 1.0. This assigned a zoom factor that matched the native pixels per point but could not additionally support zooming.

libs/content/workspace/src/show.rs
```diff
@@ -46,6 +46,11 @@ impl Workspace {
             self.cfg.set_tabs(&self.tabs, self.current_tab);
         }
 
+        let zoom = self.ctx.zoom_factor();
+        if zoom != self.cfg.get_zoom_factor() {
+            self.cfg.set_zoom_factor(zoom);
+        }
+
```
tvanderstad, line 53: note: here's where we persist the zoom factor when it changes

PR #4265, opened 2026-03-16
ws: theme refinements

tvanderstad:
Non-functional changes:
* factors uses of `lerp_gamma()` for neutrals into pallete_v2.rs
* removes editor theme; editor now uses the ws theme directly
* editor code block syntax highlighting now respects the ws theme

Functional changes:
* foreground green, yellow, and cyan: darkened for light mode
* egui sync button: blue -> grey
* egui suggested docs icon from blue -> magenta
* md cursor/selection, inline code, active fold buttons in dark mode: green -> blue
* md inline code: re-introduces borders
* md inline code and table header rows: lightens background in light mode
* md code blocks: colors updated to match theme

**Dark Mode**
Before:
Screenshot 2026-03-16 at 5 26 08 PM

After:
Screenshot 2026-03-16 at 5 26 45 PM

**Light Mode**
Before:
Screenshot 2026-03-16 at 5 25 57 PM

After:
Screenshot 2026-03-16 at 5 26 56 PM


libs/content/workspace/src/theme/palette_v2.rs
```diff
@@ -105,32 +162,32 @@ impl ThemeVariant {
...
+        Self::travis(current)
     }
 
-    pub fn travis_prophecy(current: Mode) -> Self {
+    pub fn travis(current: Mode) -> Self {
         Self {
             current,
             light: ThemeVariant {
-                black: hex_color!("#101010"),
+                black: hex_color!("#000000"),
```
Parth, line 172: @tvanderstad perhaps the mnemonic tables need to be updated?

Parth: https://github.com/mnemonic-theme/zed
https://github.com/mnemonic-theme/palette

tvanderstad: Perhaps, I may have just chosen darker versions of some colors for text only in other themes

Parth: wym

tvanderstad: Like I may have three greens in my Zed theme, with the green used for text in light mode just slightly darker than the green for other foreground things in light mode like icons

Parth: i see

libs/content/workspace/src/theme/palette_v2.rs
```diff
@@ -68,6 +68,63 @@ impl Theme {
...
+            Mode::Light => self.light.white.lerp_to_gamma(self.light.grey, 0.2),
+            Mode::Dark => self.dark.black.lerp_to_gamma(self.dark.grey, 0.2),
+        }
+    }
+
+    /// Returns the background neutral color i.e. white in light mode,
+    /// near-black in dark mode. Used for areas with editable content, similar
+    /// to egui's `extreme_bg_color`.
+    pub fn neutral_bg(&self) -> Color32 {
+        match self.current {
```
Parth, line 123: isn't this philosophically wrong? Shouldn't we be going to the dark variant's colors for this?

tvanderstad: why so? seems right to pull a color from the active mode

Parth: I thought the background was pulled from the inverse palette

tvanderstad: I just wanted the darkest black and lightest white here, didn't think too much harder about it

tvanderstad: I interpret `Mode::Dark` as dark theme and `self.dark` as darker colors

tvanderstad: 'dim' could be a reasonable ANSI-inspired alternate term for the second definition

Parth: but yeah the darkest black is found in light theme's black, the lightest white is dark theme's white

Parth: fairly sure extreme_bg_color was also pure black earlier, have been which is certainly what we want at-least on mobile as a background, but open minded generally on this front. Regardless the before / after is misleading here.

PR #4289, opened 2026-03-21
server, apple: support http external links

smailbarkouch:
Expose two new endpoints:
+ `.well-known/apple-app-site-association` - will be used by apple to auto-route to our app
+ `open/<file_uuid>` - to render a basic HTML site as a fallback

After this - I am going to create a new subdomain (`open.lockbook.net`) that will DNS link to `files` route. I believe I need assistance from @Parth in order to set this up.

### DIFF
--- a/server/src/static_files.rs
+++ b/server/src/static_files.rs
@@ -1,8 +1,7 @@
 use tracing::{Level, info, span};
 use warp::{Filter, http::Response, hyper::Body};
 
-const APPLE_APP_SITE_ASSOCIATION: &str =
-    include_str!("../static/apple-app-site-association");
+const APPLE_APP_SITE_ASSOCIATION: &str = include_str!("../static/apple-app-site-association");
 
 pub fn static_routes() -> impl Filter<Extract = impl warp::Reply, Error = warp::Rejection> + Clone {
     open_route().or(well_known_route())


server/static/apple-app-site-association
Parth, line 1: nit: i think `etc` -> `static`

server/src/main.rs
```diff
@@ -52,7 +52,8 @@ async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
         .or(build_info())
         .or(stripe_webhooks(&server_state))
         .or(google_play_notification_webhooks(&server_state))
-        .or(app_store_notification_webhooks(&server_state));
+        .or(app_store_notification_webhooks(&server_state))
+        .or(files_routes());
```
Parth: given the nature of our business, and the existing use of `File` I think for this sort of idea you should prefer the word `static`

server/templates/redirect.html
```diff
@@ -0,0 +1,13 @@
...
+<html lang="en">
+    <head>
+        <meta charset="UTF-8" />
+        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+        <title>Open in Lockbook</title>
+    </head>
+    <body>
+        <h1>Unable to open your file</h1>
+        <p>Click the link below to open it in Lockbook:</p>
+        <a id="openAppLink" href="lb://{{ uuid }}">Open in App</a>
```
Parth: can we make this so it does not require human interaction? On chrome iiuc `lb://` will open another system dialog asking for confirmation. So it's 2 very similar clicks I have to deal with

@tvanderstad is going to add links to things like headers will this seamlessly support that? I imagine so uuid will just also contain a path, no one seems to be parsing it

server/src/router_service.rs
```diff
@@ -467,6 +472,42 @@ where
         })
 }
 
+static EXTERNAL_LINK_ROUTE: &str = "files";
+const APPLE_APP_SITE_ASSOCIATION: &str =
+    include_str!("../etc/well-known/apple-app-site-association");
+
+pub fn files_routes() -> impl Filter<Extract = impl warp::Reply, Error = warp::Rejection> + Clone {
+    let open_route = warp::path(EXTERNAL_LINK_ROUTE)
+        .and(warp::path("open"))
```
Parth: can we just have open, kinda regardless of subdomain it'll be better imo

server/templates/redirect.html
Parth: also, maybe this will get more sophisticated later, but you could skip adding the dependency and just stick this in a giant `format!` statement like we do in releaser for the basic stuff. 

even the more sophisticated version, I'm not sure we're going to need a template engine in the mix

server/src/router_service.rs
```diff
@@ -467,6 +472,42 @@ where
         })
 }
 
+static EXTERNAL_LINK_ROUTE: &str = "files";
+const APPLE_APP_SITE_ASSOCIATION: &str =
+    include_str!("../etc/well-known/apple-app-site-association");
+
+pub fn files_routes() -> impl Filter<Extract = impl warp::Reply, Error = warp::Rejection> + Clone {
```
Parth: I think you can stick everything related to this feature in a single file in it's own module

Parth:
Okay @smailbarkouch our api is now additionally reachable via `app.lockbook.net` and I think the links people see should be `https://app.lockbook.net/open/{uuid}`. [Community is on board](https://discord.com/channels/1014184997751619664/1368631361321177108/1486046621929046119).

If you're curious about some of the infra stuff taken care off see this thread: https://discord.com/channels/1014184997751619664/1486050098176594091/1486050210756165632

server/static/apple-app-site-association
smailbarkouch: `etc` is an existing directory

Parth: The files in that directory aren't served out

server/src/main.rs
smailbarkouch: I believe static is a reserved keyword

Parth: oh sad, maybe some other web-dev-adjacent word for "assets" or "public" or something, same with the `etc` comment. Etc is like config stuff like `/etc` on linux (that's literally where those files get copied).

server/src/static_files.rs
```diff
@@ -0,0 +1,64 @@
...
+const APPLE_APP_SITE_ASSOCIATION: &str = include_str!("../static/apple-app-site-association");
+
+pub fn static_routes() -> impl Filter<Extract = impl warp::Reply, Error = warp::Rejection> + Clone {
+    open_route().or(well_known_route())
+}
+
+fn open_route() -> impl Filter<Extract = impl warp::Reply, Error = warp::Rejection> + Clone {
+    warp::path("open")
+        .and(warp::path::param::<String>())
+        .map(|uuid: String| {
```
Parth, line 13: if you're not gonna parse this i would just call it a path or a route or something

server/src/static_files.rs
```diff
@@ -0,0 +1,64 @@
...
+            let span = span!(Level::INFO, "matched_request", method = "GET", route = "/open");
```
Parth, line 14: this is good, there's a few other things we probably want to extract like IP, try to be uniform with the stuff in the big and unfortunate macro

server/src/static_files.rs
```diff
@@ -0,0 +1,64 @@
...
+                // Try to open the app
+                window.location.href = url;
+            }};
+        </script>
+    </head>
+    <body>
+        <h1>Opening Lockbook...</h1>
+        <p>If nothing happens, use the link below:</p>
+        <div>
+            <a href="lb://{uuid}">Open in App</a>
```
Parth, line 57: perf

PR #4295, opened 2026-03-22
md: open all non-binary files in plaintext mode

tvanderstad:
`content_inspector` is super simple, I read the whole codebase before adding it


Parth:
Nice

PR #4300, opened 2026-03-24
feat(ios): Auto-titling (smart paste) for URLs in markdown editor

Luca409:
When a user pastes a bare HTTP/HTTPS URL, the app fetches the page title and replaces it with [Title](URL) markdown.

- Add workspace_replace_literal FFI for replacing last occurrence
- Add URLTitleFetcher module: parseBareURL, extractTitleFromHTML, cleanTitle, formatMarkdownLink, decodeHTMLEntities
- Use crawler User-Agents for YouTube/X/Twitter meta tags
- Integrate with paste and insertText flows
- Add 24 unit tests for URLTitleFetcher

Made-with: Cursor

https://github.com/user-attachments/assets/9be42010-64cc-4303-a07a-b9f53f670ceb


docs/building/apple.md
```diff
@@ -32,4 +32,12 @@ XCode needs to be able to do these things before `lbdev` will start to work.
...
+`lbdev apple run ios <device>`  allows you to select an iOS device and send a build to it. This is mostly a shortcut for people who have setup these devices already in XCode.
+
+## Can't run in Xcode?
+
+1. **Run `lbdev apple ws all` first** – Xcode needs the native libs. From repo root: `cargo install --path utils/lbdev` then `lbdev apple ws all`.
+2. **Open the workspace** – Open `clients/apple/Lockbook.xcworkspace` (not `lockbook.xcodeproj`).
+3. **Signing** – Xcode → Signing & Capabilities → set Team to your Apple ID. The project uses team `39ZS78S25U`; if you're not on that team, change it to yours or use "Sign to Run Locally" for simulators.
+4. **Simulator destination** – For iOS, pick a simulator in the scheme selector (e.g. "iPhone 16" or "iPhone 16e"). If "iPhone 16" fails, try another like "iPhone 16e" or "iPhone 17".
+5. **macOS code signing** – If macOS fails with "No signing certificate", either add your "Mac Development" cert in Keychain, or build with: `xcodebuild ... CODE_SIGN_IDENTITY="-" build`.
```
Parth, line 43: This is valuable and I'm just going to merge these changes directly to the docsite

libs/content/workspace/src/tab/markdown_editor/input/events.rs
```diff
@@ -19,7 +20,12 @@ impl<'ast> Editor {
     pub fn process_events(&mut self, ctx: &Context, root: &'ast AstNode<'ast>) -> Response {
         let mut ops = Vec::new();
         let mut response = buffer::Response::default();
-        for event in mem::take(&mut self.event.internal_events) {
+        let mut events: Vec<Event> = mem::take(&mut self.event.internal_events);
+        for (from, to) in url_title_fetcher::try_recv_pending_replacements() {
+            events.push(Event::ReplaceLastOccurrence { from, to });
```
Parth: This event gives me bad vibes but I will keep reading

libs/content/workspace/src/url_title_fetcher.rs
```diff
@@ -0,0 +1,280 @@
...
+use std::sync::mpsc;
+use std::thread;
+use url::Url;
+
+const CRAWLER_USER_AGENTS: &[&str] = &[
+    "facebookexternalhit/1.1 (+http://www.facebook.com/externalhit_uatext.php)",
+    "Mozilla/5.0 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)",
+];
+
+static REPLACEMENT_CHANNEL: std::sync::OnceLock<(mpsc::Sender<(String, String)>, std::sync::Mutex<mpsc::Receiver<(String, String)>>)> =
```
Parth: I personally use `use` statements a bit more aggressively to see fewer :s

libs/content/workspace/src/url_title_fetcher.rs
```diff
@@ -0,0 +1,280 @@
+//! Auto-titling for pasted/inserted bare URLs: parses URLs, fetches page titles, extracts from HTML,
+//! and replaces with [Title](url) markdown links. Works on all platforms (iOS, macOS, Linux, Windows, Android).
+
+use regex::Regex;
```
Parth: this would be the first production use of regex, wonder if avoidable

libs/content/workspace/src/url_title_fetcher.rs
```diff
@@ -0,0 +1,280 @@
...
+        if !title.is_empty() {
+            let lower = title.to_lowercase();
+            let is_generic = GENERIC_TITLES
+                .iter()
+                .any(|g| lower.starts_with(g) || lower == *g);
+            if !is_generic {
+                return Some(clean_title(&decode_html_entities(&title)));
+            }
+        }
+    }
```
Parth: how married to the automatic trimming are you? 

personally I would probably just grab an xml parser (we probably already use one somewhere in workspace because of svg support)

and I would just grab the title tag and not really parse it

my personal ideal behavior as well, I can trim if needed but even still for a title I would like further disambiguation

potentially I wonder if @tvanderstad would prefer this sort of cleverness to happen within the `![]` image tag where we take a bit more liberty with behavior. I wonder if we could render the social media preview that happens (more rich content) if someone like image links to something that is not an image.

similarly I think you can side step a few security concerns if you instead put this workflow behind it's own keystroke and or it's own UI element. 

so maybe you paste a link with no text and a little ✨ emoji is there as an option for you to just go and figure out what the link is

possibly it could get it's own markdown toolbar entry or something.

Parth: also related #4301

libs/content/workspace/src/url_title_fetcher.rs
```diff
@@ -0,0 +1,280 @@
...
+    } else if let Some(n) = name {
+        ("name", n)
+    } else {
+        return None;
+    };
+    let escaped = regex::escape(attr_value);
+    let patterns = [
+        format!(r#"<meta[^>]+{attr}=["']{escaped}["'][^>]+content=["']([^"']*)["']"#),
+        format!(r#"<meta[^>]+content=["']([^"']*)["'][^>]+{attr}=["']{escaped}["']"#),
+    ];
```
Parth: I don't want to understand regex well enough to review this

libs/content/workspace/src/url_title_fetcher.rs
```diff
@@ -0,0 +1,280 @@
...
+        ("&lt;", "<"),
+        ("&gt;", ">"),
+        ("&quot;", "\""),
+        ("&#39;", "'"),
+        ("&apos;", "'"),
+        ("&nbsp;", "\u{00A0}"),
+    ];
+    for (entity, replacement) in replacements {
+        result = result.replace(entity, replacement);
+    }
```
Parth: is this the same thing as what we use the `urlencoding` crate for? consider using a similar strategy unless I'm mixing up concepts.

libs/content/workspace/src/url_title_fetcher.rs
```diff
@@ -0,0 +1,280 @@
...
+    result
+}
+
+/// Formats title and URL as [title](url), escaping brackets in the title.
+pub fn format_markdown_link(title: &str, url: &str) -> String {
+    let escaped = title
+        .replace('\\', "\\\\")
+        .replace(']', "\\]");
+    format!("[{escaped}]({url})")
```
Parth: consider handing a mutable string around, t prob needs to weigh in on if theres a better way to yeet markdown in

libs/content/workspace/src/url_title_fetcher.rs
```diff
@@ -0,0 +1,280 @@
...
+}
+
+#[cfg(not(target_family = "wasm"))]
```
Parth: I believe the blocking client is safe on wasm, was there a specific reason you've gated this off wasm?

libs/content/workspace/src/url_title_fetcher.rs
```diff
@@ -0,0 +1,280 @@
...
+fn fetch_page_title_blocking(url: &Url) -> Option<String> {
+    let client = reqwest::blocking::Client::builder()
+        .timeout(std::time::Duration::from_secs(5))
+        .build()
+        .ok()?;
+
+    for user_agent in CRAWLER_USER_AGENTS {
+        let res = client
+            .get(url.as_str())
+            .header("User-Agent", *user_agent)
```
Parth: were the results without spoofing the user agent bad? 

we do something similar in the mind map part of our codebase, which could use some love from someone looking for a good time. 

and we do this there, not sure the quality of results

feels kinda weird to do, but maybe increases privacy. @tvanderstad maybe we should be doing something similar for images

libs/content/workspace/src/url_title_fetcher.rs
```diff
@@ -0,0 +1,280 @@
...
+            }
+        }
+    }
+    None
+}
+
+/// Spawns a background thread to fetch the page title and enqueue a replacement. When the fetch
+/// completes, (from_text, markdown_link) is sent to the channel for the main thread to process.
+/// Only works on non-wasm targets.
+#[cfg(not(target_family = "wasm"))]
```
Parth: I see, if I've gotten the blocking / nonblocking thing mixed up with wasm then consider structuring the code using a module which is entirely disabled on wasm and has a single no op function in the non-wasm case. It willl significantly reduce the amount of these cfg statements

Parth (changes requested):
prob want @tvanderstad's thoughts as well

libs/content/workspace/src/url_title_fetcher.rs
Luca409: What's wrong with regex?

libs/content/workspace/src/tab/markdown_editor/input/events.rs
Luca409: **Changes:** Renamed to `ReplacePastedUrlTitle { pasted_url, markdown }` so the event reads as domain-specific rather than a generic string hack. Expanded the doc on `Event` to state why **last match** is used: the fetch is async and we only have the pasted substring as an anchor; last occurrence is the usual case for the paste that triggered the fetch, and duplicates are rare.

**Disagreement / limits:** A stronger design is to store a stable range or request id at paste time; that is more invasive (edits between paste and response). Keeping substring + last match is a deliberate tradeoff until someone wants to invest in range-based replacement.

libs/content/workspace/src/url_title_fetcher.rs
Luca409: **Change:** `Mutex`, `OnceLock`, and `mpsc` are imported at module scope; a `ReplacementPair` type alias avoids the `OnceLock<...>` angle-bracket parsing issue and shortens the channel types.

libs/content/workspace/src/url_title_fetcher.rs
Luca409: **Change:** Dropped `regex` from this module entirely. Title and meta extraction now use **`scraper`** (HTML tree + CSS selectors), aligned with the mind map URL title path in `mind_map/data.rs`.

**Note:** The `workspace` crate already depended on `regex` for syntect and other uses; the concern was still valid for keeping *new* parsing logic off ad hoc regex.

libs/content/workspace/src/url_title_fetcher.rs
Luca409: **Changes:**

- Parsing goes through **scraper** / html5ever rather than regex on raw tags.
- Removed **site-specific suffix stripping** (`clean_title` was trimming things like `" - YouTube"`). After decode, we only **trim whitespace**, so titles stay closer to what the page declares.

**Product Decision:** I think that a markdown toolbar entry is the right idea -- sort of like an action button.  You could frame it as a sort of "Lockbook AI" feature that always gets a special icon (think of the Gemini icon embedded all over Google now)

Luca409: Could also just make it "opt in" for now, not on by default and something in the settings

libs/content/workspace/src/url_title_fetcher.rs
Luca409: **Change:** Reimplemented as a **single `String` build** with `push` / `push_str` and a small match for escapes, instead of `replace` chains + `format!`. Still no dependency on buffer APIs until Travis prefers a shared markdown-insert helper.

libs/content/workspace/src/url_title_fetcher.rs
Luca409: **Change:** Fetch logic lives in **`url_title_fetcher/native.rs`**, gated with `#[cfg(not(target_family = "wasm"))]`. The parent module exposes a **wasm no-op** `spawn_fetch_and_replace`.

**Reasoning:** Our **wasm `reqwest` dependency does not enable the `blocking` feature** (see `workspace/Cargo.toml`). `reqwest::blocking` is effectively a **native-only** API in this setup, so the old `#[cfg]` was required for **compilation**, not just policy. Documented on the `native` module.

libs/content/workspace/src/url_title_fetcher.rs
Luca409: **Change:** Comment in `native.rs` explains that many sites serve **minimal HTML to generic browsers** but richer **Open Graph** markup to **link-preview / crawler** clients, and points to **`mind_map/data.rs`** (desktop Chrome UA) as the parallel in-tree behavior.

We did not A/B test default UA vs crawler UA in this pass; keeping the bot rotation is justified for YouTube/X-style pages. Unifying mind map + editor fetch policy could be a follow-up.

libs/content/workspace/src/url_title_fetcher.rs
Luca409: **Change:** Implemented as **`mod native`** + `pub use native::spawn_fetch_and_replace` on native targets and a **single** wasm stub in the parent module. Shared parsing stays in the parent so tests and wasm builds still compile without pulling in blocking HTTP.

libs/content/workspace/src/url_title_fetcher.rs
Luca409: **Change:** Comment on `decode_html_entities` clarifies that **`urlencoding` is for percent-encoding in URLs** (`%20`, etc.), not HTML named/numeric character references (`&amp;`, `&#39;`, …). They are different layers; we should not replace entity decoding with `urlencoding`.

docs/building/apple.md
tvanderstad: #4 seems kind of unhelpful to me

libs/content/workspace/src/tab/markdown_editor/input/canonical.rs
tvanderstad, line 106: This is where we already have code that operates when you paste a link, making this the ideal location to hook into for your feature

libs/content/workspace/src/tab/markdown_editor/input/canonical.rs
```diff
@@ -111,13 +111,21 @@ impl<'ast> Editor {
...
                 }
             }
             egui::Event::Text(text) => {
                 if self.readonly {
                     return None;
                 }
+                if let Some(url) = crate::url_title_fetcher::parse_bare_url(&text) {
+                    let trimmed = text.trim().to_string();
+                    crate::url_title_fetcher::spawn_fetch_and_replace(trimmed, url);
+                }
```
tvanderstad, line 128: You don't want to respond to text events, which are what occurs when the user is physically typing

libs/content/workspace/src/tab/markdown_editor/input/mutation.rs
```diff
@@ -44,6 +44,17 @@ impl<'ast> Editor {
...
+                let text = self.buffer.current.text.as_str();
+                if let Some(byte_start) = text.rfind(&pasted_url) {
+                    let char_start = text[..byte_start].chars().count();
+                    let char_end = char_start + pasted_url.chars().count();
+                    let range = (DocCharOffset(char_start), DocCharOffset(char_end));
+                    operations.push(Operation::Replace(Replace { range, text: markdown.clone() }));
+                    let end_offset = DocCharOffset(char_end);
+                    operations.push(Operation::Select((end_offset, end_offset)));
+                }
+            }
```
tvanderstad, line 57: This looks for the first instance of the URL and replaces it, which could replace other text in the user's document. Ideally we just replace the (empty) title we inserted along with the markdown link we generated

libs/content/workspace/src/url_title_fetcher/mod.rs
```diff
@@ -0,0 +1,243 @@
...
+use std::sync::{Mutex, OnceLock};
+
+use scraper::{Html, Selector};
+use url::Url;
+
+type ReplacementPair = (String, String);
+static REPLACEMENT_CHANNEL: OnceLock<(
+    mpsc::Sender<ReplacementPair>,
+    Mutex<mpsc::Receiver<ReplacementPair>>,
+)> = OnceLock::new();
```
tvanderstad, line 17: This being static is going to prevent us from using the editor in other parts of the app, like search results, hover previews, canvas text boxes, and for transclusion

libs/content/workspace/src/tab/markdown_editor/input/canonical.rs
tvanderstad: It already checks for bare links, for example, and it uses exactly the algorithm the editor already uses to detect bare links via the parser

libs/content/workspace/src/url_title_fetcher/mod.rs
```diff
@@ -0,0 +1,243 @@
...
+    if scheme != "http" && scheme != "https" {
+        return None;
+    }
+    Some(url)
+}
+
+/// Extracts the title from HTML: prefers `<title>`, falls back to `og:title` and `twitter:title` for SPAs.
+fn extract_title_from_html(html: &str) -> Option<String> {
+    const GENERIC_TITLES: &[&str] =
+        &["youtube", "twitter", "x.com", "instagram", "facebook", "t.co"];
```
tvanderstad, line 49: Can you say more about why this is here? What happens when it's not?

libs/content/workspace/src/url_title_fetcher/mod.rs
```diff
@@ -0,0 +1,243 @@
...
+        ("&quot;", "\""),
+        ("&#39;", "'"),
+        ("&apos;", "'"),
+        ("&nbsp;", "\u{00A0}"),
+    ];
+    for (entity, replacement) in replacements {
+        result = result.replace(entity, replacement);
+    }
+    result
+}
```
tvanderstad, line 118: Can you say more about why this is here? What happens when it's not?

libs/content/workspace/src/url_title_fetcher/mod.rs
```diff
@@ -0,0 +1,243 @@
...
+            '\\' => s.push_str("\\\\"),
+            ']' => s.push_str("\\]"),
+            c => s.push(c),
+        }
+    }
+    s.push_str("](");
+    s.push_str(url);
+    s.push(')');
+    s
+}
```
tvanderstad, line 135: This definitely duplicates logic we already have (pasting a bare link, cmd+k or toolbar button to insert a link)

libs/content/workspace/src/url_title_fetcher/native.rs
```diff
@@ -0,0 +1,59 @@
...
+
+use super::{channel, extract_title_from_html, format_markdown_link};
+
+/// Crawler-style user agents: many sites (YouTube, X) serve minimal HTML to generic browsers but
+/// include `og:title` / richer markup for link-preview bots. [`crate::mind_map::data`] uses a
+/// desktop Chrome UA; we rotate bot UAs here because paste auto-title cares about those meta tags.
+const CRAWLER_USER_AGENTS: &[&str] = &[
+    "facebookexternalhit/1.1 (+http://www.facebook.com/externalhit_uatext.php)",
+    "Mozilla/5.0 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)",
+];
```
tvanderstad, line 19: Can you say more about why this is here? What happens when it's not?

libs/content/workspace/src/tab/markdown_editor/input/mod.rs
```diff
@@ -93,13 +93,39 @@ pub enum Region {
...
+        advance_cursor: bool,
+    }, // replace region with text and optionally advance cursor to end of new text
+    /// After async URL title fetch: replace the last occurrence of `pasted_url` with `markdown`.
+    /// Last-match is intentional—the user may have typed or pasted the same URL earlier; we only
+    /// upgrade the paste that triggered the background fetch (still the trailing occurrence in the
+    /// common case).
+    ReplacePastedUrlTitle {
+        pasted_url: String,
+        markdown: String,
+    },
```
tvanderstad, line 111: I don't think we need a new input event for this; I think the existing Replace should be able to cover it

Parth:
@tvanderstad thoughts on this UX? https://github.com/lockbook/lockbook/pull/4300#discussion_r2978654065

tvanderstad:
> @tvanderstad thoughts on this UX? [#4300 (comment)](https://github.com/lockbook/lockbook/pull/4300#discussion_r2978654065)

I am down to automatically insert the title, seems nice and doesn't conflict with an existing UX. When you have text selected, that will be the title, otherwise we can fetch one. For now, feels equally secure to the image case (we automatically fetch images). That security concern isn't something that's resonated with any users I've talked to so far so we can figure out both issues later imo

tvanderstad:
I see how link previews (hover on desktop) or embeds (image-style links) would show the title anyway but we are far from that atm and it seems like Luca prioritizes this feature

tvanderstad:
closing in favor of https://github.com/lockbook/lockbook/pull/4302

docs/building/apple.md
Parth: https://lockbook.net/docs/building/apple.html#cant-run-in-xcode

Issue #4301, opened 2026-03-24
have a setting to ask before networking from a markdown doc


Parth:
- images
- links

PR #4302, opened 2026-03-24
md: show fetched link titles

tvanderstad:
Shows a fetched title for title-less `[]()` markdown links and `www.destination.com` autolinks.
* `../note.md` relative links and `lb://` internal links resolve the title as the note file name minus `.md` if present
* `[[]]` wikilinks and `` image links are unaffected
* supports all platforms (including wasm, untested) & applies to existing notes
* url titles are computed asynchronously, show "Loading..." while loading, cached for the lifetime of the tab

https://github.com/user-attachments/assets/de733743-144a-4a3e-9d5e-892120a77829

### DIFF
--- a/libs/content/workspace/src/tab/markdown_editor/widget/inline/link.rs
+++ b/libs/content/workspace/src/tab/markdown_editor/widget/inline/link.rs
@@ -39,7 +39,7 @@ impl<'ast> Editor {
         let node_range = self.node_range(node);
         let selection = &self.buffer.current.selection;
         // auto links also reveal when cursor sits at a boundary, so backspacing
-        // from outside the link doesn't leave the user inside a captured title
+        // from the right side doesn't repeatedly replace the display text
         node_range.intersects(selection, is_auto)
     }
 


Parth:
Very cool and nice

Parth:
worth loosely associating with #4301 when we tackle that

I'm mostly concerned with tackling that in a post "shared with me" world

personally I think this feature does deserve it's own visual identity, even if it's just that the link color is slightly different or something to make it clear what's going on

libs/content/workspace/src/tab/markdown_editor/widget/inline/link.rs
```diff
@@ -295,3 +323,41 @@ fn extract_html_title(html: &str) -> Option<String> {
         .filter(|t| !t.is_empty())?;
     Some(title)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
```
Parth: ultimately would like `#[ignore]` on these tests

tvanderstad: would like to discard them, thanks

Parth:
> personally I think this feature does deserve it's own visual identity, even if it's just that the link color is slightly different or something to make it clear what's going on

PR #4305, opened 2026-03-24
infra: update the deploy-flatpak.yml file to automate PRs

Amumug:


.github/workflows/deploy-flatpak.yml
```diff
@@ -10,12 +10,17 @@ jobs:
...
         run: |
-          git config --global user.name "github-actions[bot]"
-          git config --global user.email "github-actions[bot]@users.noreply.github.com"
-          nix-shell -p curl python3 python313Packages.aiohttp python313Packages.tomlkit --command "cargo run -p lbdev -- releaser linux desktop flatpak"
+          gh pr create \
+            --title "New Release ${{ github.ref_name }}" \
+            --body "Automated release PR " \
+            --base master \
+            --head ${{ github.ref_name }} \
+            --repo flathub/net.lockbook.Lockbook
```
Parth: you have added this as a step 2 after the `lbdev deploy flatpak` command.

I think this is a bad idea because I am pretty sure you will not have the correct value for `--head`

but instead if you leave this file alone and do this in rust at [this-location](https://github.com/lockbook/lockbook/blob/master/utils/lbdev/src/releaser/linux/desktop.rs#L434). 

you'll have access to the right branch name.

Parth (changes requested):

Parth:
also this is not ignorable 
image

Amumug:
now?

utils/lbdev/src/releaser/linux/desktop.rs
```diff
@@ -432,5 +432,22 @@ pub fn push_flatpak(version: &str, flatpak_repo: &str) -> CliResult<()> {
         .args(["push", "origin", "master"])
         .current_dir(flatpak_repo)
         .assert_success()?;
+    Command::new("gh")
+        .args([
+            "pr",
+            "create",
+            "--title",
+            &format!("New Release {}", version),
```
Parth:
```suggestion
            &format!("update to {}", version),
```

Parth: I'll test this you can address in the following PR

.github/workflows/deploy-flatpak.yml
Parth: This is still unresolved

Parth: as in you have to delete the old code you are creating a PR twice

Issue #3548, opened 2025-05-20
advanced search


Parth:
will fill this out later

Issue #2912, opened 2024-10-08
Top 3


Parth:

Parth:
@tvanderstad:
1. #2713
1. canvas diagramming #2847
1. #190

**Completed:**
* #2544 
* #2604 
* #2882
* #3463
* #2732
* #4844 
* #4038

Parth:
@Parth:

1. #1939 
2. #2847 
3. #2908

Parth:
@MehrotraRahul:
1. #2848 
2. #1301 
3. #2505

steverusso:
1. https://github.com/lockbook/lockbook/issues/4779
2. https://github.com/lockbook/lockbook/issues/190
3. https://github.com/lockbook/lockbook/issues/2940

**Completed:**
+ https://github.com/lockbook/lockbook/issues/2337
+ https://github.com/lockbook/lockbook/issues/1918

tvanderstad:
My dad's:
1. Transfer notes from OneNote
2. Polish Windows client

BuriedInTheGround:
1. Improve editor UX (in order of importance/preference: #2938; #1894; #2732; #2025)
2. #2937 
3. #2607

smailbarkouch:
@smailbarkouch 
1. https://github.com/lockbook/lockbook/issues/2847
2. https://github.com/lockbook/lockbook/issues/2732
3. https://github.com/lockbook/lockbook/issues/2544

ad-tra:
@ad-tra 
1. preserve tab history across sessions 
2. better search: #2899 #2431
3. canvas: #2850 and #2852

Parth:
@CoreyCole's
1. https://github.com/lockbook/lockbook/issues/3975
2. probably #2775 (for mentions) [src](https://discord.com/channels/1014184997751619664/1437939209158656211/1438004818638540903)

Parth:
@raayan's

1. https://github.com/lockbook/lockbook/issues/3988
2. https://github.com/lockbook/lockbook/issues/3987
3. https://github.com/lockbook/lockbook/issues/3989

paulhovey:
For whatever reference, as of 06/05/2026:
1. https://github.com/lockbook/lockbook/issues/3366
2. https://github.com/lockbook/lockbook/issues/3628
3. Sharing updates (largely https://github.com/lockbook/lockbook/issues/2775 , although I don't care about document history, but some type of notification for shared items and a "Shared by me" would be nice)

[...]

PR #4312, opened 2026-03-27
all: new search

Parth:
problem:
image

solution:
make search not mid


Parth:
Let's take stock of the quality of search here:

# Path Search Quality

adam has a file here called tech-debt

image

and what's most intuitive for him is to use a space to quickly delimit his search, old search cannot fathom this behavior

image

but new search uses [nucleo](https://github.com/helix-editor/nucleo) which is the fuzzy funder that powers [helix](https://github.com/helix-editor/nucleo/) (a very telescope like experience, the sort of experience multiple users have asked us to take inspiration from. And nucleo has lots of rules coming from a path centric search environment that work really nicely.

image

old path search would also take a lot of characters to match documents, even though i was typing the name of the file (very common workflow), I added bonuses for this sort of situation but the code is just too naively implemented to work well in practice. Without the help of suggested docs I have to type to get `huddle.md`:

image

but new search, doesn't even need the help of suggested docs to behave the way I want:

image

I'll leave one more example of a flow I do very often and never have it behave exactly the way I expect:

before:
image
after:
image

# Content Search Quality

There are two fundamental issues with our use of tantivy, arguably one of these is a skill issue. Let's start with the one that's not a skill issue:

Tantivy doesn't support substring search, which means you don't get search results until you've finished typing the word you're looking for. So luca gets practically no search results until you're done forming a word:

image

new content search is a handrolled implementation that behaves closer to the behavior of other popular note taking apps (deprecates tantivy): 

image

it performs a substring search across all text documents. The query is split by space into search terms. Each term must be mached at-least once in the document for it to appear in the list, and any documents that have exact matches are grouped above documents that have partial matches.

image

There's a lot of tiny nuisances caused by this the main tradeoff is that the search executes very quickly. But my handrolled-unoptimized-more-flexible implementation is executing these queries in less than 5ms. I could probably make this much faster as well if needed, the workload is embarrassingly parallel, additionally I could incrementally process the inputs. Finally if that's not enough there are some interesting algorithms in this space. But I'm not too concerned about that so far.

The other category of badness is sorta a double edged sword:

image

I'm getting no results here because of tantivy's support of advanced query constructs. 

image

image

We could wrap each search term in `""`, but then we'd be giving up access to tantivy's advanced search (#3548). Overall feels like we have a vibe mismatch with our goals and tantivy's. If I think about what I personally want out of an advanced search experience, it's the ability to use `grep` expressions and these could be handled directly by the current content search implementation as a flag, similar to what travis is showing with find in doc. #4426. 

Another request is for people who have long documents, they don't just want the search result to find "the doc". Raayan has a wine.md and search returns this single doc as the result, he wants it to show all the results (like telescope does) and be able to jump to them. Presumably #4426 gets him part of the way there, but imagine a user with a few large docs, they still want content results that are more usable for search terms that occur frequently inside a document. So as you can see in these screenshots we're aware of all the matches, and we'll show them in a sorta tree hierarchy. See tuned for ui beautification I alude to below.

Lastly this search is structured in a way to have other searchable modules alongside, potentially sharing infrastructure like the in memory cache of documents. There's room here for specifically for in memory semantic search, a command palette and other ideas.

You may notice these things are a little ugly right now, I spent most of my effort so far on data quality, and I'm personally satisfied. Feel free to comment on this PR with other situations you'd like me to check. 

Now I'm going to move onto a beautification stage. The UI is ready for contribution from @tvanderstad to embed the editor, I'm going to make the query and results panel much prettier at the same time.

On mobile there's two paths we could take. We could allow the input to be native, and present the search results in egui. Or I could firm up the data contract and do it all natively. The search performance is quite good and I'm tempted to backport this to lb-rs now, but I think fixing the rest of the UI will give me more information regardless.

Parth:
image

featuring bolded sections, now going to have a "selection"

Parth:
better icons 
image

Parth:
image

I like how this is shaping up

Parth:
@ad-tra has convinced me that the search implementation should in fact be native, so I'm going to move this over into lb-rs and work on a generalizable contract. I think macOS will have the same egui-like experience, but iOS will get some special attention (from me).

Parth:
note to self: handle the folder activation situation properly

Parth:
prob: image

clients/apple/macOS/Screens/HomeView.swift
```diff
@@ -19,27 +19,35 @@ struct HomeView: View {
...
-                        }
-                    })
-                    .navigationSplitViewColumnWidth(min: 250, ideal: 300)
-                },
-                detail: {
-                    NavigationStack {
-                        DetailView()
-                            .modifier(OutOfSpaceAlert())
+        // Search is handled by the workspace (egui) on macOS, so the Swift
+        // search containers that used to wrap this view have been removed.
```
Parth:
```suggestion

```

clients/apple/Shared/Screens/PathSearchContainerView.swift
```diff
@@ -1,6 +1,10 @@
 import SwiftUI
 import SwiftWorkspace
 
+// The ⌘O path-search overlay is iOS-only now; macOS searches via the workspace
+// (egui) "Open Quickly" tab instead.
```
Parth:
```suggestion

```

libs/content/egui_wgpu_renderer/src/lib.rs
```diff
@@ -17,7 +17,7 @@ pub use wgpu;
 
 mod shame;
 
-const NOOB_FACTOR: u64 = 7;
+const NOOB_FACTOR: u64 = 20;
```
Parth:
```suggestion
const NOOB_FACTOR: u64 = 7;
```

libs/content/workspace-ffi/SwiftWorkspace/Sources/Workspace/Lb/Search.swift
```diff
@@ -1,157 +1,170 @@
...
+    public let matched: String
+    public let suffix: String
+}
 
-        for match in self {
-            matches.append(ContentMatch(match))
-        }
+/// Thin wrapper over the Rust `ContentSearcher`. Not thread-safe on its own — callers must
+/// serialize access (e.g. via a dedicated dispatch queue). Calling `query` and `snippet`
+/// concurrently from different threads is undefined behavior.
```
Parth:
```suggestion

```

libs/content/workspace-ffi/SwiftWorkspace/Sources/Workspace/Lb/Search.swift
```diff
@@ -1,157 +1,170 @@
...
+        }
+        return SearcherSnippet(
+            prefix: Self.clean(String(cString: prefix)),
+            matched: Self.clean(String(cString: matched)),
+            suffix: Self.clean(String(cString: suffix))
+        )
     }
+
+    /// Flatten newlines so the snippet fits on a single line. Matches the cleanup the egui
+    /// search UI does in `workspace::search::content::extract_snippet`.
```
Parth:
```suggestion

```

libs/content/workspace/src/search/mod.rs
```diff
@@ -0,0 +1,345 @@
+pub mod content;
+pub mod path;
+
+/// The search experience, owned by its tab ([`crate::tab::TabContent::Search`]).
+/// The query box lives here; the results/preview split is rendered by
+/// [`Workspace::show_search_tab`] because the preview pane reuses the
+/// workspace's async file-loading (`self.preview` / `set_preview`).
```
Parth:
```suggestion
```

libs/content/workspace/src/tab/mod.rs
```diff
@@ -35,12 +36,15 @@ pub enum Destination {
     File(Uuid),
     MindMap(Uuid),
     SpaceInspector(Uuid),
+    /// Experimental: the search experience embedded as a tab (Cmd+T).
```
Parth:
```suggestion
```

libs/content/workspace/src/tab/mod.rs
```diff
@@ -220,6 +227,10 @@ impl Tab {
                     TabContent::SpaceInspector(sv) => {
                         sv.show(ui);
                     }
+                    // Rendered by `Workspace::show_search_tab` instead of here:
+                    // the results/preview split needs workspace-owned async file
+                    // loading that a `Tab` can't reach.
```
Parth:
```suggestion

```

Parth:
@tvanderstad from here until it's merged (estimated to be monday) is a good time for you to review the editor bits, and optionally any preview tab bits.

Parth:
# Search Update

The last update about search quality is still true. The remaining time on this PR was spent exploring different UX approaches that use this search engine while soliciting feedback from @tvanderstad & @ad-tra. As things have solidified I'm pretty happy with how search feels in various places.

On egui, search is now a new tab:

image

It can be accessed 3 ways:
- search icon in the file tree
- command (or ctrl) shift f to jump to content search
- command (or ctrl) o to jump to a path search

Previews as initially prototyped by @tvanderstad are in here and they work great. Accepting a search result from either flow will replace the tab with the already fetched content. The previews are especially nice for quickly evaluating a large number of images. 

When there is no query we show suggested docs, I believe this is what the future of suggested docs will become. It will not have a presence in the file trees. See future for more information.

content search looks something like this:
image

and the preview editor will jump to whatever snippet you're interested in looking at. You can also drill into a file if you want to focus on that file.

I prototyped a native version of this for macOS but ultimately the new-tab experience felt much more appropriate especially as the whole replace the current tab, quickly open transient tabs, discussion isn't really resolved. This is more obvious, search will open in a new tab, you can close that tab, and if you select a file it will replace the search tab. This seems the be the least controversial of the approaches.

iOS gets a native search experience:

image

you can click around these search results and it will keep opening them in a new tab. The search will stay open until you explicitly dismiss it, at which point it will take you back to the file tree focused on whatever the last accepted search result is. I believe the default in mobile land should be all tabs are replaced, but this is an orthogonal idea to search.

content searching also has a dedicated experience: 
image image

Looking forward to seeing what @ad-tra whips up on android!

# Future of search

This is scheduled to be merged on Monday Jun 8th. This will give internal testers early access. After testers are satisfied it will released (likely 4-10 days after merge). Regardless feedback is welcome at this point, either on this PR, the code, the screenshots etc.

Some thoughts about how I see this evolving over time:

- @tvanderstad and I share a vision that search and landing page will get merged and this will be your one point of contact to *do something* and we can show it in more places, and during onboarding, during first app startup we can take some more liberties with the experience.
- @ad-tra strongly advocated for an `all` search. Personally I think this would be hard to get right, so I didn't attempt it for this increment. Android leans a little more in the direction of an all search as you'll see. But next time I want to invest here I'll add a third search module for an all search and we can start to explore comparing scores across various search strategies.
- @ khalil-akremi (with @ad-tra's guidance) is working on semantic search which should help with a more open ended search strategy ("find documents with uncompleted todo list items", and hopefully one day "find the pdf or image that looks like my lease agreement" (multi modal search)). These search strategies should be able to use this infra structure easily.
- Another thing I'd like to invest in later is search controls. I'd like case-sensitivity to be a toggle (right now all search is case insensitive), and I'd like to be able to scope search. I'd also like to be able to do regex/grep style searches. These are all pretty easy ad-ons that can come in the medium term that continue to increase the power of the search experience, a critical component to people's second brain. Just take a look at #2912 to see how many people rank a better search experience as the most desired item.

tvanderstad:
congrats on merge 🎉

Issue #4324, opened 2026-03-30
shared with me problems


Parth:
+ [ ] relative links in shared with me sections don't work correctly
+ [ ] [creating a folder in shared with me crashes](https://discord.com/channels/@me/1434482120796930161/1488065738974564453)
+ [ ] cannot create a folder in macOS shared with me

Parth:
#4611

PR #4406, opened 2026-04-08
ws: links that support pending shares

### LINKED ISSUE #4325: read only mode regression?
#2349 appears to not be working properly with shared with me (maybe #4324)

Image

tvanderstad:
principles:
* pending shares are viewable, editable first-class citizens
* files in different pending shares are considered to be in separate trees
* a file can be shared directly and also via ancestor folder at the same time; outermost share is what counts for tree separation

changes:
* include pending shares in file cache helpers (new 'default behavior')
* if some collaborators of this file don't have access to a link destination, highlight in yellow (link is broken for someone else)
* complete links into/out of/across pending share trees using uuid's; outermost share determines tree separation
* show pending shares in the workspace file table if focused parent is root

sugar:
* cmd+w on the landing page now sets focus to root (no breadcrumb when in pending share root)
* egui file tree icon color system extended to landing page
* in link completion and landing page file name hover, shared file names in the file path are colored (helps disambiguate pending shares from files in root)

fixes an issue where pending shares aren't previewable
fixes a few pending share related crashes on the landing page (e.g. navigating into a pending share folder)
fixes #4325 
fixes first task item in https://github.com/lockbook/lockbook/issues/4324

### DIFF
--- a/libs/content/workspace/src/tab/markdown_editor/widget/inline/image/image.rs
+++ b/libs/content/workspace/src/tab/markdown_editor/widget/inline/image/image.rs
@@ -12,18 +12,13 @@ use lb_rs::model::text::offset_types::DocCharOffset;
 use crate::tab::markdown_editor::Editor;
 
 use crate::tab::markdown_editor::widget::inline::Response;
-use crate::tab::markdown_editor::widget::utils::wrap_layout::{Format, Wrap};
+use crate::tab::markdown_editor::widget::utils::wrap_layout::Wrap;
 use crate::theme::icons::Icon;
 use crate::theme::palette_v2::ThemeExt as _;
 
 use super::cache::ImageState;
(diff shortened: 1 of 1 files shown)


clients/egui/src/account/mod.rs
```diff
@@ -236,13 +236,11 @@ impl AccountScreen {
                 }
 
                 if let Some(file) = wso.selected_file {
-                    if !self.tree.selected.contains(&file) {
```
tvanderstad, line 239: This check was interfering with workspace's ability to set the focused parent and I'm not sure for what reason - all the focused parent workflows that I checked are still good. Without this change, I experienced this issue:
* click folder in file tree (both selects and expands)
* click same folder (collapses)
* landing page breadcrumb no longer moves you up a directory

Issue #4410, opened 2026-04-08
calculate rich usage statistics locally


ad-tra:
- [ ] calculate it locally, both unpushed changes as well as what the server has
- [ ] refresh it upon metadata change
- [ ] make the status field not be optional

Parth:
good guidance for someone interested in doing this here: https://github.com/lockbook/lockbook/issues/3454

PR #4489, opened 2026-04-19
lb & server: the usage api we deserve

Parth:
we calculate storage too naively in many situations, locally we just ask the server and that itself is wrong #1879 in a subtle way.

this pr gives us the usage api we deserve, with the tricky calculations living in our tree logic (do you own the file? Is it delete? metadata fee? etc). The server will calculate a `usage_report` for the per-tree impact of any change to fix the bug in the server. And this API will allow us to compute rich usage statistics locally (https://github.com/lockbook/lockbook/issues/4410) and visualize the impact of unpushed changes: 
image


Parth:
this will be deployed tomorrow

PR #4497, opened 2026-04-21
iOS: migrate lb-rs to app group location

Parth:
whatup @smailbarkouch this migrates lb-rs into the app group area which unlocks our ability to have widgets and other fun things I'd like to produce on the side, I tested that this is good in a variety of situations


Parth:
going to solve multiple lb-rs' next but want to give you a chance to at-least ack the existence of this. I believe it's going to be fine with share-sheets but I'll check on that at release time QA. Once we have multiple lb-rs' we'll also have the ability to do nicer ingress flows if we feel like it.

PR #4501, opened 2026-04-23
lb-rs: support multiple instances

### LINKED ISSUE #2152: multiple co-operative lb-rs instances

Parth:
fixes: #2152 

Thanks @Kousay-Jebir to your prototyping on our earlier attempts, the things we talked about in that PR helped inform this one!


Parth:
this is working pretty well, and I'm happy with the ergonomics of the situation too. I have two known problems:
- if I open the same tab across two windows it seems like the second one gets stuck in a perpetual re-read required. Presumably this is due to master instability.
- things like logging init'ing shouldn't happen twice, maybe they should never happen on the remote process

Parth:
My reason to assume this is a @tvanderstad regression is that I haven't touched master and this is passing all the integration tests, but I'll wait for clarification about the nature of the `master` data loss before I merge this. Though I am eager to merge this due to the size. This is also pretty much the smallest this PR could be.

Parth:
working nicely across two egui instances
https://github.com/user-attachments/assets/09b88ecf-42ad-46cf-a9c1-129cbecdc6c5

PR #4515, opened 2026-04-28
md: affine scroll

tvanderstad:
Replace the editor's `egui::ScrollArea` with a custom `AffineScrollArea` that uses approximations to avoid fully measuring off screen content.

On init, we measure the approximate height of all blocks in the document to size the scroll bar. The scroll offset is stored in approximate coordinates. Dragging the scroll bar navigates the scroll offset smoothly.

The first content on the screen is determined using the approximate heights of the blocks above (which are already computed for the scrollbar). The content is sized precisely starting at this point until the screen is full.

When you submit a scroll event, it is interpreted in precise coordinates. We examine the blocks in the distance spanned by the offset from an anchor offset on the screen and multiply piecewise by the approx/precise ratio to convert to an approx scroll amount, then add that to the scroll offset.

The gains depend on our implementation of block heights being fast enough to layout the whole document quickly and accurate enough to not cause observable funkiness - an under approximation for a block makes you scroll by it more quickly when dragging the scrollbar. Our approximation pretends, basically, that the block's text is monospace and character-wrapped with char width = .5 * char height (an empirical choice).

The result: a 30x speedup of cold-starting the editor to any scroll offset, targeting the search preview use case.

### DIFF
--- a/libs/content/workspace/src/tab/markdown_editor/input/cursor.rs
+++ b/libs/content/workspace/src/tab/markdown_editor/input/cursor.rs
@@ -186,14 +186,38 @@ impl MdEdit {
         }
     }
 
-    pub fn scroll_to_cursor(&self, ui: &mut Ui) {
-        let selection = self
+    pub fn scroll_to_cursor(&mut self, ui: &mut Ui, scroll_id: egui::Id, viewport_height: f32) {
+        let target = self
             .in_progress_selection
-            .unwrap_or(self.renderer.buffer.current.selection);
+            .unwrap_or(self.renderer.buffer.current.selection)
+            .1;
 
-        if let Some([top, bot]) = self.cursor_line(selection.1) {
-            let rect = Rect::from_min_max(top, bot);
-            ui.scroll_to_rect(rect.expand(rect.height()), None);
+        let arena = comrak::Arena::new();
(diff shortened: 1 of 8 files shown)


tvanderstad:
We went with a design where the top-level blocks of the editor are the atomic "row" unit to the scroll area - we never draw part of a top-level block. That means this optimization doesn't help when your document is one super long list, which seems reasonable to me.

The level up I experimented with recurses into the tree. This breaks apart blocks that are nested in top level blocks and helps with the long list case but creates complexity because ancestor nodes of what's on screen may want to paint decorations: think, rendering the second paragraph only inside a block quote. Once the first paragraph is off screen it becomes complex to detect that the second paragraph still needs the block quote bar drawn. We succeeded but the complexity didn't feel warranted to me.

The further level up I experimented divide the document into individual lines, bringing code editor level performance to plaintext docs as well as adversarial performance situations like a single long code block. This required factoring block decoration drawing to its own helper - think, when you can only see a few lines of a code block on screen, we draw the code block background as if these are all the lines contained in the code block - measuring the rest defeats the optimization. We succeeded but I didn't like the ergonomics impact to block widget authors.

tvanderstad:
other aspects that came up during the work: scroll persistence, scroll to cursor, touch scroll, scroll momentum

PR #4537, opened 2026-05-04
lb-rs: reduce cpu usage for background syncs

Parth:
knocks down two high spots within lb-rs's background sync operations:
- for beta users, we used to send debug_info for every sync. But now we only do if there's something new to report, or if we haven't sent one in a long time (-8% cpu usage)
- sync wants to "ensure a file exists" but often doesn't care about it's contents unless it's doing a merge operation. We were using a wasteful `maybe_get` which reads a files when it doesn't need to just to ensure they're available locally.

beyond this there's two more optimizations that we could pursue:
- we could stop hashing uuid's, spend a surprising amount of CPU on this, can't stop on the server, but can prob stop here.
- can cache a different category of decryption in memory.

these are well suited for flamegraphs produced while running the exhaustive syncer which I plan to restore soon, so I'll defer these for now.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,8 +23,8 @@ members = [
 ]
 
 # restore for big binaries but good backtraces
-# [profile.release]
-# debug = true
+[profile.release]
+debug = true
 # [profile.bench]
 # debug = true
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,8 +23,8 @@ members = [
 ]
 
 # restore for big binaries but good backtraces
-[profile.release]
-debug = true
+# [profile.release]
+# debug = true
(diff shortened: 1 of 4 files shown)


Parth:
the behavior prior to this was that in the background while the user was still "active" we would have a spike to 20% cpu usage every 5 secs for the first 5 minutes of them being active.

The optimizations reduces the cpu consumption for the first minute of the app being open by 50% (4.7k samples -> 2.2k samples).

checking to see how this shows up in xcode cpu now. Confirmed via Xcode, significant reduction in spiky usage, more can be done but later. The next high spot comes from our landing page -- mainly the `show_files` function, going to take a quick look at it:

image

PR #4544, opened 2026-05-05
windows, linux: fix many integration woes

### LINKED ISSUE #2267: windows frame trackpad zoom

### LINKED ISSUE #2353: windows save open files before closing

### LINKED ISSUE #2354: linux save open files before closing

### LINKED ISSUE #3318: alt modifier input broken on windows
may be some conflict with system reserved shortcuts or accessibility features

### LINKED ISSUE #3411: german keyboard layout on linux and windows
[this](https://github.com/lockbook/lockbook/pull/3400) change, but in more places, and ideally with factoring that would prevent this kind of thing from being an issue in the first place

### LINKED ISSUE #3568: linux does not recognize a caps lock that has been re-mapped to esc

### LINKED ISSUE #3578: linux integration log spam
linux integration is presently spamming

```
Failed to get Xft.dpi
```

from here: https://github.com/lockbook/lockbook/blob/d467adc076415b829966d5b91cff4c93066fdae4/clients/linux/src/window.rs#L243

we probably need a more sophisticated interaction here, I briefly glanced at instances of `Xft.dpi` in `winit` and it seems slightly more involved.

### LINKED ISSUE #3579: linux egui integration energy efficiency
presently we poll and wait 10ms for events on this integration

### LINKED ISSUE #3615: windows task bar & start menu icons broken
https://github.com/lockbook/lockbook/issues/3437#issuecomment-3082251208

reported/requested by @CoreyCole

### LINKED ISSUE #4545: task manager begin shutdown should wait for tabs to be saved before shutting returning from shutdown
basically we need a syncronous save_all_tabs

Parth:
This PR goes up a level of abstraction and creates our integration using winit instead of targeting x11 or win api directly. It's still an integration that we write, we still support our custom events for pasting images and dropping files. Still mostly the same shape, we just use winit instead, and so far this seems to clear out our entire backlog of integration problems on windows and linux, although more thorough testing needs to be completed.

- fixes #3578 
- fixes #3579
- fixes #3568
- fixes #3411
- fixes #2354
- fixes #2353
- fixes #3615
- fixes #3318
- fixes #2267
- fixes #4545 
- fixes lack of fractional scaling support
- fixes "dead keys" (press a modifier key on lockbook, switch windows, lift the key, lockbook thinks key is stuck)

For now this takes the old clients are just going to point themselves to this new client called *desktop*, this will mean our deployment scripts won't break. I'll update all the deployment scripts to point to this location and then we can delete all the old clients as well.


Parth:
external urls are opening twice

Issue #4481, opened 2026-04-19
Input laggy on desktop app v26.4.13


paulhovey:
I have used 26.4.13 on both Windows and Linux flatpak (thanks, Parth!) and the typing input is laggy.  Not android bad, but noticeable.

On Windows, the input lag was so bad that the keyboard strokes were getting confused.  I would type a parenthesis, and instead of seeing something like "(so this is quiet)" the output in lockbook would be more like "9SO thi sis quieT)".  Linux I didn't see the issue with the shift key, but typing was not instant like it is in every other app.

Parth:
also reported by @Amumug https://discord.com/channels/1014184997751619664/1026208854280777789/1489112676708782101

on `26.2.11` it's fine

Parth:
# wip

my attempt to repro: I go back to `26.2.11` and jiggle my mouse around and observe what's up in xcode. I'm getting 30% CPU and low energy consumption @ 120fps I also never see CPU exceed roughly 35%.

<img width="1512" height="982" alt="Image" src="https://github.com/user-attachments/assets/aefce723-4113-4ed6-97af-5126fc3a36a1" />

Now I'm going to test `master`: I observe a slightly higher cpu usage maybe 35%, and spikes to 55%

<img width="1512" height="982" alt="Image" src="https://github.com/user-attachments/assets/2d157834-ee0e-4432-b56c-dc074ce33d78" />

The issue is much more pronounced on egui where amumu reported it: 

<img width="1512" height="982" alt="Image" src="https://github.com/user-attachments/assets/9ee72635-932f-4f25-a555-23ef4300ae7c" />

able to reproduce high cpu usage just by opening a tiny markdown file and jiggling my mouse, seeing about 70% pretty consistently, which is similar to what amumu is reporting.

Parth:
We spent some time investigating things across the stack and found some areas of waste. Specifically what I think you're running into is #3579 and the problem exists on both linux and windows. We also identified some areas that we're being wasteful that's particularly noticeable for people with weaker CPUs. And we're going to work through those as well, but I won't mention that here because I'm pretty confident you're not running into that.

Parth:
@paulhovey can you give the linux client a spin from our CD pipeline and tell me if the text input is still laggy? In #4544 I fixed who we process events - eliminating a 10ms sleep during which if our frames were taking too long you could be observing say less than 50fps. 

[...]

Issue #4433, opened 2026-04-10
Android typing is laggy and crashes the app


paulhovey:
Android version: 26.4.9
Phone:  Pixel 8 Pro

I created a new note from the android app and started typing, and there is a noticeable lag of maybe around ~0.5s between keypresses and the character being shown on the screen.

Also, pressing and holding the backspace key crashes the app reproducibly for me.

paulhovey:
I just tested on the flatpak desktop image and typing is laggy there too, I'm guessing it's accentuated on android because of processing power?  Pressing and holding the backspace key on the desktop doesn't crash the app, but it's not a smooth effect, either.

Parth:
also reported by @Amumug https://discord.com/channels/1014184997751619664/1026208854280777789/1489112676708782101

Parth:
Paul confirmed this is fixed 🎉 though @ad-tra may have some unmerged android fixes that go along with this, @ad-tra can you associate these things when you get a chance

ad-tra:
everything i have is merged

Parth:
I believe this is the association I'm looking for if that's the case? https://github.com/lockbook/lockbook/pull/4564

PR #4557, opened 2026-05-10
faster lb-rs, background file-cache, and shaming of devs

Parth:
This PR tackles a few problems that are triggering #4481:
- @paulhovey's logs painted a clear picture on android and linux that `list_metadata` was taking a long time. It also seemed like we were doing this computation on the `ui` thread causing serious stutters in input. It seems like it was also playing a role in #4433, although it seems like there was an independent set of improvements that @ad-tra needed to make to reveal that to be the problem (crashing was not caused by this for instance).
- So I dug into flamegraphs and what dominated our CPU time was hashing uuids. This is not unexpected as our previous perf inquiry left off with this knowledge in mind (#3231). But now it's time to sort this out. UUID's especially the ones generated by clients themselves have sufficient uniform randomness that we don't need to burn CPU cycles doing a cryptographically secure hash.
- Once this was sorted out the remaining CPU time is spent doing aes decryption of the names in `File`, which is pretty essential, but I wondered what the deal was with `soft::Aes`, we're also due for a version bump on this front so I dug into the details. A more recent version of aes will perform hardware aes specific cpu instructions, and we can opt into this behavior on arm as well. This resulted in another round of speedups. I imagine the results will be much more dramatic for how long it takes to read a document.
- Lastly I expanded the use of @tvanderstad's `FileCache` making it so basically all the rust code paths only `list_metadata` as infrequently as possible. And I also backgrounded this computation, despite the speedup I don't think there's any reason to block the UI thread while it is being computed.
- This PR also now crashes the app if the user is @tvanderstad, @ad-tra, or myself if a frame takes longer than 4x 120fps. We call this the noob factor and want to get this down to 1 which is a single frame budget of 8ms.

Going to seek validation from a super old samsung device if this fixed our problem, as well as reach out to Paul once we sanity check on our end.


.cargo/config.toml
```diff
@@ -1,2 +1,5 @@
 [target.x86_64-pc-windows-msvc]
 rustflags = ["-C", "target-feature=+crt-static"]
+
+[target.'cfg(target_arch = "aarch64")']
```
Parth, line 4: also checkout android

Parth: android is covered by this

PR #4576, opened 2026-05-18
md: text layout increment

tvanderstad:
This PR improves line wrapping:
* fixes release-blocking regression where one leading syntax character could get its own line for no reason
* line breaks prefer whitespace, then punctuation (like at hyphens or `/`'s in a link), then individual characters
* wrap logic supports (unimplemented) inline images, folded section placeholders, inline math - really any content, which can be interactive and space will be created for any size

This is accomplished by shaping instances of wrapped text as a whole unit before painting. Improved information at layout time enables more sophisticated controls.

The diff adds about as many lines as it removes but do not mistake it for a re-arrangement of existing code. 80% of additions are the new wrap layout logic with improved behavior and support for upcoming work. The deletions largely result from a unification of layout and paint logic for inlines: the prior span (measurement of the space the element needs) and show (routine to draw the element) functions, implemented for every inline node type, are replaced by a single layout function each which satisfies both needs.


Parth:
hype, curious about impact to perf, better, worse or not relevant?

tvanderstad:
> hype, curious about impact to perf, better, worse or not relevant?

perf looks similar, possibly worse cold start times but still single ms warm render times at 100kb

Parth:
Assuming you have the noob factor stuff on your branch?

tvanderstad:
> Assuming you have the noob factor stuff on your branch?

yeah, had to check, didn't run into it during testing

Parth:
great

Issue #4607, opened 2026-05-23
wayland DnD situation


Parth:
We migrated to winit in #4544 which squashed a lot of important bugs.

One of the more *obscure* bugs this fixed was our lack of wayland support. Our motivation to support wayland is that it's now the default on all the linux distros that are gnome-centric which is most people. But X11 works fine here, the impetus is really that some people will have fractional display scaling, which X11-on-wayland does not support, and caused our app to be blurry.

However DnD inherently doesn't work well on wayland, or at-least winit's implementation of it. You can find more information here: https://github.com/rust-windowing/winit/issues/1881

and looks like there's some promising progress: https://github.com/rust-windowing/winit/pull/4571, which makes me want to not do something dramatic like fork winit and try to implement it myself. But just to be balanced, this has been broken since 2022 within winit.

But in the meantime we have a choice between supporting DnD or Fractional display scaling (X11 or Wayland). And I think what we should do is go to X11 by default, and allow the user to configure this via settings. 

This is only going to be needed while this hole exists on Wayland, once this is fixed I look forward to falling back to the auto backend detection in winit and seamlessly supporting both.

eira-fransham:
Hey, author of https://github.com/rust-windowing/winit/pull/4571 here. We still haven’t decided if Wayland DnD is going to be included in that PR or if it’s going to be in a follow-up but we need it in Slint (my employer) due to an external client request. That request has a deadline of the end of June, so you should expect it to be implemented in our fork of winit within the next couple of weeks. I can’t make any promises about if/when it’ll be upstreamed though.

Parth:
Sounds good, thanks for clarifying. I'd probably be happy to use your fork if it's not upstreamed.

eira-fransham:
Hey, my DnD PR was just merged!

PR #4608, opened 2026-05-23
linux: restore drag and drop

Parth:
see #4607 for more info, this is a workaround


Parth:
QA'd with travis, and confirmed restored nicely

Issue #3277, opened 2025-01-07
lb-rs holds lock while doing file io for content merge


tvanderstad:
from logging added in #3275:
```
  2025-01-07T20:16:36.421157Z  WARN  sync merge held lock for 2.977357667s
    at libs/lb/lb-rs/src/service/sync.rs:951
    in sync
```
(file being merged is a 22MB svg after compression and encryption)

Parth:
another category of badness with our current merge is that whatever costs we incurred during the initial compute, say while importing a large file, sync will re-incurr these costs while trying to perform a merge.

it takes us 18s to import this file:

<img width="1505" height="498" alt="Image" src="https://github.com/user-attachments/assets/eb6e6e28-feb0-4f7e-9dd3-96b85e5ed72c" />

at first glance it looks like most of that compute is needed -- it's just spent asking a lib to compress a file, but then the same compute happens again, perhaps just due to the shape of the API.

will take a look at the compression library to see if there are some easy ways to speed up compression like there were for cryptography in the metadata operations.

Issue #4672, opened 2026-05-27
are we large files yet?


Parth:
+ [x] #1939 
+ [x] #1918 

Finishing these two quests meant that virtually any amount of any sized image is a reasonable experience in lockbook. It will eventually upload and it won't cause you to have something like an O(n) initial sync which may never finish or leave you in a corrupted state.

Our next set of milestones are going to be addressed by addressing low hanging fruit across the stack. Sorting out little papercuts like #3241, #3277, and other performance optimizations like the ones done in #4557.

+ [ ] ingress/egress a 2g video on macOS, linux & windows
+ [ ] ingress/egress a folder with lots of lil-shits (`.git` or `target/`)
+ [ ] a way to view the files (lb-fs?) #2505
+ [ ] can't be large files without better status (byte level, and file level status)
+ [ ] egress a 2g video on android and iOS

Beyond this we'll need proper #3401 which should bring us into unlimited max file size territory.

PR #4676, opened 2026-05-28
lb-rs, server: unlock larger files

### LINKED ISSUE #4620: macos not surfacing unexpected errors

Parth:
This PR does a handful of no-regret-low-hanging things to squeeze out more milage out of our existing non-streaming setup:

# wire format

when you ingressed a file into lockbook, despite compression that file would balloon to about 4x the size upon json serializing. This PR adds an optional header to `server` and `lb-rs` which allows lb-rs to specify a wire-format. Once the server is deployed the default wire format will become bincode which is way faster and doesn't balloon like json does.

# compression level

based on the ext of the file we now choose a compression level. This is transparent to people on older code as the compression format doesn't change. Most of our ingress / egress time is spent negotiating compression. See the screenshot in #3277. 

# don't hold locks while doing compression & big crypto

two places we do compression on ingress. on initial write and once again during sync merge.

I updated it so we're not holding a lock while doing IO, compression and crypto during a write.

I also updated merge to enable transfering an edit to the merge changeset without having to do compression and encryption.

# QA

+ [x] check that older clients will be able to interact with the new server
+ [x] check that old clients will be able to uncompress any compression level
+ [x] import export of large folders and files

fixes https://github.com/lockbook/lockbook/issues/4620

### DIFF
--- a/libs/lb/lb-rs/src/io/network.rs
+++ b/libs/lb/lb-rs/src/io/network.rs
@@ -123,15 +123,6 @@ impl Network {
     }
 }
 
-/// Pick a request body. Small payloads stay on the simple `Vec<u8>` path so
-/// the request goes out with `Content-Length` and identity encoding (which
-/// the server is happy with for GETs too). Only payloads above
-/// [`STREAM_BODY_THRESHOLD`] switch to a chunked stream of bounded `Bytes`
-/// pieces — that's what avoids the macOS `write(2)` `INT_MAX` cap, at the
-/// cost of `Transfer-Encoding: chunked`.
-///
-/// The `Bytes` chunks are `split_to` slices of one ref-counted backing
-/// allocation, so chunking doesn't copy the payload.
 fn body_for(serialized_request: Vec<u8>) -> Body {
     if serialized_request.len() < STREAM_BODY_THRESHOLD {
         return Body::from(serialized_request);
--- a/libs/lb/lb-rs/src/io/network.rs
+++ b/libs/lb/lb-rs/src/io/network.rs
@@ -66,10 +66,7 @@ impl Network {
 
         let wire_format = WireFormat::CLIENT_DEFAULT;
         let serialized_request = wire_format
-            .serialize(&RequestWrapper {
-                signed_request,
-                client_version: client_version.clone(),
-            })
+            .serialize(&RequestWrapper { signed_request, client_version: client_version.clone() })
             .map_err(|err| ApiError::Serialize(err.to_string()))?;
 
         if serialized_request.len() > 10 * 1024 * 1024 {
--- a/libs/lb/lb-rs/src/io/network.rs
+++ b/libs/lb/lb-rs/src/io/network.rs
@@ -87,7 +87,10 @@ impl Network {
             .header(WIRE_FORMAT_HEADER, wire_format.as_str())
             .send()
             .await
-            .map_err(|e| ApiError::SendFailed(e.to_string()))?;
+            .map_err(|e| {
+                warn!("send failed: {e:?}");
+                ApiError::SendFailed(e.to_string())
+            })?;
         if start.elapsed() > Duration::from_millis(1000) {
             warn!("network request took {:?}", start.elapsed());
         }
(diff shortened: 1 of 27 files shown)


Parth:
related #4672

Parth:
image

wire format issue found during macOS qa

Parth:
this work surfaces 3 problems that also exist on `master`:
- [x] now syncs can take a long time, obviously. And if you press the sync button while a sync is going on you get a beachball
- [x] there is an unexpected error that's surfaced if you try to push a file that is truly beyond our current stack, that is swallowed and never shown (a release blocker for me this release #4620)
- [x] there is some sort of sync that goes on, prob right on startup which is not shown to the user. Additionally if this sync fails in an *unexpected* manner, the app never opens

Parth:
okay I confirmed with a user that big files can now be imported and exported without beachballs. Though without better status hard to know what's going on, that may be a good contender for next work.

I'll fix the macOS problem with unexpected errors. This is how egui handles it:
image
image

this is what @tvanderstad was running into while he was trying to mess around with shared files (on macOS).

Apple doesn't show this situation at all: 
image

Parth:
alright ez enough 
image

libs/lb/lb-rs/src/io/network.rs
```diff
@@ -1,14 +1,30 @@
...
 use crate::model::api::*;
 use crate::model::clock::{Timestamp, get_time};
 use crate::model::errors::LbErr;
 use crate::model::pubkey;
+use crate::model::wire::{WIRE_FORMAT_HEADER, WireFormat};
+
+/// Cap on how many bytes hyper hands to a single `write()` syscall. macOS
+/// rejects writes larger than `INT_MAX` (~2 GiB) with EINVAL, so for
+/// large request bodies we feed reqwest a stream of bounded `Bytes` chunks
+/// instead of one giant slab.
```
Parth:
```suggestion

```

libs/lb/lb-rs/src/io/network.rs
```diff
@@ -1,14 +1,30 @@
...
+const STREAM_CHUNK_BYTES: usize = 4 * 1024 * 1024;
+
+/// Above this size we switch to a streaming body. Streaming forces
+/// `Transfer-Encoding: chunked`, which doesn't play nicely with GET requests
+/// (hyper-side handling for GET-with-chunked is iffy and the server returns
+/// BadRequest), so for ordinary small payloads we stay on the simple
+/// `Vec<u8>` path with `Content-Length`. 1 GiB is well below the macOS
+/// `INT_MAX` write cap and well above any normal metadata-shaped request.
```
Parth:
```suggestion

```

libs/lb/lb-rs/src/model/compression_service.rs
```diff
@@ -6,8 +6,33 @@ use flate2::write::ZlibEncoder;
...
 
+/// How many bytes of the payload to sniff for binary detection. Matches
+/// git's `buffer_is_binary`: a NUL byte in the first ~8 KiB is a strong
+/// signal the bytes won't deflate, so we skip compression.
+const SNIFF_BYTES: usize = 8000;
+
+const MIN_COMPRESS_SIZE: usize = 256;
+
+/// Pick a compression level by looking at the bytes we're about to write.
+/// `None` for binary-looking or tiny payloads; `Default` otherwise.
```
Parth, line 17:
```suggestion

```

libs/lb/lb-rs/src/service/documents.rs
```diff
@@ -103,37 +122,55 @@ impl LocalLb {
...
+            let db = tx.db();
+            let mut tree = (&db.base_metadata).to_staged(&db.local_metadata).to_lazy();
+            let file = tree.find(&id)?;
+            if file.document_hmac() != old_hmac.as_ref() {
+                return Err(LbErrKind::ReReadRequired.into());
+            }
+            let target_id = match file.file_type() {
+                FileType::Document | FileType::Folder => id,
+                FileType::Link { target } => target,
+            };
```
Parth, line 137: Doesn't the target need to be resolved before we compare hmacs?

Parth:
able to saturate 1gbps ul/dl
image

Parth:
subtle but impactful export bug discovered, has presumably been on `master` for many years: https://github.com/lockbook/lockbook/blob/master/libs/lb/lb-rs/src/service/import_export.rs#L173

needs to be `write_all`. `write` is allowed to do a partial write. 

doesn't apply to `read_document`, just specifically the import / export flows. Will be fixed in this PR as well.

PR #4684, opened 2026-06-01
md: fix android selection handle hit boxes

### LINKED ISSUE #4666: android selection handle touch targets too big
image you have this text foo bar where your cursor is right before the `r`
> `foo ba|r`

if you try to click on `b` your cursor won't jump there. the buffer to double tap to prompt the context menu vs change the selection range is a bit too strong.

tvanderstad:
fixes https://github.com/lockbook/lockbook/issues/4666


ad-tra:
works great. nice improvement. 

In the future it would be nice if pressing on the text above where the selection handle is drawn has the same effect as pressing the selection handle. currently it doesn't. it will be interpreted as you changed the selection to the line above that text.

PR #4699, opened 2026-06-04
all: narrow shame-frame-timing to egui ui code

Parth:


Parth:
when the app is in the background sending the instructions to the gpu are blocked, this narrows the scope of ui shame timing to just what the ui dev is responsible for

PR #4694, opened 2026-06-01
semantic search experiment

khalil-akremi:

PR #4704, opened 2026-06-06
semantic search experiment

khalil-akremi:


Parth:
👋 what's the difference between this and #4694? Also, consider keeping your PRs as drafts until they're actually ready for review.

khalil-akremi:
Thanks for the feedback.

This PR implements semantic search using USearch for vector similarity and nearest-neighbor retrieval. My understanding is that #4694 was based on an earlier attempt that ran into issues because my fork was out of date with the main branch, which led to merge conflicts. This PR contains the updated and corrected implementation.
I'm currently working on a follow-up improvement to support multiple embedding models instead of a single text embedding model.
Thanks for the suggestion regarding draft PRs as well , I'll keep future work as drafts until they're ready for review.

Parth:
If the older PR is no longer relevant you can close it, you can still convert this PR to a draft 
image

PR #4708, opened 2026-06-06
windows, linux: more reliable titles

### LINKED ISSUE #4621: egui title includes file extensions and don't clear

Parth:
fixes #4621


tvanderstad:
Unfortunately sending a viewport command also requests a repaint and we are now pinned at 120fps

https://github.com/emilk/egui/blob/main/crates/egui/src/context.rs#L3864-L3882

PR #4710, opened 2026-06-07
md: fix drag handles on wrong selection rect

tvanderstad:


Parth:
say more so QA is easier?

tvanderstad:
@Parth without this change the drag handles on iOS selection rects are sometimes attached to middle rects instead of the first and last ones. The cause is that the new selection rects for wrapping whitespace like newlines and for block indentation/syntax were reported out of order to iOS

Issue #4375, opened 2026-04-04
Change cli "copy" to "import"


paulhovey:
I know this is pedantic, but I think that `lockbook copy` as it currently stands (cli version 26.3.20) should really be named `lockbook import`.  The help text even agrees with me:

> copy         import files from your file system into lockbook

There's already `lockbook export` so this lines up.  Plus, if a user ever wants to actually _copy_ a file in the future and is using the cli, then `lockbook copy` would be the logical command to use.  There's already `lockbook move` and `lockbook rename`, so copying a file through cli seems reasonable.

The only other thing is there's `lockbook migrate-from`, which is an import, but currently only supports the Bear platform.  Maybe this becomes `lockbook import [bear/fs/etc]`.

PR #4731, opened 2026-06-11
changing copy to import in cli command

Amumug:
#4375


Parth:
@tvanderstad will take a more thorough pass over a lot of these soon

PR #4734, opened 2026-06-12
fix id prefix in lockbook stream out

Amumug:
#3015

https://github.com/user-attachments/assets/b4949832-4614-4487-8d02-99fff3f65ace


Parth:
I don't think this PR should have been approved or merged @tvanderstad @Amumug:
- this will complete single letters as id matches, I would say if we're doing a prefix match you would need to provide whatever amount of characters the short ids include
- The variant `FileInput::Path` should probably not be the mechanism used to complete ids, ideally those bits would be re-worked more fundementally
- there is a needless collect
- the pr title is not formatted the way all other prs are

Issue #4768, opened 2026-06-18
rfc: platform user agent


Parth:
Currently our UA string is whatever the `reqwest` library provides as a default. This doesn't tell our server anything about the user's end platform. A more traditional approach would be to encode information about the client's platform, and this issue will be the starting point for research and discussion around taking such an action.

First, how traditional is it? Obviously non-privacy competitors are swallowing up all the info they can get their hands on, but that's not our standard. Our comparison point is other privacy preserving apps, like signal, notesnook, standard-notes etc.
- Signal: [encodes a bunch of information in their useragent](https://github.com/signalapp/Signal-Android/blob/69e1146e2c5bbd6f2773dfe12f723e7cc88064be/app/src/main/java/org/thoughtcrime/securesms/net/StandardUserAgentInterceptor.java). Confirmed still the case on `main`.
- Notesnook, doesn't directly author useragent code (like us) but unlike us, uses web technologies built on `fetch` which leak a whole bunch of information about OS and architecture. Neither me, opus 4.8, or 5.5 could find any evidence they actively try to suppress this information. Weak signal but likely they do as well.
- StandardNotes: [Sends](https://github.com/standardnotes/app/blob/3d8bae26e8294bb4366d733e811934bda7bfd0f4/packages/api/src/Domain/Http/FetchRequestHandler.ts#L40-L46) an [environment](https://github.com/standardnotes/app/blob/3d8bae26e8294bb4366d733e811934bda7bfd0f4/packages/models/src/Domain/Device/Environment.ts#L1-L6) object with every request. This is more coarse, for example it seems like they can't distinguish iOS / android. But wait, they're an electron app too, so they intentionally disclose coarse platform, but incidentally exposes the whole gammut through default platform behavior. Even in react-native where they fall back to `okhttp` (on android) and `NSURLSession` (on iOS) there will be enough detail there to tell those platforms apart.
- Joplin: [Sends](https://github.com/laurent22/joplin/blob/f7812d283e9d018fbb861a82229dff420bce6f0d/packages/lib/JoplinServerApi.ts#L96-L102) [client info](https://github.com/laurent22/joplin/blob/f7812d283e9d018fbb861a82229dff420bce6f0d/packages/lib/services/joplinCloudUtils.ts#L67-L83) with ever request, and likely the whole electron fetch situation applies there. 

Are these services just not based enough? What about literally TOR? Even tor is sending coarse OS information: https://support.torproject.org/tor-browser/features/fingerprinting-protections/
[...]

PR #4773, opened 2026-06-23
lb-rs, server: user-agents #4768

Parth:


Parth:
@ad-tra broke some unrelated test is fine locally

Issue #4791, opened 2026-06-24
mobile scoped search


tvanderstad:
search results scoped to a folder; folder is captured from workspace focused parent on init
* [ ] workspace
* [ ] iOS
* [ ] android

Issue #4747, opened 2026-06-16
search visual refinements


Parth:
- [ ] spinner bad
- [ ] stale file bad
- [ ] open file loading spinner ugly
- [ ] filename search scrollbar height is wrong
- [ ] remove the lines
- [ ] back button should take you back to search

tvanderstad:
- [ ] should not contribute to suggested docs activity

PR #4834, opened 2026-06-24
ws & lb-rs: scoped search

### LINKED ISSUE #4781: desktop scoped search
search results scoped to a folder; folder is captured from workspace focused parent on init
* [ ] workspace
* [ ] iOS
* [ ] android

Parth:
while searching there is now a filter button:

image

you can use it to scope search to a particular directory: 

image

image

hitting escape will clear the filters and then the next escape will clear the query.

you can also get here quickly / iteratively by searching for a folder directly:

image

you can see the folders have a lil filter icon to communicate what would happen

image

there is also a "search here" in the file tree context menu

image

this does fixes #4781

this does not address #4791 
this does not address #4747 

thinking about whipping up a quick video

### DIFF
--- a/libs/content/workspace/src/search/content.rs
+++ b/libs/content/workspace/src/search/content.rs
@@ -125,6 +125,12 @@ impl SearchExecutor for ContentSearch {
         self.focused_file = None;
     }
 
+    fn update_filter(&mut self, filter: Option<lb_rs::search::SearchFilter>) {
+        self.searcher.update_filter(filter);
+        self.selected = 0;
+        self.focused_file = None;
+    }
+
     fn set_kb_mode(&mut self, kb_mode: bool) {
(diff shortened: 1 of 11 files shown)


libs/content/workspace/src/search/content.rs
```diff
@@ -125,12 +125,18 @@ impl SearchExecutor for ContentSearch {
         self.focused_file = None;
     }
 
+    fn update_filter(&mut self, filter: Option<lb_rs::search::SearchFilter>) {
```
Parth:
```suggestion
    fn update_filter(&mut self, filter: Option<SearchFilter>) {
```

Parth:
oh also empty query no longer does suggested docs, if there's ever an empty query anywhere it will do recents-sort

PR #4835, opened 2026-06-24
editor: reduce margins on android

ad-tra:
| Before | After | 
| - | - | 
 | image  |  image |


Parth:
can either of you remind me why this is not possible on iOS, what's the diff between the two mobile platforms?

tvanderstad:
should be possible with mastery of gesture conflicts that we may already have; just have to support tapping that area + horizontal drag for file tree drawer + vertical drag for scroll

PR #4873, opened 2026-07-08
releaser: fix desktop snap build (gnome SDK libc mismatch)

Parth:
curious if an agent can auto fix the snap pipeline:

> Follow-up to #4872, which was necessary but not sufficient.
> 
> Root cause, confirmed by a minimal repro in CI: the gnome extension injects the SDK snap's library dirs — including its own `libc.so.6` — into `LD_LIBRARY_PATH` in the part's build environment. The GitHub runner image updated host glibc (20260628 / Ubuntu 24.04.4) while `gnome-46-2404-sdk` rev 187 still ships an older one, so host binaries executed under that environment die with:
> 
> ```
> /snap/rustup/1504/bin/rustup: symbol lookup error: /snap/gnome-46-2404-sdk/current/usr/lib/x86_64-linux-gnu/libc.so.6: undefined symbol: __nptl_change_stack_perm, version GLIBC_PRIVATE
> ```
> 
> The dynamic loader exits 127, which craft-parts misinterprets as "'rustup' not found". The CLI snap has no gnome extension, which is why only the desktop build failed.
> 
> Overriding `LD_LIBRARY_PATH` to empty in the part's own `build-environment` (part entries land after the extension's) makes everything resolve against host libc. Verified in CI: a minimal gnome+rust part fails validation without the override and packs successfully with it.


Parth:
(so far it is not doing good)

PR #4975, opened 2026-07-31
cli: improve subcommand naming

### LINKED ISSUE #2457: stream in can't handle new files
cannot author new file via `stream in`

### LINKED ISSUE #3279: cli should have a way to understand size information

### LINKED ISSUE #3561: cli export inflexible
I wanted to export a file to a folder that exists already and not create another folder inside it

### LINKED ISSUE #4375: Change cli "copy" to "import"

tvanderstad:
### Breaking Command Renames
* `copy` → `import`
* `stream out` → `cat`
* `stream in` → `write`

### Commands
* `write` now creates the path if missing
* new `duplicate` command for in-place document copies
* `export` now supports writing a document straight to a file path, overwriting with `--force`, and unpacking folder contents into dest with `--contents`
* new option `account export --phrase` exports your account phrase (no surprises here)

### Behavior
* `list --long` shows type, size, and mtime (similar to `ls -l`)
* `account status` shows username, unsynced count, used/cap, near-full warning

### Docs
* `docs/cli.md` reworked

fixes: #4375
fixes: #3279
fixes: #2457
fixes: #3561


docs/cli.md
```diff
@@ -1,63 +1,126 @@
 # CLI
-See [installing](installing.md) or [building](building.md) to aquire Lockbook's CLI.
+The Lockbook CLI is a terminal client for your encrypted files: edit notes, list and search the tree, sync with the server, and copy files between Lockbook and ordinary OS paths when you need plain files outside the app.
```
Parth, line 2: these changes aren't present in lb-docs/

Parth: things that feel like are missing from these docs:
- an emphasis on the idea that you completions exist and are worth ensuring are properly setup 
- virtual fs feels like it's lost in the table. The table I would ideally not maintain here.

docs/cli.md
```diff
@@ -1,63 +1,126 @@
...
+lockbook list -l /work
+lockbook edit /work/todo.md
+lockbook cat /work/todo.md
+echo "hi" | lockbook write /work/todo.md
+lockbook sync
+```
 
-## Customizing CLI's editor
-By default `lockbook` will try to determine an editor based on `$EDITOR` or `$VISUAL`. If you want to select an editor specifically for lockbook, distinct from these values, you can set `$LOCKBOOK_EDITOR`
+Copy between Lockbook and ordinary OS paths. `--contents` puts a folder’s children in dest (not `dest/<folder>/`):
```
Parth, line 70: I like this pattern for common tasks:
human readable explanation followed by the examples, I think streaming operations deserve that sort of carve out ideally with some commonly used tools (fzf, or grabbing some server logs, or whatever else feels appropriate for whatever you're trying to highlight)

Parth (approved):

PR #4978, opened 2026-07-31
docs: cli doc updates

tvanderstad:


docs/cli.md
```diff
@@ -2,38 +2,36 @@
...
 Install via a [package manager](installing.md) or [build from source](building.md). Flags and arguments for any command are authoritative in the binary:
-```
+```sh
 lockbook --help
 lockbook <command> --help
 ```
 
 This page is a tour and task guide. Prefer `--help` when you need every option.
 
-Tab completion is a core part of the CLI: it completes subcommands and expands paths and IDs from your Lockbook tree. Spending a minute to [set completions up](#completions) for your shell is worth it—without them, the CLI is much harder to navigate.
```
Parth, line 12: dropped intentionally?

tvanderstad: no good catch, some kind of fumble while managing diffs

tvanderstad: this whole pr is

PR #5007, opened 2026-08-19
ws: file cache children() performance

tvanderstad:
Makes `children()` O(1) (was O(n)) and `descendants()` O(n) (was O(N^2)) by storing files as a sorted vector with a hashmap by id that keys into it. The vector is sorted by parent, file type, name - the same as the file tree - which improves file tree performance because it doesn't have to do any intermediate sorts.


Parth:
In addition, consider using a faster scheme for hashing uuids: https://github.com/lockbook/lockbook/blob/4fb916573a42bf88831c203e690b788ee2905f76/libs/lb/lb-rs/src/model/lazy.rs#L568-L570

PR #5008, opened 2026-08-20
all: support traces for engineering work

tvanderstad:
Configures [`tracing`](https://docs.rs/tracing/latest/tracing/) so [`#[instrument]`](https://docs.rs/tracing/latest/tracing/attr.instrument.html) tags emit structured traces in addition to logs. Use `level = "trace"` to emit something to traces but not to logs. Traces are stored in [Perfetto](https://perfetto.dev/)'s open format.

Tracing is a more recent alternative/supplement to sampling-based methods. Sampling tools like `samply` or `perf` (powers `cargo flamegraph` at least on linux) either attach to your process with `sudo`-level permission or host it as a subprocess. They record thread call stacks to measure where threads are spending CPU time. Tool/platform support is mixed for observing threads waiting on locks or otherwise idle, making it sometimes a struggle to find non-CPU work holding up the UI thread. Flamegraphs include all functions in your app, leveraging debug symbols.

In contrast, traces record start/end times for deliberately instrumented intervals, producing a structured timeline of all meaningful work performed by your system. The trade-off is that the burden of choosing which functions to instrument is on you - but when done well, this is a benefit, since your traces/flamegraphs will only include functions you care about. Traces can also include single events / values which we leverage to record CPU & memory usage.

There are many libraries for writing trace data in the Perfetto format. Grok 4.6 selected [`tracing-perfetto-file`](https://github.com/LalitMaganti/tracing-perfetto-file), released just two weeks ago and with no GitHub stars but my own (kudos for whatever AI-SEO they've done). The repo is maintained by a senior staff engineer who works on Perfetto at Google, claims best performance, and exposes features we rely on (e.g. `.span_mode(SpanMode::Both)` for per-thread and per-function tracks). We've security reviewed the crate; it lives in `lb-rs` pinned exactly at `=0.1.1` and scoped to non-wasm targets, and its own dependency tree is small (`dashmap`, `hashbrown`, `rtrb`).

**Enabling Traces**:
There is no cargo feature. The subscriber is composed in one place, `lb-rs::service::logging`, and installed by `Lb::init`, so any client picks this up; the desktop installs it earlier from `host::run` so boot is traced too. Tracing is off unless the `LB_TRACE` env var says otherwise:
* unset, `0`, or `false`: off
* `1` or `true`: writes `{writeable_path}/traces/<timestamp>.pftrace`
* anything else: used as the trace file path

`LOG_LEVEL` still controls `lockbook.log` independently, so `level = "trace"` spans stay out of the log.

**Viewing A Trace**:
1. Run lb desktop with `LB_TRACE=1 cargo run -p lockbook-desktop`
2. Navigate to https://ui.perfetto.dev
3. Open `~/.lockbook/egui/traces/<YYYYMMDD-HHMMSS>.pftrace`

A long-running function on the main thread looks like this:
CleanShot 2026-08-20 at 14 31 13@2x

You can drag-select any time period and produce a flamegraph:
CleanShot 2026-08-20 at 14 31 52@2x

**Tips**:
* When you first open a trace you probably want to click that 'expand all' button
* Traces are emitted both per-thread (`thread-xxx` tracks at the top) and per-function (function-named tracks below thread tracks)
* Tracks `cpu_us` and `rss_bytes` indicate OS-reported CPU and memory consumption


Parth (changes requested):
I propose the following changes:
any subscribers live in lb-rs managed centrally. This is sorta the pattern we already have, and these global singleton loggers can be confusing to diagnose, it's easier to understand what's going on if all the subscribers are in one place. It's also the only way that I could generate perfetto files for workspace / lb-rs when doing something like iOS development.


I think the dev dependency is a mild form of bad outcome mitigation, for a lot of complexity. Compromising on of the dev machines is still a very bad outcome. I see we're depending on this nacent library via `=` and that's a great step. Ideally you've audited what comes from crates.io (not github) yourself (probably you did). The complexity associated with feature flags (especially if this is lowered into lb-rs) is that each build process is going to handle this slightly differently. For iOS builds / workspace builds the complexity bubbles up to be a required feature in workspace as well.

I propose that instead of juggling features across the project we just have env vars controlling the behavior of the logs. This is already a mild pattern in loggers, you can audit this and change the behavior to your liking. I'm open to just having a logging level mechanism toggling this behavior and/or explicit control of whether these files are produced or not. This is also relatively well established in rust tooling (think `RUST_BACKTRACE=1` for instance).

I am open to 3 outcomes:
1. pushback
2. you agree and fix
3. you agree, but dont feel like doing, and I do this

PR #5018, opened 2026-08-27
build: clone cargo deps with git cli

CoreyCole:
Configures Cargo to fetch Git dependencies with the system Git CLI instead of its built-in Git client.

This allows dependency fetches to use existing Git and SSH configuration, including keys available to OpenSSH but not loaded into `ssh-agent`. It fixes misleading "revision not found" failures when GitHub HTTPS URLs are rewritten to SSH.

Verified with:

```sh
cargo fetch --locked
```

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,3 +1,6 @@
+[net]
+git-fetch-with-cli = true
+
 [target.x86_64-pc-windows-msvc]
 rustflags = ["-C", "target-feature=+crt-static"]
 


Parth:
I'm not sure I understand how you ended up in a state where this fixed your problem. I checkout these deps locally and wire them up sometimes for local dev, I use ssh, I do so across macOS windows and linux and do not require this config.

It is possible you require this config because you work on private repos? It seems like you can specify this config globally for your system, rather than for our project. Potentially a better path forward. 

Feel free to re-open if I'm way off.

PR #5021, opened 2026-08-29
android: stripe billing and account upgrade improvements 

ad-tra:
| before | after |
| - | - |
| Screenshot_1788009374 | Screenshot_1788008216 |
| Screenshot_1788009380 | Screenshot_1788008257 |


Parth:
this is sick I will do something like this on apple too

PR #5039, opened 2026-09-04
lb-rs: do not panic when the syncer lags or closes

tvanderstad:


libs/lb/lb-rs/src/subscribers/syncer.rs
```diff
@@ -1374,14 +1379,17 @@ impl LocalLb {
...
+                    Ok(Event::Sync(SyncIncrement::SyncFinished(_))) => {
+                        self.fetcher().await.map_unexpected().log_and_ignore();
+                        self.populate_pk_cache()
+                            .await
+                            .map_unexpected()
+                            .log_and_ignore();
+                    }
+                    Ok(_) => {}
+                    Err(_) => return,
+                }
```
Parth, line 1392: uncontroversial lgtm

libs/lb/lb-rs/src/subscribers/syncer.rs
```diff
@@ -1314,33 +1314,38 @@ impl LocalLb {
...
+                    match events.recv().await {
+                        Ok(event) => {
+                            if sync_criteria(event) {
+                                self.sync().await.map_unexpected().log_and_ignore();
+                            }
+                        }
+                        Err(_) => return,
                     }
+                } else {
+                    self.sync().await.map_unexpected().log_and_ignore();
```
Parth, line 1348: lgtm

libs/lb/lb-rs/src/subscribers/syncer.rs
```diff
@@ -1314,33 +1314,38 @@ impl LocalLb {
...
                             }
                         }
                         Err(TryRecvError::Empty) => break,
-                        _ => {
-                            panic!(
-                                "unexpected broadcast receive error, returning local_change_worker"
-                            );
+                        Err(TryRecvError::Lagged(_)) => {
+                            // Missed events — treat as "something changed".
+                            should_sync = true;
```
Parth, line 1330: I don't understand what situation leads to this

tvanderstad: The broadcast channel is defined for 10k events:
```rs
let (tx, _) = broadcast::channel::<Event>(10000);
```

My agent tells me we emit up to 4 events per pulled metadata during sync and file fetching. So, initial sync on a new client with 2500+ markdown files and this would be triggered. Maybe we haven't hit it yet but seems like good hygiene

tvanderstad: if that works for you, you can merge this PR

Parth: I don't like the speculative nature of syncing under this circumstance. I think the other code changes are good hygiene, this change I wouldn't make unless we have a specific circumstance / failing test or something like that.

tvanderstad: How about I make a test account with 2500 metadata, clean sync it, and share the outcome? Or does it need to be real world use

tvanderstad: I'll go ahead and do that

Parth: a tidy testcase is fine for proof

PR #5052, opened 2026-09-14
Infra: Restore fuzzer

krisha-s10:


Parth:
- read the fuzzer post & related blog posts
- write an enum which specifies a set of actions, you're going choose n random actions from that enum and execute them
- then let's check in

PR #5058, opened 2026-09-16
lb-rs: TCP keepalive so sync can fail after a dropped response

tvanderstad:
Enable TCP keepalive on the lb-rs HTTP client to unblock sync on stalled connections without timing out large downloads or uploads.

QA:
- Close laptop during file download, wait ~5 minutes, re-open...
  - [ ] ...on master, reproducing sync breakage
  - [ ] ...on this branch, demonstrating recovery
- [x] Sync a large document (should still succeed; connection is not idle)


Parth (approved):
I like it, good qa steps good tests, light touch with the approach. If that test takes more than like 100ms (by design) just `#ignore` it

Parth:
> pagerduty-rs and gh_release stay on reqwest 0.11 because they haven't upgraded

sad

tvanderstad:
This is causing large uploads to fail because the server doesn't respond until the upload completes i.e. we are using only the receive side to indicate liveness and ignoring the send side. It appears the `reqwest` [feature](https://github.com/seanmonstar/reqwest/issues/2403) we'd need for this is not yet implemented.

Parth:
I wonder if `tcp_keepalive` is a viable path for us to go down? I believe it would mean no upgrade would be required

tvanderstad:
struggled to reproduce the issue so we're just going to dogfood this for a bit and see if I see no more instances of the issue in my logs from here

PR #5072, opened 2026-09-18
android: insert images from within a document

### LINKED ISSUE #5075: support the editor's camera toolbar button on android

farboundd:
fixes: #5075


Parth:
@farboundd in your testing what was the maximum sized image you were able to ingress? If you exceed that iirc @ad-tra reported previously that we would get an OOM on the java side is that what you're seeing too? 

What I always wondered about: If that threshold is N mb, and you ingress M mb from some platform that can handle it (macOS?) where M > N. Are you able to view that image in workspace on android?

I would expect, yes, which would suggest that that our limits in the rust world are larger than the limits in the java world. If you confirm this is the case you'd have some options to ingress ios-sized files on android.

Regardless a streaming api should be coming relatively soon.

farboundd:
25mb is a conservative limit to accommodate for android devices on the lower end. I was able to ingress larger files than that on my device. When you import aren't you allocating a buffer to read the file then another buffer to compress it and then another buffer to encrypt and serialize it vs viewing the image you just decrypt and decompress into one buffer. so importing is more memory intense than reading

Parth:
what you're describing applies to all platforms where the limits are massively higher, like 100x higher. So I think my question still stands. 

Take whatever you're able to ingress on your phone, assuming it's higher end let's say you can get 50mb in

then take your mac (?) and ingress a 200mb image in. Can you open that on android? Can you see how that tells us something about the peculiarity of android?

regardless if that's the case you should be able to do stop-gap streaming on just the JNI side until true streaming comes.

PR #5078, opened 2026-09-20
Migrate lb-rs and server to db-rs 0.4

Parth:


Parth:
@tvanderstad first QA task:
- [ ] hop on this branch, it's up to date with `master`
- [ ] go to the data directory of the client of your choice, make a copy of it in case something goes wrong so we can analyze.
- [ ] `ls` the contents dump them here. You should see a `CoreV4.db` somewhere in there
- [ ] run the client from this branch, use the app for a bit, you should see no perceivable difference in experience
- [ ] `ls` the contents of the data directory. dump them here. You should no longer see a `CoreV4`, you *should* now see a `CoreV5` folder with a `db.0.log` in there and a `db.lock` in there or something very similar.
- [ ] approve this PR if all that happens without a problem

PR #5090, opened 2026-09-21
lb-rs: reuse a conflict duplicate's key across merge retries

Luca409:
## Symptom

An iOS app that stores its data as `.jsonl` documents (via lb-rs through the Swift FFI) and syncs often from one phone ended up with 19 sync-conflict duplicates over a few days. 18 of them fail every read, on the phone and with the `lockbook` CLI on another machine, with:

```
unexpected crypto error: Decryption(Error)
```

For example, one `.jsonl` file spawned duplicates `-1`, `-2`, … `-16`. The **first** duplicate of a file decrypts fine. **Later** duplicates of the same file don't. Syncs were slow (2–13s) and frequent, so the conflicts likely came from uploads whose responses were lost. The trigger doesn't matter for this bug: any repeated conflict on the same non-mergeable file hits it.

## Fix

`duplicate_file_ids` already lives for the whole merge, including construction retries. The duplicate's AES key now lives in that map too, generated once when the id is allocated. Every `create_unvalidated` reuses it.

A fresh key per attempt wrote the new key into metadata while `decrypt_key` still encrypted the document with the key cached from the failed attempt. Other clients, and the creating client after restart, then fail AES-GCM. Documents already corrupted this way cannot be recovered from metadata. Only a process that still has the original key in memory could re-encrypt them.

## Test

`different_content_edit_not_mergable_twice` in `libs/lb/lb-rs/tests/sync_service_concurrent_change_tests.rs`. Two clients conflict twice on `/document.jsonl`. The second duplicate is `/document-2.jsonl`, and both the other client and a freshly synced client must read it.

## Related (separate, not fixed here)

`NameComponents::from` / `generate_incremented` in `libs/lb/lb-rs/src/model/filename.rs` treats a trailing `-<number>` in the stem as an existing variant. So a duplicate of `2026-09.jsonl` is named `2026-10.jsonl`, then `2026-11.jsonl`, and so on. These names look like real, unrelated files and can collide with them.


tvanderstad:
From QA, how to reproduce the issue. Not a clean or fast demo but we got it.


https://github.com/user-attachments/assets/26998acc-61b0-4409-9839-0b9961f2a958

tvanderstad:
Proof of fix:

https://github.com/user-attachments/assets/9c44ca38-8feb-4bba-8302-40d0a74d62a1

tvanderstad:
introduced in https://github.com/lockbook/lockbook/pull/3231

